PR #18, opened 2020-11-27
Fixed freecam bug, added announceInChat to TotemPopNotifier and added more rendering options to HoleESP.

seasnail8169:


supakeks:
Merge it I want totempopannounce minegame merge now merge

PR #35, opened 2020-12-06
Added InfinityMiner HUD and Module

Inclemental-55:
Hi I added a module and an HUD to go along with it called InfinityMiner. This allows indefinite mining, primarily in the nether, provided one has the mending enchantment. The user specifies the target block, and then the block they wish to repair off of. 

For instance one might select ancient_debris and quartz. Via Fabritone, ancient_debris will be mined until a desired durability threshold is reached, at which point it will switch to quartz to repair the pickaxe.

 It also includes an optional setting which will path-find back to where they started once their inventory is full. If their pickaxe reaches the durability threshold on the way home, it will divert to repair it, and then continue on its way home.

### DIFF
--- a/src/main/java/minegame159/meteorclient/modules/ModuleManager.java
+++ b/src/main/java/minegame159/meteorclient/modules/ModuleManager.java
@@ -308,6 +308,7 @@ public class ModuleManager extends Savable<ModuleManager> implements Listenable 
         addModule(new AntiCactus());
         addModule(new FakePlayer());
         addModule(new NameProtect());
+        addModule(new InfinityMiner());
     }
 
     private void initMovement() {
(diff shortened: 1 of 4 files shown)


notseanyay:
based

MineGame159:
Good job

Inclemental-55:
Thank you! If there are any issues let me know, happy to keep it up to date. Will be adding auto-skulker filling next I hope.

MineGame159:
Auto skulker filling?

squidoodly:
When you inventory gets full it puts the stuff you are mining into shulkers for you.

Inclemental-55:
Exactly, would be an optional setting. But would extend the mining duration almost infinitely.

squidoodly:
I had the same idea a while ago but I was too lazy to figure it out. lol

Best of luck. I don't think it will be too hard.

MineGame159:
Ohhh, ok good luck

Inclemental-55:
Thank you! Also you guys have done an amazing job with this client. Thank you for letting me contribute.

linustouchtips:
chad commit

PR #36, opened 2020-12-07
Infinity miner

Inclemental-55:
Added Autologout feature. If desired, will automatically logout once your inventory is full. If walk home is also enabled it will walk home first and then logout. 
Added SmartModuleToggle feature. Will automatically toggle useful modules. Once InfinityMiner is toggled off, it will restore previous settings minus any changed made by the user. 
Changed the durability threshold to a percentage, much easier to understand as a user. 
Vastly improved the code-base. 
Fixed crash on setting change issue. 

Cheers


notseanyay:
looks cool

MineGame159:
If you want I can make a setting type thats just for one block and replace the string settings

Inclemental-55:
Oh lord that would be amazing. Please do so .

MineGame159:
Also I just noticed you named the settings like "Traget Block", they are supposed to be "target-block". Gonna fix that when adding the block settings

Inclemental-55:
Got it sorry about that. Since you are adding a single block selection setting... Should an automine module be added as a graphical way to use baritone?

MineGame159:
Nah, thats unnecessary. Also you were executing the baritone commands as raw string, changed them to use the respective api methods instead.

Inclemental-55:
Perfect, I couldn't find those methods for some reason so I settled for using strings. I'm still working on the side figuring out Shulker box filling, having some trouble but i should be able to get it done.

PR #41, opened 2020-12-09
Fixed not-saving bug for moduleInfoModules in HUD

dakone22:
Module list info wasn't saving before. (I hope) it fixed now

example


dakone22:
_(As you know, I don't know Java and Fabric but)_ Why there are in `meteorclient/settings` so many same list settings with the same code?
Like `BlockListSetting`, `EnchListSetting`, `EntityTypeListSetting`, `ItemListSetting`, `ModuleListSetting`, `SoundEventListSetting`, `StorageBlockListSetting`.
There is no way to make base class like `ListSetting`?

MineGame159:
Definitively, if you look there is a new branch called settings-rewrite that I started yesterday.

PR #44, opened 2020-12-10
Added outline customization to CrystalAura & a few more changes

supakeks:


squidoodly:
I don’t understand how the auto log commit is any better than what we already have.

PR #55, opened 2020-12-14
Renamed Xp Bottle Thrower to Auto EXP

supakeks:


supakeks:
I think I fucked up something so I closed this

PR #66, opened 2020-12-16
Speed Mine

fluse1367:
Should be clear.
- Normal Mode (modifies final breaking speed value)
- Haste 1&2 Mode
- Packet Mode

### DIFF
--- a/src/main/java/minegame159/meteorclient/events/EventStore.java
+++ b/src/main/java/minegame159/meteorclient/events/EventStore.java
@@ -84,7 +84,6 @@ public class EventStore {
     private static final PlaySoundEvent playSoundEvent = new PlaySoundEvent();
     private static final WaypointListChangedEvent waypointListChangedEvent = new WaypointListChangedEvent();
     private static final BreakBlockEvent breakBlockEvent = new BreakBlockEvent();
-    private static final BlockBreakingProgressEvent blockBreakingProgressEvent = new BlockBreakingProgressEvent();
     private static final PlaceBlockEvent placeBlockEvent = new PlaceBlockEvent();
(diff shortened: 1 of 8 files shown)


MineGame159:
Packet mine is already a separate module

fluse1367:
Look up the code, it's a different thing

MineGame159:
I see the code is not the same but the description is the same

fluse1367:
Wdym

MineGame159:
"While in PACKET mode, hitting a block %sonce%s will eventually break it." that is what you send in chat when enabling packet mod  and it is exactly what packet mine module does

fluse1367:
Yea ok i got the point, but as far as i tested it speedmine in packet mode is still faster than packetmine. my fault if i'm wrong ¯\_(ツ)_/¯

MineGame159:
How does the packet mode work? Looking at the code and all I see it doing is sending a STOP_DESTROY_BLOCK player action packet once the block break progress is greater or equal to 1.

fluse1367:
I think you overlooked the return statement: it will immediately send the STOP_DESTROY_BLOCK action _unless_ the break progress is greater or equal to 1

MineGame159:
Oh yeah, sorry, So then its doing the same thing as packet mine. Spamming the server with START and STOP packets, it propably just felt faster but werent.

fluse1367:
lmao, prolly. should i remove it then or rather packet mine (i think its clearer if its all in the same module)

squidoodly:
Why would you move packet mine to speed mine when it doesn’t make it faster?

fluse1367:
True, my bad. I'm gonna remove packet mode then

PR #86, opened 2020-12-27
add noclip

LeftyIsHere:
pls


squidoodly:
There is a lot of unnecessary code here. net.minecraft.client.MinecraftClient.getInstance().player can just be changed to mc.player. net.minecraft.util.math.Vec3d can be changed to just Vec3d. Fix it and then I’ll have a look.

seasnail8169:
Squid literally dont bother. This code is bullshit and will not work well. I really don’t think this is something we want to add cause we will just get people saying “WHY DOES NOCLIP NOT WORK???????”. Also this code is taken directly from the meteor skid “blackhack” lol

PR #88, opened 2020-12-27
Improved VisualRange - Removed MessageAura (easy)

supakeks:


squidoodly:
We don’t remove features unless they don’t work anymore.

supakeks:
Ok have a good day :(

PR #91, opened 2020-12-28
ModMenu integration and AntiInvisibility

JFronny:
Adds a ModMenuApi impl that opens the config screen (useful for configuring things without joining a world) and a simple module that reenables rendering for invisible entities


MineGame159:
You can open the gui in main menu the same way as you would in-game.

JFronny:
What about the other half of this PR?

MineGame159:
If you separate it then I guess its fine.

JFronny:
Separate as in new PR? The relevant commit is already separate from the modmenu commit

MineGame159:
Yes, I can't merge only some commits.

PR #92, opened 2020-12-28
AntiInvisibility

JFronny:
See #91


JFronny:
This is now included in ESP, I will close this PR as it seems to no longer be required

PR #111, opened 2021-01-05
armor flip fix

kecleon:
loop through array values for item slots, use index for hud positioning


MineGame159:
Sorry but Im not merging that. Allocating int array every frame just to use it as indexes is bad.

PR #112, opened 2021-01-05
better armor flip fix 

kecleon:
keeping track of item slot and hud position

hopefully better than the last one


MineGame159:
Doesnt this do the same thing as before?

kecleon:
> Doesnt this do the same thing as before?

The newer commit in this pr doesn't, it's one int for position on hud, one int for item slot, and it functions properly in game when toggling the flip armor option. If you don't want the previous bad commit I can make a new branch/pr without it.

PR #118, opened 2021-01-07
capitalization of module names, fullbright rename

notseanyay:
…cally the correct one


squidoodly:
All the module names are lower case for a reason.

PR #123, opened 2021-01-07
Added Skylight Updates to NoRender

Wide-Cat:
Skylight updated still cause fps lag in 1.16, so this is for fixing that

also first pr, pog


Wide-Cat:
ez

PR #145, opened 2021-01-14
Totem Count Hud

tyrannus00:
New Feature for the hud, shows the amount of totems in your inventory. Quite useful.


seasnail8169:
nice, how about some other items whilst you are at it? crystals gaps obby etc

seasnail8169:
but i dont think it should use the armor scale, maybe make a seperate slider called items scale and make them use that

PR #146, opened 2021-01-14
should organize the aura's settings somewhat? i hope so

notseanyay:


squidoodly:
Why would you change the epic description of the multi-targeting?

PR #148, opened 2021-01-15
Improved KillAura rotation setting

supakeks:
Some servers may have anticheats that detect hitting/facing a specific part of the target's body (in this case feet) through walls, so instead of only having the rotation always looking at the target's feet, you can choose where to look at.


MineGame159:
Make it so the "eyePos" and "chestPos" Vec3ds are not created if their mode is not selected. Aka move them to the correct switch case.

Issue #142, opened 2021-01-13
Knockback strange


Clashbestie:
Knockback is acting very strange, as seen in this video (I only pressed space and w), playing on a Paper-Server

https://user-images.githubusercontent.com/57112209/104510645-c07c9e00-55eb-11eb-9f0c-2f0c61e82b5a.mp4

seasnail8169:
You clearly only have the HUD active therefore its clearly not meteor that is causing whatever issue you are having. A module would have to be active for meteor to cause something remotely similar to that.

Clashbestie:
Relax, xD.
I did some testing:
Yeah you are right I only have the HUD active but Velocity still kinda manipulates the Velocity applied.
Why? Don't ask me the code seems fine, but I know [where ](https://github.com/MeteorDevelopment/meteor-client/blob/252a72910dab2ac335cf164f78fe72502dcf71b4/src/main/java/minegame159/meteorclient/mixin/EntityMixin.java#L53).
Adding `if (velocity.isActive())` infront of the line resolves this problem.
Idk why the returned values from Velocity have a effect when disabled, because y*1 is still y?
I logged the values, they are still the same after multipling, but the y(the one thats provided as arg to the funtion) spikes sometimes, idk why.

And btw the client is awesome.

Issue #42, opened 2020-12-09
AutoReconnect and events


dakone22:
AutoReconnect won't working for me at all. 

![kicked](https://user-images.githubusercontent.com/35692622/101563540-798efd80-39da-11eb-8923-35aa26367c87.png)
![settings](https://user-images.githubusercontent.com/35692622/101563563-8ad80a00-39da-11eb-90c5-d86929d74fe2.png)

It is not only mine issue? 
(If only mine, maybe it depends on server? Like server's way to kick)

dakone22:
[As I see in code](https://github.com/MeteorDevelopment/meteor-client/blob/master/src/main/java/minegame159/meteorclient/modules/misc/AutoReconnect.java#L49), AutoReconnect module listen `OpenScreenEvent`, but I think it never (at least for me) calls when I got disconnected.

dakone22:
And as I see [here](https://github.com/MeteorDevelopment/meteor-client/blob/master/src/main/java/minegame159/meteorclient/mixin/ClientConnectionMixin.java#L33), sometime before, there was `onDisconnect` event. Maybe it could be better for AutoReconnect's listener?

PR #151, opened 2021-01-15
Autoreconnect and velocity fix

Clashbestie:
Fixes [AutoReconnect and events](https://github.com/MeteorDevelopment/meteor-client/issues/42#issue-759922788) and [Knockback strange](https://github.com/MeteorDevelopment/meteor-client/issues/142#issue-785459139).

By using mixins to display the custom DisconnectScreen and not the OpenScreenEvent witch return null as Screen when opening most of the Screens, and adding adding a check to the EntityMixin as explained in the second link.


MineGame159:
Can you remove the INSTANCE field on AutoReconnect and use ModuleManager instead, if you dont want to get it every tick then you can cachce it to a field.

Clashbestie:
> 
> 
> Can you remove the INSTANCE field on AutoReconnect and use ModuleManager instead, if you dont want to get it every tick then you can cachce it to a field.

like that?

MineGame159:
No :sob: using strings is unefficient (only used for saving) and in code you can do it like .get(AutoReconnect.class)

MineGame159:
Also there is a helper ModuleManager.INSTANCE.isAction(AutoReconnect.class) method

MineGame159:
@Clashbestie do you not test your changes? The build failes.

MineGame159:
Hello? @Clashbestie

PR #158, opened 2021-01-16
Jesus improve

dakone22:
By [special request](https://discord.com/channels/689197705683140636/689197706169942135/799737702857179157) from **Ricky The Raccoon#9202**.
You can now set up separately for water and lava behaviour. 
You can adjust whether or not to go under the liquid if you fall from a certain height. 
\+ go under the lava to avoid fall damage if you have a *Fire Resistance* effect.


MineGame159:
Good changes but you used the same setting name 2 times (for water and lava). This is currently a limitation of the system that every setting must have unique name.

dakone22:
@MineGame159,
[one](https://github.com/MeteorDevelopment/meteor-client/blob/master/src/main/java/minegame159/meteorclient/modules/misc/BetterChat.java#L105)
[two](https://github.com/MeteorDevelopment/meteor-client/blob/master/src/main/java/minegame159/meteorclient/modules/misc/BetterChat.java#L138)

And they are in different `SettingGroup`s

MineGame159:
I know they are in different groups. As I said its a limitation of the system. And I didnt know about BetterChat having it too. Gonna fix that.

dakone22:
Ok, I will fix this.
And I can fix `BetterChat` because I am doing chat protection for it.

MineGame159:
As an explanation, setting groups didnt exist from the start, they were added later on and when adding them I didnt want to break every already existing config. So how it works rn the settings are all just saved in one big list and setting groups do nothing for saving.

Ok perfect.

MineGame159:
Also a tip for the future. Use separate branch for every pull request.

PR #159, opened 2021-01-16
Added chat protection from sending messages with coords in global chat, fixed and improved nbt command

dakone22:
Added chat protection from sending messages and messages with coords in global chat.
By [suggestion](https://discord.com/channels/689197705683140636/689202616441503870/793566248574386247) from `@supakeks#0001` and [suggestion](https://discord.com/channels/689197705683140636/689202616441503870/793556976503619644) from `Chompasaurus513#3571`.
chat_protection

Also fixed `.nbt get` crash with empty nbt data. Added `.nbt copy`. Added chat button for copy in `.nbt get`.
nbt_copy2

### DIFF
--- a/src/main/java/minegame159/meteorclient/commands/CommandManager.java
+++ b/src/main/java/minegame159/meteorclient/commands/CommandManager.java
@@ -47,7 +47,7 @@ public class CommandManager {
         addCommand(new Clip());
         addCommand(new Bind());
         addCommand(new Toggle());
-        addCommand(new Nbt());
+        addCommand(new NBT());
     }
 
     public static void dispatch(String message) throws CommandSyntaxException {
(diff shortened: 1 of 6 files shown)


MineGame159:
image
Why is this there?

dakone22:
In order not to forget. Should it be removed?

MineGame159:
image
I dont think its worth it replacing this with that method.

dakone22:
Why? I think it's realy should be replaced than to have `Config.INSTANCE.getPrefix()` + getting name + space + arg

MineGame159:
`Config.INSTANCE.getPrifx() + this.name + " copy"` vs `CommandManager.getCommandString(this.name, "copy")`

You basically just got rid of getting the prefix and replaced it with calling method that requires more characters.

MineGame159:
If you really wanna get fancy then sommething like `CommandManager.get(NBT.class).run("copy")` could work.

dakone22:
> You basically just got rid of getting the prefix and replaced it with calling method that requires more characters.

Yes. And as well you can use muliple args instead adding spaces between them.
```java
CommandManager.getCommandString(

MineGame159:
Imo the run() method would be better.

PR #160, opened 2021-01-16
Count display for any item on hud

tyrannus00:
also removes the totemhud since that one is unnecessary now.
Shit I didnt notice I had duped commits in here. Sorry for the mess :/


baiorett:
monkening

PR #165, opened 2021-01-18
Some Jesus safe improvement and NoFall

dakone22:
Added some Jesus safe improvement: setting if you burning you will dip into water, setting if Fire Resistance effect is nearly over it won't be allowed be under lava.
Added checking if creative mode for NoFall.


dakone22:
@seasnail8169, ? Why closed?

PR #170, opened 2021-01-19
Added "Fast" option to AutoTotem

vitrack:
Added a "Fast" option to AutoTotem that switches to a totem at the start of a tick, improving the speed of it, and making the user less likely to die when being damaged quickly.

P.S. I also left some comments in the module, you can remove them if you dont like them, or just tell me, and i will remove them myself.


squidoodly:
This will not make auto totem any faster. The time it takes the client to do these checks is miniscule compared to the time between ticks. If you really want to make it faster then try moving it to onDamageTaken or onPreTick. This also completely invalidates smart totem and makes offhanding impossible. And to top is all off, why not just fix the issue instead of putting a try catch block around code?

PR #171, opened 2021-01-19
ESP/Tracer Colour by Name Tag

Mersid:
Adds an option to show players' ESP and Tracers dynamically based off the colour of their name tag. For games and servers where teams or faction members may be coloured differently, this makes it easy to ascertain whose team a player is on, without needing to add them to a friends list.

This is the default setup. Note all players have the same colour. I'm running this on Hypixel as they grant VIP members different coloured name tags, but this system should work with most* servers that apply name colours for any reason.
image

Users now show up with the colour of their name tags. This setting can be toggled on or off in the ESP and Tracers menu for their respective components.
image

Obviously, the utility of this function depends entirely on how the server opts to colour their users, and how this is utilized is dependent on the specific server and the Meteor user; it merely grants them another option.

Some things to note about this:

1. It only works on players, and will not apply to named mobs
2. For players with multiple colours, it will select the colour that is most prominent (has the most characters)
3. *It does not work for servers that still use the old formatting style (format codes)


seasnail8169:
Implemented this in a better way :P

PR #195, opened 2021-01-23
various setting changes and additions

notseanyay:


src/main/java/minegame159/meteorclient/modules/combat/Swarm.java
```diff
@@ -60,13 +60,13 @@ public Swarm() {
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
 
     private final Setting<String> ipAddress = sgGeneral.add(new StringSetting.Builder()
-            .name("ip-address")
+            .name("iP-address")
```
notseanyay, line 63: this is not a typo, it's so that it reads IP Address instead of Ip Address 😼

PR #199, opened 2021-01-25
Improved ElytraPlus AutoPilot setting + Crystal/Anchor Aura PauseOnHealth setting.

supakeks:
Improved elytra+ autopilot settings.

New "Firework" setting bypasses the speed patch part of the elytra fly patch epearl made on ec.me lol

CrystalAura and AnchorAura now have a PauseOnHealth setting. If your health is lower than your selected value, the module will get paused. By default both AnchorAura and CrystalAura have it at 0 which is pretty much no pause for it.

### DIFF
--- a/src/main/java/minegame159/meteorclient/modules/combat/AnchorAura.java
+++ b/src/main/java/minegame159/meteorclient/modules/combat/AnchorAura.java
@@ -57,6 +57,7 @@ public class AnchorAura extends Module {
 
     private final SettingGroup sgPlace = settings.createGroup("Place");
     private final SettingGroup sgBreak = settings.createGroup("Break");
+    private final SettingGroup sgPause = settings.createGroup("Pause");
     private final SettingGroup sgMisc = settings.createGroup("Misc");
     private final SettingGroup sgRender = settings.createGroup("Render");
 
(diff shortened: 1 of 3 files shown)


squidoodly:
If you scrolled down literally 5 lines in either Crystal Aura or Anchor Aura you would see they already have something similar to PauseOnHealth where they stop calculating if your health is lower then min health.

supakeks:
I did see that, but the minHealth setting is only used if place mode is safe in this line (in CA):
`if (getTotalHealth(mc.player) <= minHealth.get() && placeMode.get() != Mode.Suicide) return;`

The 2nd time minHealth is used is in this line (in CA): `return (!(breakMode.get() == Mode.Safe) || (getTotalHealth(mc.player) - DamageCalcUtils.crystalDamage(mc.player, crystalPos) > **minHealth.get()** && DamageCalcUtils.crystalDamage(mc.player, crystalPos) < maxDamage.get()));`

this isn't really what pause on health does as it wouldn't be doing anything if your place mode was suicide or if those calculations from the 2nd time minHealth is used are false.

Let me know what I could change.

squidoodly:
But why would you want it to pause on low health if you have it on suicide mode?

supakeks:
You might wanna deal high damage even if you're gonna take a lot of damage but might not wanna pop a totem/suicide

squidoodly:
That is exactly what minHealth already does. I don't understand the difference. If your setting activates then the whole thing is paused when you are at low health. If minHealth activates then it gets paused when you are on low health and it stops it from doing to much damage to yourself. Your setting just seems like a worse version of what is already there for no reason.

supakeks:
The thing is it'll still break crystals/anchors which you may not want. When you're on really low health anyone will try to reheal, and if crystal/anchor aura keeps breaking, you won't be able to heal which will end up putting you at .5 hearts. My setting completely pauses Crystal/Anchor aura, just like the 3 other settings above it. It would be a bit weird to have that pause on health setting not fully pause the module but have the 3 other pause settings fully stop the module.

squidoodly:
If you are on low health then it won't break anything anyway because any damage you would take would bring your health lower than min which means it won't break anything when you are at min health and let you heal. Like I said, I don't think this adds anything. You'll have to ask someone else what they think.

supakeks:
Ok, how about I remove the min health setting and just replace it with the pauseonhealth setting? Since it technically is pausing Crystal/Anchor Aura then it would be better to have it as a pause setting, just like the 3 other pause settings. I would replace the minHealth.get() methods with pauseOnHealth.get().

PR #201, opened 2021-01-26
Elytra+ AutoPilot improvements + Anchor/CrystalAura pause on health

supakeks:
Replaced "min health" with pause on health as it does the same as pause on drink, mine and eat, which is pausing if something happens, which in this case is pausing if your health is lower than the value you chose.


squidoodly:
You’re not replacing what is already there for no reason. Just give up on pauseOnHealth.

asdffdsdaf:
ezn't

PR #206, opened 2021-01-27
CrystalAura rotation changes bruh

supakeks:
Default rotation settings send 2 times less packets now

Added an option to disable the remove entity thing as that's what was desyncing the fuck out of players on ncp servers (still on by default)

Also made the rotation resets only be sent when disabling because sending them after breaking/placing is packet spamming

Rotation Lock setting can be useful for servers with way more stricter rotation checks that may require you to look at a specific part of the support block

### DIFF
--- a/src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
+++ b/src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
@@ -316,7 +316,7 @@ public class CrystalAura extends Module {
 
     private final Setting<Double> rotationLock = sgRotations.add(new DoubleSetting.Builder()
             .name("place-rotation-lock")
-            .description("Where to look at when placing crystals.")
+            .description("How high to look at the support block you're placing your crystals on.")
             .min(0)
             .defaultValue(100)
             .sliderMax(100)
@@ -458,7 +458,7 @@ public class CrystalAura extends Module {
     @Override
     public void onDeactivate() {
         assert mc.player != null;
-        if (switchBack.get()) if (preSlot != -1) mc.player.inventory.selectedSlot = preSlot;
+        if (switchBack.get() && preSlot != -1) mc.player.inventory.selectedSlot = preSlot;
         for (RenderBlock renderBlock : renderBlocks) {
             renderBlockPool.free(renderBlock);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
```diff
@@ -422,11 +458,19 @@ public void onActivate() {
     @Override
     public void onDeactivate() {
         assert mc.player != null;
-        if (preSlot != -1) mc.player.inventory.selectedSlot = preSlot;
+        if (switchBack.get()) if (preSlot != -1) mc.player.inventory.selectedSlot = preSlot;
         for (RenderBlock renderBlock : renderBlocks) {
             renderBlockPool.free(renderBlock);
         }
         renderBlocks.clear();
+        if (target != null && resetRotations.get()) {
```
squidoodly, line 466: What does this actually do? It seems to just set where the player is looking to where they are already looking.

src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
```diff
@@ -295,6 +305,32 @@
...
+    private final Setting<RotationMode> rotationMode = sgRotations.add(new EnumSetting.Builder<RotationMode>()
+            .name("rotation-mode")
+            .description("When to rotate when using Crystal Aura.")
+            .defaultValue(RotationMode.Place)
+            .build()
+    );
+
+    private final Setting<Double> rotationLock = sgRotations.add(new DoubleSetting.Builder()
+            .name("place-rotation-lock")
+            .description("Where to look at when placing crystals.")
```
squidoodly, line 319: Would be nice to get a better description of this.

src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
```diff
@@ -422,11 +458,19 @@ public void onActivate() {
     @Override
     public void onDeactivate() {
         assert mc.player != null;
-        if (preSlot != -1) mc.player.inventory.selectedSlot = preSlot;
+        if (switchBack.get()) if (preSlot != -1) mc.player.inventory.selectedSlot = preSlot;
```
squidoodly, line 461: Can this be made into one if statement?

squidoodly (changes requested):

src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
supakeks: It's doing what the return mode used to do, reset your rotation server-side as you would still be looking at the support block/end crystal

PR #208, opened 2021-01-29
Rolled back squidoodly's change with CrystalAura vertical range.

supakeks:
Most players use place range to around 5/6, this absolutely kills performance on most PCs. I know creating more settings is annoying but either this or keeping it at 3 will prevent performance from being killed.

My fps dropped to 1 when placing on a block at my head's height, which is something crystal aura does almost all the time while in a safe hole.


squidoodly:
I told you so.

PR #209, opened 2021-01-29
Removed NoRotate

supakeks:
1. You can cancel the packet using packet canceller
2. It wasn't cancelling the right packets
3. It broke every rotation sent to the server


jiaaluo:
monke

supakeks:
cope

PR #210, opened 2021-01-29
Made NoRotate work (I think)

supakeks:
Old NoRotate did the following:

1. Break rotations sent by the client to the server
2. Server would still force you to look at a specific spot, making the module not work


jiaaluo:
still monke

seanbutbased:
Rofl DOUBLE DENY

PR #212, opened 2021-01-30
renamed AntiAutoAnvil to AntiAnvils, overall setting changes.

seanbutbased:


TechAle (commented):
Why .name("gui-scale") -> .name("gUI-scale") ?
and also .name("xp") -> .name("xP")

seanbutbased:
because it makes UI in gui capitalized. otherwise its like Gui

PR #214, opened 2021-01-31
Added RayTracing, toggelable MultiPlace, and implemented new Rotations system into CrystalAura

zyktex:
- RayTracing will check if you can see blocks
- Place direction will now be dynamic instead of always being UP
- Rotations will now be done using the new Rotations system
- Rotations will now be aimed towards the side you're placing against if strict-look is enabled

### DIFF
--- a/src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
+++ b/src/main/java/minegame159/meteorclient/modules/combat/CrystalAura.java
@@ -130,7 +130,7 @@ public class CrystalAura extends Module {
 
     private final Setting<Boolean> rayTrace = sgPlace.add(new BoolSetting.Builder()
             .name("ray-trace")
-            .description("Only places when you can see block, unless it it closer than walls range.")
+            .description("Whether or not to place through walls.")
             .defaultValue(true)
             .build()
     );
@@ -167,13 +167,6 @@ public class CrystalAura extends Module {
             .name("strict")
             .description("Won't place in one block holes to help compatibility with some servers.")
             .defaultValue(false)
-            .build()
-    );
-
-    private final Setting<Boolean> ignoreWalls = sgPlace.add(new BoolSetting.Builder()
-            .name("ignore-walls")
-            .description("Whether or not to place through walls.")
-            .defaultValue(true)
             .build()
     );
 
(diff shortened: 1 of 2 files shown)


jiaaluo:
good

seasnail8169:
good

squidoodly:
My problem with the muti-place is that it counts any crystal that are valid targets. This doesn't guarantee that a crystal CA is paused for is actually going to do any damage at all. It assumes that all placed crystals are placed by the player (which they aren't).

Unless there is some logic which I am missing this seems to be a bad solution and I would rather see a solution similar to held crystals to guarantee that what we are waiting for is something we have placed. Perhaps there could be some damage calculation also to make sure that what we are waiting for is worth waiting for.

zyktex:
I could modify the multi-place algorithm to only not place after you've actually hit a crystal until the crystal is destroyed. I could use a HashMap<Integer (entity id), Long (break time)> and check if the world still contains that entity, or if the break time is more than 100*timeout ms away.

squidoodly:
A HashMap seems excessive to hold a single value, you could just use a pair. If you can make something that holds a crystal that has been placed by CA then that will be fine. You can add the timer if you want (I assume that will be assigned the breakDelay time) to not have it paused for a long time. Like I said, a damage calculation or position check would be better if you want to not waste crystals and have a lower chance of desync but either way works.

I don't know if you want to add something to check for distance and walls distance as well. Could be useful to not be waiting forever for the crystal to time out.

zyktex:
@squidoodly that would not work when there are multiple crystals to be broken. Using a single pair is a hacky solution and when I tried it, it didn't work too well. Instead I implemented @seasnail8169 's original anti-multiplace idea, which works consistently

squidoodly:
I don’t see what the doesn’t work because the client doesn’t place multiple times in the same tick and so there will only ever be one crystal that is placed by us. Therefor, we only ever need to store one crystal at a time (assuming single place is on)

squidoodly:
I also really disagree with the break delay left being set to 0 once a crystal is broken. That means that there will essentially never be any break delay because CA will break a crystal, set the delay to whatever it should be, then the removed event will fire and because the crystal will always be inside the break range, it will set the break delay to 0 and so CA will go to break a crystal the very next tick. This means that no matter what the delay is, if you have single place on then you will always break every tick.

If I’m missing something then let me know because I am on mobile right now but from what I can see this is a terrible way to make this and is barely a solution.

zyktex:
Firstly, although the first statement applies when you're the only one placing crystals, it here are other players, there could be multiple crystals being broken. Secondly, the crystal removed event is only fired when the server sends the destroy entites packet; therefore, delay will only be set to 0 when there are no crystals. There is an edge case when break range is greater than 6; however, I can make a Math.min check if you want.

zyktex:
Also, since you and @seasnail8169  have opposing views on this, could we re-add you to the group chat so it's easier to agree on something

PR #223, opened 2021-02-04
KillAura: target multiple

JFronny:
Allows the killaura module to target multiple entities each tick. This is useful when farming (for example) slimes or magma cubes and might increase dps in environments where lower damage because of high attack speed is not a thing.

### DIFF
--- a/src/main/java/minegame159/meteorclient/modules/combat/KillAura.java
+++ b/src/main/java/minegame159/meteorclient/modules/combat/KillAura.java
@@ -128,6 +128,13 @@ public class KillAura extends Module {
             .build()
     );
 
+    private final Setting<Boolean> targetMultiple = sgGeneral.add(new BoolSetting.Builder()
+            .name("target-multiple")
+            .description("Target multiple entities at once")
+            .defaultValue(false)
+            .build()
+    );
+
     // Rotations
 
     private final Setting<RotationMode> rotationMode = sgRotations.add(new EnumSetting.Builder<RotationMode>()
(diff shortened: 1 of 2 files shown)


asdffdsdaf:
mfw merge :scream:

MineGame159:
You basically just removed every delay check for multi mode? Why? In my opinion the delays should still apply, but of course hit all of them at once.

JFronny:
Delays still work, SmartDelay is implemented [here](https://github.com/MeteorDevelopment/meteor-client/pull/223/files#diff-ce5f3d591c2e63e66dab7e0f9de0fec528bbe3db8dd7163bc7b78b47f1dca026R248) for MultiTarget, RandomDelay was moved outside of attack and is always applied the same way

MineGame159:
Mhm, but the whole delay system now just looks scuffed. For example why is smart delay duplicated? Can't you just put every delay check into one place.

MineGame159:
In entity utils make the "entities" static field private and make "getAll" take a list it will add the entities to, having to access the entities field is just a bad practise. And for "get" you can just pass the private static entities field.

MineGame159:
Can you fix conflicts?

JFronny:
According to GitHub conflicts have already been fixed in 3a3d0ae1936e9e841b8e16ac7031bdf41eb60d19

MineGame159:
Well for me they are still there

JFronny:
Are the two commits after a5619bdcad53056d7e982eafe1ad4642fcd6cb24 visible to you?

MineGame159:
Yes

PR #232, opened 2021-02-10
AutoAnvil update && BugFix

TechAle:
BugFix:
- You were not able to place a block on your enemy (like webs and buttons)

AutoAnvil:
- Added antiStep
- Now it's used a different algorithm for the height of the anvil. Now it's dynamic, it will be placed based on your distance from the enemy (and from the paramers you choosen)


MineGame159:
Can you resolve conflicts and move `private final ArrayList<Vec3d> antiStepStructure` above constructor? Fields shouldnt be below.

MineGame159:
image
Use BlockPos.Mutable to not create new BlockPos every time, just store one inside a final field and modify it.

seasnail8169:
Or alternatively, in that method, you could use pos.north(), pos.south(), etc which would give you the same thing.

TechAle:
Uh, we i'll be free, i'll try, and i'll resolve everything

MineGame159:
No, dont use .north() etc, they still create a new BlockPos instance.

TechAle:
something is wrong with my commit, why are speed, elytraplus and drawside added? bruh

MineGame159:
Idk but they are still there and have been changed so I cannot merge it without overwriting them.

seasnail8169:
> Idk but they are still there and have been changed so I cannot merge it without overwriting them.

i moved them

TechAle:
So, now this should be fine since snail moved them?

PR #244, opened 2021-02-13
Made NoSwing setting in CrystalAura better.

supakeks:
Changed the name to "Swing" and moved it to Render group, as what it does now is just change if you can see your swing.

the swingHand(hand) method renders the swing and sends the swing packet, so if it's off it'll only send the packet to avoid getting flagged for "NoSwing" on most strict anticheat servers.


seasnail8169:
ok keks

supakeks:
mad?

PR #246, opened 2021-02-14
Bruh how is this not a module already (Ghosthand)

C10udburst:
Opening chests through walls and stuff


MineGame159:
1. Use `mc.interactionManager.reachDistance` instead of hardcoding it
2. Make `posList` a private final field and clear it at the end of each tick
3. Why is every word in the description capitalized?

MineGame159:
It is currently looping 50 times which equals to about 5 reach distance, shouldn't that be adjusted dynamically based on the reach distance?

C10udburst:
Damn, sorry I'm stupid.

PR #251, opened 2021-02-14
Filter anti-cheat bots by gamemode

Mersid:
Gets rid of dumb shit like this by eliminating players with invalid gamemodes (note the bots with randomly generated names).

If someone finds a server that bypasses this maybe add a way to check if the user is in the players list or something.

### DIFF
--- a/src/main/java/minegame159/meteorclient/modules/render/Nametags.java
+++ b/src/main/java/minegame159/meteorclient/modules/render/Nametags.java
@@ -161,6 +161,13 @@ public class Nametags extends Module {
             .build()
     );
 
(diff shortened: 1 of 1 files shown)


seasnail8169:
@Mersid i feel like this could be better implemented as an anti bots module

Mersid:
That could work too. How would you propose we implement that? We could write another module, and from Nametags, check if that anti bots module is active, and if it is, we don't render the bot nametags?

seasnail8169:
No, basically just if the entity is a bot, remove it from the world, therefore nametags/tracers wont see them and neither will stuff like kill aura.

PR #252, opened 2021-02-14
Added AntiAnchor

TechAle:
Basically, if someone place an anchor above our head, it will crouch and place a slab between your head and the anchor.
This is going to reduce the damage since the slab is going to absorbe the damage.

### DIFF
--- a/src/main/java/minegame159/meteorclient/modules/combat/AntiAnchor.java
+++ b/src/main/java/minegame159/meteorclient/modules/combat/AntiAnchor.java
@@ -1,0 +1,89 @@
+/*
+ * This file is part of the Meteor Client distribution (https://github.com/MeteorDevelopment/meteor-client/).
+ * Copyright (c) 2021 Meteor Development.
+ */
+package minegame159.meteorclient.modules.combat;
+
+import meteordevelopment.orbit.EventHandler;
+import minegame159.meteorclient.events.world.TickEvent;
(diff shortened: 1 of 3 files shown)


src/main/java/minegame159/meteorclient/modules/combat/AntiAnchor.java
```diff
@@ -0,0 +1,89 @@
...
+
+                // Check the rotation
+                rotateCheck(slot);
+            }
+
+        }
+
+    }
+
+    public int getFloorSlot() {
```
seasnail8169, line 62: use InvUtils.findItemInHotbar method for this

TechAle: (Block.getBlockFromItem(item) instanceof net.minecraft.block.SlabBlock)
findItemInHotbar allow to search if an item is in a class like this?

seasnail8169: yes, you can use a predicate:

InvUtils. findItemInHotbar(itemstack -> itemstack.getItem() instanceof SlabBlock);

something like this (it may mot be 100% correct but its an example)

TechAle: i didn't know this could even work, i take note because it's interesting

PR #253, opened 2021-02-15
Added TNT fuse countdown to Nametags

C10udburst:


seasnail8169:
@C10udburst can you make this a setting please? like the others

MineGame159:
@seasnail8169 it already is a setting, only renders it when tnt is selected.

PR #259, opened 2021-02-18
Added AntiBot

Mersid:
Deletes PlayerEntities if they have no gamemode (are likely a bot). Has a toggle to only do so for invisible bots, which might be useful on servers that use fake players for things like lobby selection.


seasnail8169:
seasnail = 🐒

PR #260, opened 2021-02-18
Added api opt-out setting

C10udburst:
I don't think that telemetry should be mandatory.

### DIFF
--- a/src/main/java/minegame159/meteorclient/gui/screens/topbar/TopBarConfig.java
+++ b/src/main/java/minegame159/meteorclient/gui/screens/topbar/TopBarConfig.java
@@ -61,6 +61,15 @@ public class TopBarConfig extends TopBarWindowScreen {
                 .build()
         );
 
+        sgGeneral.add(new BoolSetting.Builder()
+        .name("send-data-to-api")
+        .description("If checked, your UUID will be send to Meteor's servers.")
+        .defaultValue(Config.get().sendDataToApi)
+        .onChanged(aBoolean -> Config.get().sendDataToApi = aBoolean)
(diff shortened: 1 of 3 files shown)


MineGame159:
There is literally nothing we can do with your UUID

C10udburst:
Yeah, I know, but it's good practice.

MineGame159:
It can't cause any harm, we can't use it for anything malicious, it doesn't slow the client (it has been there for a good few months), literally the only purpose of it is to make the client better.

C10udburst:
What if Hypixel for example abuses the `http://meteorclient.com/api/online/usingMeteor` endpoint to autoban everyone using Meteor?

MineGame159:
🤔 I didn't think of that

MineGame159:
Ok but make it so when the setting is disable it still sends the ping but without the ?uuid parameter.

src/main/java/minegame159/meteorclient/gui/screens/topbar/TopBarConfig.java
```diff
@@ -61,6 +61,15 @@ protected void initWidgets() {
                 .build()
         );
 
+        sgGeneral.add(new BoolSetting.Builder()
+                .name("send-data-to-api")
+                .description("If checked, your UUID will be send to Meteor's servers.")
+                .defaultValue(Config.get().sendDataToApi)
```
MineGame159, line 67: Also this should be true, not the value that it is currently set to

C10udburst: ah, ok

PR #271, opened 2021-02-25
now Fast Use looks cool

EurekaMan:


src/main/java/minegame159/meteorclient/modules/player/FastUse.java
```diff
@@ -8,41 +8,29 @@
...
+import net.minecraft.item.*;
 
 public class FastUse extends Module {
-    public enum Mode {
+    public enum Item {
         All,
-        Some
+        Exp,
+        Crystal,
+        ExpAndCrystal
```
supakeks, line 25: I'd recommend changing this to "Both" tbf

src/main/java/minegame159/meteorclient/modules/player/FastUse.java
```diff
@@ -8,41 +8,29 @@
...
-            .build()
-    );
-
-    private final Setting<Boolean> blocks = sgGeneral.add(new BoolSetting.Builder()
-            .name("blocks")
-            .description("Fast-places blocks if the mode is \"Some\".")
-            .defaultValue(false)
+    private final Setting<FastUse.Item> itemChoose = sgGeneral.add(new EnumSetting.Builder<FastUse.Item>()
+            .name("Which item")
+            .description(".")
```
seasnail8169, line 32: dont have an empty description like this

src/main/java/minegame159/meteorclient/modules/player/FastUse.java
```diff
@@ -8,41 +8,29 @@
...
-            .defaultValue(false)
-            .build()
-    );
-
-    private final Setting<Boolean> blocks = sgGeneral.add(new BoolSetting.Builder()
-            .name("blocks")
-            .description("Fast-places blocks if the mode is \"Some\".")
-            .defaultValue(false)
+    private final Setting<FastUse.Item> itemChoose = sgGeneral.add(new EnumSetting.Builder<FastUse.Item>()
+            .name("Which item")
```
seasnail8169, line 31: this is not a proper setting name, it should be "item"

src/main/java/minegame159/meteorclient/modules/player/FastUse.java
```diff
@@ -8,41 +8,29 @@
...
+            .description(".")
+            .defaultValue(FastUse.Item.All)
```
seasnail8169, line 33: you do not need to reference the enum like this, it can just be Item.All

src/main/java/minegame159/meteorclient/modules/player/FastUse.java
```diff
@@ -8,41 +8,29 @@
...
-            .description("Fast-throws XP bottles if the mode is \"Some\".")
-            .defaultValue(false)
-            .build()
-    );
-
-    private final Setting<Boolean> blocks = sgGeneral.add(new BoolSetting.Builder()
-            .name("blocks")
-            .description("Fast-places blocks if the mode is \"Some\".")
-            .defaultValue(false)
+    private final Setting<FastUse.Item> itemChoose = sgGeneral.add(new EnumSetting.Builder<FastUse.Item>()
```
seasnail8169, line 30: again this reference can just be Item not FastUse.Item

PR #274, opened 2021-02-27
AutoPot V3 Ez

GildedS:
Took 5 hours 32 minutes. Pls accept


Wide-Cat:
inb4 its ignored like the other one

Neux0:
Potion of turtle master is missing from autopot

PR #283, opened 2021-03-06
Slight GUI changes

C10udburst:
Added icons to modules. Disabled by default
screenshot


MineGame159:
I don't like how you implemented it. First of all you have a `getIcon()` method that checks for the category, why not put a field into the `Category` class itself and set it through the constructor. That way addons can provide icons. And in multiple widget classes you just duplicated the entire constructor's code with one line changed, cant you avoid the duplication by create a shared method?

seasnail8169:
Also this would probably just be best to wait for after the planned gui rewrite happens, no?

MineGame159:
Its fine, its not really a big change.

PR #284, opened 2021-03-06
Added ArrowDodge

C10udburst:


MineGame159:
image
Can you please move those below the settings. :sob: There are a few things that I would do differently for performance but whatever.

MineGame159:
Don't move the enum btw

C10udburst:
And what about the List? Leave at the top or also move it?

MineGame159:
Yeah move the list too

PR #286, opened 2021-03-07
Added .locate

C10udburst:
Syntax just like vanilla /locate.
Uses maps' NBT data, Baritone's chunk scanning and ender eyes to locate different structures.


C10udburst:
Nwm don't merge yet. I forgot to unsubscribe when a stronghold was found

PR #291, opened 2021-03-10
Delete Monkey

RickyTheRacc:
Who even used this anyways?


supakeks:
???????????

supakeks:
Also this would fail because in the ModuleManager it would be trying to add a class that doesn't exist

PR #309, opened 2021-03-16
Added separate module for AntiBreak

StormyBytes:
On default, the option is True.
When off, it just does anti-break, doesn't pick the best tool.

I've been having issues with ghost-blocks (blocks breaking on client side but not on server side) because of this. So i added an option to use Best Tool or not.
I also sometimes just want the anti break and not the best tool.


ekcire06:
but the whole point of autotool is that it switches to the best tool...

I think it would be better to make antibreak a seperate module instead of adding an autotool setting to autotool

StormyBytes:
> but the whole point of autotool is that it switches to the best tool...
> 
> I think it would be better to make antibreak a seperate module instead of adding an autotool setting to autotool

Hmm, yeah, i think its better to just make it a seperate module.

PR #311, opened 2021-03-17
KillAura: Fixed Always rotations and added AutoSwitch

supakeks:
Makes KA work on 2b2t now (if u use "always" in rotation mode)

Also added autoswitch bc Its really good and cool


seasnail8169:
COPE

PR #313, opened 2021-03-19
Added map preview

C10udburst:
image


MineGame159:
Enable bleding so the transparent pixels at the edges are actually transparent and not black.

PR #360, opened 2021-03-28
Added Player Coordinates to PlayersHUD and fixed a bug.

StormyBytes:
Player Coordinates Preview:
image

The bug was, when there isnt any other player (other than self) in render distance, the PlayersHUD bounding box will have a width of 0. Meaning it cannot be moved or selected in the HUD editor.


seasnail8169:
Can you format them to whole numbers please.

StormyBytes:
Whoops i used `Double.toString` instead of `Long.toString`

seasnail8169:
I don't like the way you have done this tbh.

StormyBytes:
My apologies, I am quite new to minecraft modding. Is there any thing I could do to improve it?

supakeks:
> I don't like the way you have done this tbh.

what doesn't? the code or how it looks in-game? since in-game I can't think of any other way to do it and this looks decent imo, would maybe make the coords color different so it's easy to know what's a username and what are coords

ekcire06:
im with seasnail on this one. it just seems unneccesary, and to use it youd have to cross reference it with your cords and figure out what direction your facing. would be too hard to use to be useful. tracers are already perfect for showing where a player is in relation to you, and are very intuitive to use.

RedCarlos26:
The bug fix should be implemented tough

MineGame159:
I would make the coords a different color than the names

seasnail8169:
The coordinates should use the secondary HUD color like almost every other HUD element. Also can you send a screenshot of how this looks with your changes.

seasnail8169:
The coordinates should use the secondary HUD color like almost every other HUD element. Also can you send a screenshot of how this looks with your changes.

StormyBytes:
image
might need to implement antibot to playershud...

seasnail8169:
That looks awful, i will rewrite it when i get home.

PR #361, opened 2021-03-28
OptiFine Capes

fluse1367:
- Renders OptiFine capes for players (if the player has one ofc).
 - Doesn't check a player if it already has a meteor cape.
 - OptiFine cape editor session generator ("Open OptiFine Cape Editor" button in skin option menu).

### DIFF
--- a/src/main/java/minegame159/meteorclient/MeteorClient.java
+++ b/src/main/java/minegame159/meteorclient/MeteorClient.java
@@ -30,6 +30,7 @@ import minegame159.meteorclient.utils.misc.input.KeyBinds;
 import minegame159.meteorclient.utils.network.Capes;
 import minegame159.meteorclient.utils.network.MeteorExecutor;
 import minegame159.meteorclient.utils.network.OnlinePlayers;
+import minegame159.meteorclient.utils.network.OptifineCapes;
 import minegame159.meteorclient.utils.player.EChestMemory;
 import minegame159.meteorclient.utils.player.Rotations;
 import minegame159.meteorclient.utils.render.color.RainbowColors;
@@ -134,6 +135,7 @@ public class MeteorClient implements ClientModInitializer {
     @EventHandler
     private void onTick(TickEvent.Post event) {
         Capes.tick();
+        OptifineCapes.tick();
 
         if (screenToOpen != null && mc.currentScreen == null) {
             mc.openScreen(screenToOpen);
--- a/src/main/java/minegame159/meteorclient/mixin/AbstractClientPlayerEntityMixin.java
+++ b/src/main/java/minegame159/meteorclient/mixin/AbstractClientPlayerEntityMixin.java
@@ -7,6 +7,7 @@ package minegame159.meteorclient.mixin;
 
 import minegame159.meteorclient.utils.misc.FakeClientPlayer;
 import minegame159.meteorclient.utils.network.Capes;
+import minegame159.meteorclient.utils.network.OptifineCapes;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.network.AbstractClientPlayerEntity;
 import net.minecraft.client.network.PlayerListEntry;
@@ -22,6 +23,7 @@ public class AbstractClientPlayerEntityMixin {
     @Inject(method = "getCapeTexture", at = @At("HEAD"), cancellable = true)
     private void onGetCapeTexture(CallbackInfoReturnable<Identifier> info) {
         Identifier id = Capes.get((PlayerEntity) (Object) this);
+        if(id == null) id = OptifineCapes.get((PlayerEntity) (Object) this);
         if (id != null) info.setReturnValue(id);
     }
 
--- a/src/main/java/minegame159/meteorclient/mixin/ScreenMixin.java
+++ b/src/main/java/minegame159/meteorclient/mixin/ScreenMixin.java
@@ -8,17 +8,13 @@ package minegame159.meteorclient.mixin;
 import minegame159.meteorclient.modules.Modules;
 import minegame159.meteorclient.modules.render.NoRender;
 import minegame159.meteorclient.utils.Utils;
-import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.gui.screen.Screen;
 import net.minecraft.client.gui.widget.AbstractButtonWidget;
-import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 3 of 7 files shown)


MineGame159:
Why did you make `SkinOptionsScreenMixin` exten `ScreeMixin`. I don't think thats a good practise.

fluse1367:
Because DisconnectedScreenMixin does the same.

MineGame159:
Oh it does? Well gonna change that at some point.

fluse1367:
I also to access to the `buttons` and the `client` field. I'm not that much into mixins, is there another way to obtain these fields? Maybe a acessor mixin?

MineGame159:
You can just put them into the `SkinOptionsScreenMixin` itself.

fluse1367:
Using `@Shadow`? That doesnt work. The Client crashes with this error:
```
org.spongepowered.asm.mixin.transformer.throwables.InvalidMixinException: @Shadow method addButton in meteor-client.mixins.json:SkinOptionsScreenMixin was not located in the target class net.minecraft.client.gui.screen.options.SkinOptionsScreen. No refMap loaded.
```
I'm just going to use an accessor then.

MineGame159:
No, dont use accessor. The `SkinOptionsScreenMixin` needs to extend `Screen` (not the mixin, just the normal class) for you to be able to access fields / methods from it. Also you should get mc dev plugin for intellij if you dont have it already.

fluse1367:
Ohhh i tought i wouldn't work bc i had to super the constructor from Screen, but it works just fine!
Also, i already have the mc dev plugin, ty ^^

seasnail8169:
I don't like this, I don't see any reason for meteor to show optifine capes. I am going to close this and it can be reopened if MineGame changes his mind.

Wide-Cat:
bruh

RedCarlos26:
Why seeing optifine capes and meteor capes would be POG

fluse1367:
Agreed.

seasnail8169:
But there is literally no reason to do this. Meteor doesn't even support optifine... there is literally mo reason for us to show a cape that someone has given other people money to receive. The reason we show capes is to show gratitude to the people who support us and this literally defeats the point of that.

fluse1367:
`But there is literally no reason to do this.`
Yes there is? Capes are cool and i wanna see other player's capes. And i bet other players wanna see them too.

`The reason we show capes is to show gratitude to the people who support us and this literally defeats the point of that.`
As I said eariler, if a player has a meteor cape it is shown instead of the optifine one.

@seasnail8169  I think you shouldn't be even able to say "no" to any PR out of your personal liking (no offense btw).
Meteor is still a Open Source project (protected by the GPL-v3 btw), so when you reject a PR bc of your personal liking i'm wondering why you have put meteor open src in the first place. It's supposed to be a community driven project and I think it is just not fair to reject PRs because you don't like them.

I would understand, if there would be a downside that comes with this PR, but there is none. literally.

seasnail8169:
I understand your point but its not just my preference, i talked to minegame about it too. Bear in mind that we are the owners of the project and, whilst yes it is open source, we still do decide what is accepted and what is not. A project being open source does not mean that decisions are made on a community level.

seasnail8169:
Also the downside is that it does not mean so much to donate to meteor. As you can imagine with the project being opensource, there is not much of a reason to donate. One of the few benefits we offer is being able to have a cape that is displayed to all meteor users, as a way of showing our gratitude for people supporting us. Allowing people to have a cape by other means with absolutely no benefit for us feels, to me, like it reduces the benefits of donating even more. Also, it just seems weird to me that we would display something that is literally got by giving someone else money.

MineGame159:
I agree with everything snail has said above

MineGame159:
Also because OptiFine is not compatible with Meteor (and pratically 90% of Fabric mods) then the capes support for it can easily be made into a separate Fabric mod that can be used alongside any other client / mod.

fluse1367:
The reason to donate is not to get a cape, it's to support the developers. The cape is just a "thank you". And the "thank you" cape is still being displayed.
If your intention with meteor is to earn money, then I really dont get why you made it open source.

`feels, to me, like it reduces the benefits of donating even more.`
It doesn't even reduce the benefits in any way.

`we would display something that is literally got by giving someone else money.`
Wouldn't it make more sense to be open minded and kinda support the optifine devs in this way? Also, someone who got an optifine cape is more likely to get an meteor cape anyways.

fluse1367:
> Also because OptiFine is not compatible with Meteor (and pratically 90% of Fabric mods) then the capes support for it can easily be made into a separate Fabric mod that can be used alongside any other client / mod.

True, but then anyone would have to install another mod that needs to be maintained. Also this is a so small feature, an own mod for it would be kinda overdone.

seasnail8169:
> Wouldn't it make more sense to be open minded and kinda support the optifine devs in this way?

Huh? Why do we have any reason whatsoever to support the developer of optifine?

MineGame159:
`If your intention with meteor is to earn money, then I really dont get why you made it open source.`
Its not but keeping the servers running is not free.

`Wouldn't it make more sense to be open minded and kinda support the optifine devs in this way?`
Support OptiFine devs in what? In having a closed source mod that breaks every single mod there is and nobody can even fix it because we cant see the source code? OptiFine is currently nothing more than just a legacy mod that people are starting to not care about.

seasnail8169:
Just to add, neither me nor minegame have ever made any profit from the months we have put into this free, opensource project. We do not ask for money for our own benefit, purely for the benefit of our users (keeping the web server and mc server alive). I really don't understand why you have such a strong need to question our decisions in this situation.

fluse1367:
I dont wanna argume with u any longer tbh.
It's just sad that if one of you both doesn't like an addition/change you're just rejecting it without even considering if it'd be nice for the community.
This isnt the intention of an open source project. This is also why @Walaryne and me did tanuki in the first place.

MineGame159:
From your point of view everything you are saying is correct but you never owned something on this scale so its kinda hard exaplaining it to you but whatever. Good luck.

seasnail8169:
I don't understand what you mean to be honest. Meteor is open source because its simple and convenient, it makes no sense for us to protect our code. While we obviously accept contributions and changes, we do still own the project and therefore have the final say in what goes into our client. Being opensource does *not* mean that it is somehow owned by the community or whatever.

fluse1367:
As you said, you have to final say.

C10udburst:
just use [this mod](https://github.com/CaelTheColher/Capes). its compatible with meteor

PR #382, opened 2021-03-31
Added notebot

C10udburst:
fucking finally done


MineGame159:
1. Pressing stop still doesn't work while tuning noteblocks.
2. Rendering doesn't clear when pressing play button when a song is already playing.
3. When its playing and you press stop then yes, it makes the rendering invisible but does not clear the blocks. So once you try to play another song all of the previous rendering comes back up and just the new ones that it tuned.

PR #389, opened 2021-04-01
Fix BoundingBox

StormyBytes:
Fixed a bug where the PlayersHUD bounding box is 0 when there aren't other players around (no players in the PlayersHUD list). Meaning it cannot be selected/moved in the HUD editor.


StormyBytes:
Made a new pr so the bug fix can be implemented while seasnail works on rewriting the playershud coordinates

seasnail8169:
ok ill do it today

PR #409, opened 2021-04-05
Slight changes to .plugins and BetterTooltips

C10udburst:
- `.plugins` will now highlight anti ~~cheat~~ utility plugins with red
- Merged ItemByteSize with BetterTooltips
- Shulker previews can now be colored according to the shulker type
- added alias to `.inventory`


RedCarlos26:
This seems really nice

src/main/java/minegame159/meteorclient/mixin/HandledScreenMixin.java
```diff
@@ -181,7 +181,7 @@ private void drawMapPreview(MatrixStack matrices, ItemStack stack, int x, int y,
         buffer.vertex(x1, y1, z).texture(0.0f, 0.0f).next();
         tessellator.draw();
 
-        MapState mapState = FilledMapItem.getMapState(stack, mc.world);
+        MapState mapState = FilledMapItem.getOrCreateMapState(stack, mc.world);
```
MineGame159, line 184: Why? Why create the map state if it doesn't exist?

C10udburst: I though it might help with the issue, where maps dont render until viewed manually once, but it doesnt seem to do anything different

MineGame159: Oh

MineGame159: I think it might be just that the server didnt send the data yet because you havent held it yet. If thats the case then it cant really be fixed.

C10udburst: Huh. Maybe swapping the map with the selected slot for a single tick might fix that, but that seems like a rather ugly fix

PR #465, opened 2021-04-10
added a sync option to crystalaura

ekcire06:
sync is great and all, but it does slow down ca significantly if your ping is high. so I added an option for it. I also renamed cancel crystal to sync mode. its enabled by default btw.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,6 @@
 # Done to increase the memory available to gradle.
 org.gradle.jvmargs=-Xmx2G
+org.gradle.java.home=C:\\Program Files\\Java\\jdk-11.0.1
 
 # Fabric Properties
 	# check these on https://fabricmc.net/use
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,5 @@
 # Done to increase the memory available to gradle.
 org.gradle.jvmargs=-Xmx2G
-org.gradle.java.home=C:\\Program Files\\Java\\jdk-11.0.1
 
 # Fabric Properties
 	# check these on https://fabricmc.net/use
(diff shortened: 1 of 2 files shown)


supakeks:
What is the point of this

MineGame159:
On an unrelated note, what the actual fuck is that formatting

ekcire06:
the point of this is because sync (even on hit mode) can slow down crystalaura (by about 1/3rd in my tests at 130 ping, I tested it by seeing how long it took to eat through a stack of crystals, and then reloged after to make sure there were no ghostplaces.), so I just added an option to use the old desyncy but faster crystalaura. 

sorry about the formatting lmao, im kinda new to this.

supakeks:
There is literally 0 speed difference, it looked faster before because the crystals weren't synced, making it look faster when looking at the crystals in your hotbar

ekcire06:
what ping are you testing at? I tried to ensure that wasnt a factor by measuring how long it took to use a set amount of crystals instead of looking at how fast the numbers were moving in hotbar...

supakeks:
130
The OnHit and OnSound have nothing to do with syncing, all it's for is when to remove the crystal from the world

and again, there's 0 speed difference with the new syncing

ekcire06:
but the crystal has to be removed for it to try to place another one where the previous one was

supakeks:
Yes and that's exactly what CrystalAura does with OnSound

OnHit is stupid and breaks ca on ncp servers, I don't know why it's still an option, there's a way to remove the crystal on sound without breaking ca but I still can't figure that out on 1.16

ekcire06:
I tested this on avas.cc, a server without ncp or any other ac... that probably explains the difference in results we are getting.

supakeks:
OnHit is just bad, there's ways to make ca even faster than onHit without actually desyncing, plus when you attack a crystal it removes from the world (server side) after 1 tick afaik, and since onhit removes it from the world (client side) as soon as you hit it it just creates desync

ekcire06:
how about I make so instead of a sync setting I make it a "instantly" option of cancelcrystal and put a warning in the description that it will cause signifigant desync?

ekcire06:
for some reason github is being weird with my formatting. shows up as normal and sensible in github desktop and my ide, but messes it up when I commit it. im gonna try to fix it.

seasnail8169:
jesus christ please look up java formatting standards

ekcire06:
Screenshot (325)

what it looks like in my ide vs on github lmao

ekcire06:
oops. didnt mean to commit that gradle thing lmao

ekcire06:
im just gonna close this garbage

PR #480, opened 2021-04-12
Bow Aimbot

StormyBytes:
Pretty self explanatory.

### DIFF
--- a/src/main/java/minegame159/meteorclient/systems/modules/combat/BowAimbot.java
+++ b/src/main/java/minegame159/meteorclient/systems/modules/combat/BowAimbot.java
@@ -25,6 +25,7 @@ import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.passive.AnimalEntity;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.BowItem;
+import net.minecraft.item.CrossbowItem;
 import net.minecraft.item.Items;
 import net.minecraft.util.math.Vec3d;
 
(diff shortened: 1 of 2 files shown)


src/main/java/minegame159/meteorclient/systems/modules/combat/BowAimbot.java
```diff
@@ -0,0 +1,181 @@
...
+        target = null;
+        wasPathing = false;
+    }
+
+    @EventHandler
+    private void onTick(TickEvent.Pre event) {
+        if (mc.player.isDead() || !mc.player.isAlive() || !itemInHand()) return;
+
+        // Target
+        if (InvUtils.findItemWithCount(Items.ARROW).slot == -1) return;
```
seasnail8169, line 105: I think you should be using `InvUtils.findItemInHotbar` for this check since `findItemWithCount` searches your whole inv iirc.

StormyBytes: Wouldn't that mean you have to have the arrows be in your hotbar for it to work?
Imo it would be annoying to have to place the arrows into the hotbar everytime.

Also, I was following the BowSpam module for that.
```
    private void onTick(TickEvent.Post event) {
        if (InvUtils.findItemWithCount(Items.ARROW).slot == -1) return;
        if (!onlyWhenHoldingRightClick.get() || mc.options.keyUse.isPressed()) {
```

seasnail8169: i shouldn't comment just after i wake up, sorry

PR #514, opened 2021-04-16
Added Bow Aimbot

StormyBytes:


MineGame159:
@StormyBytes found one issue that it doesnt work for spiders or any entity thats nto 2 blocks high. It tries to shoot them as if they were 2 blocks high and then shoots above them. Apart from that it seems to work well

PR #540, opened 2021-04-20
Merge AimAssist and BowAimbot into a single module

StormyBytes:
To avoid confusion between the two, just merge them together. (Also imo its easier to just have 1 module do both)
Upon testing, the newly merged module works fine.


enescglyn:
Can you guys also merge Auto Tool and Auto Sword together? it doesnt make sense to have both of them seperate modules.

AstroOrbis:
> To avoid confusion between the two, just merge them together. (Also imo its easier to just have 1 module do both)
> Upon testing, the newly merged module works fine.

Aimbot aims the player crosshair. BowAim aims only while you are charging a bow, and works from a farther distance. 

> Can you guys also merge Auto Tool and Auto Sword together? it doesnt make sense to have both of them seperate modules.

This makes sense however. But shouldn’t AutoTool work as AutoSword as well? Or does it just take the most readily available sword, and not the best one?

PR #541, opened 2021-04-20
Added Anti-Xray Bypass

enescglyn:
Only shows blocks that have at least one face exposed to air thus bypassing anti xray plugins found on some servers.

### Tested on
Paper built-in anti xray engine mode 2

### Notes
Since the only ores exposed to air are in caves you will see less ores compared to not using this mode, although its better to have this setting than seeing this:
image
Also can you guys tell me if you guys give special role to contributors in Discord


enescglyn:
nvm doesnt work

PR #566, opened 2021-04-22
Fixed NoRender not not rendering entities and falling blocks

StormyBytes:


seasnail8169:
can you add the falling block check too, see the norender module just above the noEntity method

StormyBytes:
Sure

PR #578, opened 2021-04-24
add anchor option to nofall bucket mode

Wide-Cat:


Wide-Cat:
wait typo

MineGame159:
I dont think it should reset horizontal velocity every tick

PR #583, opened 2021-04-25
Notebot help screen

C10udburst:
Added help to notebot

### DIFF
--- a/src/main/java/minegame159/meteorclient/gui/screens/NotebotHelpScreen.java
+++ b/src/main/java/minegame159/meteorclient/gui/screens/NotebotHelpScreen.java
@@ -1,0 +1,58 @@
+package minegame159.meteorclient.gui.screens;
+
+import minegame159.meteorclient.MeteorClient;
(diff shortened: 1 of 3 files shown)


src/main/java/minegame159/meteorclient/gui/screens/NotebotHelpScreen.java
```diff
@@ -0,0 +1,58 @@
...
+import minegame159.meteorclient.gui.WindowScreen;
+import minegame159.meteorclient.gui.widgets.containers.WTable;
+
+public class NotebotHelpScreen extends WindowScreen {
+
+
+    public NotebotHelpScreen(GuiTheme theme) {
+        super(theme, "Notebot - Help");
+
+        WTable list = add(theme.table()).expandCellX().widget();
```
MineGame159, line 14: You don't need a table, the WindowScreen itself is a vertical list.

C10udburst: ok

PR #617, opened 2021-05-01
Additional LightOverlay checks

### LINKED ISSUE #585: [Bug] Light overlay ignores if the spawn platform is a transparent block
Light overlay ignores if the spawn platform is a transparent block
For eg:- light overlay will show mobs will spawn on glass

StormyBytes:
Resolves #585


seasnail8169:
wouldnt it be better to do some kind of overall check to see if its a block that no mobs can spawn on

StormyBytes:
As far as I know the only factors preventing mob spawning are
- If the block isnt a full block
- If the block is translucent
- And if the light level is below 8

Declipsonator:
Yeah, like if slabs are at the higher position they can have mobs spawn on them but if they're lower they can't

PR #639, opened 2021-05-03
Commands consistency

C10udburst:
Changed `.damage` and `.notebot` commands to use `SimpleCommandExceptionType`, to be more consistent with the rest. Also added `.item` alias to `.give`.

### DIFF
--- a/src/main/java/minegame159/meteorclient/systems/commands/commands/DamageCommand.java
+++ b/src/main/java/minegame159/meteorclient/systems/commands/commands/DamageCommand.java
@@ -20,7 +20,7 @@ import net.minecraft.text.LiteralText;
 import static com.mojang.brigadier.Command.SINGLE_SUCCESS;
 
 public class DamageCommand extends Command {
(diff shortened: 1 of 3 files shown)


src/main/java/minegame159/meteorclient/systems/commands/commands/DamageCommand.java
```diff
@@ -7,18 +7,21 @@
...
-import minegame159.meteorclient.utils.player.ChatUtils;
 import net.minecraft.command.CommandSource;
 import net.minecraft.network.packet.c2s.play.PlayerMoveC2SPacket;
 import net.minecraft.util.math.Vec3d;
+import net.minecraft.text.LiteralText;
 
 import static com.mojang.brigadier.Command.SINGLE_SUCCESS;
 
 public class DamageCommand extends Command {
+    private final static SimpleCommandExceptionType INVULNERABLE = new SimpleCommandExceptionType(new LiteralText("You are in invulnerable."));
```
MineGame159, line 23: Im pretty sure there is a typo, the `in` shouldn't be there. And yes I know it was like that before.

MineGame159 (changes requested):

PR #657, opened 2021-05-05
Made prefix open chat, like pressing / in vanilla

C10udburst:


src/main/java/minegame159/meteorclient/systems/commands/PrefixListener.java
```diff
@@ -0,0 +1,28 @@
+package minegame159.meteorclient.systems.commands;
+
+import meteordevelopment.orbit.EventHandler;
+import minegame159.meteorclient.MeteorClient;
+import minegame159.meteorclient.events.meteor.CharTypedEvent;
+import minegame159.meteorclient.systems.config.Config;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.gui.screen.ChatScreen;
+
+public class PrefixListener {
```
seasnail8169, line 10: This really doesnt need to be its own class, just put the listener in the already subscribed MeteorClient class

src/main/java/minegame159/meteorclient/systems/commands/PrefixListener.java
```diff
@@ -0,0 +1,28 @@
...
+    public static void init() {
+        MeteorClient.EVENT_BUS.subscribe(PrefixListener.class);
+    }
+
+    @EventHandler
+    private static void onCharTyped(CharTypedEvent event) {
+        if (mc.currentScreen != null) return;
+        if (!config.openChatOnPrefix) return;
+
+        if (event.c == config.prefix.charAt(0)) {
```
seasnail8169, line 23: What if someone has a prefix that is multiple characters. This is the reason why i never implemented this because it just doesn't make sense if you have a prefix thats more than a single character

MineGame159:
This will only work for some keys, keys like control, alt and others won't work.

seasnail8169:
I feel inclined to close this. I think the only way it could work is having a keybind setting to open chat with the prefix already written.

src/main/java/minegame159/meteorclient/systems/commands/PrefixListener.java
MineGame159: If the prefix is longer then it shouldn't check for the key and do nothing.

MineGame159:
No.

src/main/java/minegame159/meteorclient/systems/commands/PrefixListener.java
C10udburst: I made it toggleable, so if someone doesn't want it they can disable it

MineGame159:
Also you shoud probably check if you are ingane. I don't remember now but I think the event is fired when in main menu too.

C10udburst:
> Also you shoud probably check if you are ingane. I don't remember now but I think the event is fired when in main menu too.

I check it with `if (mc.currentScreen != null) return;`

MineGame159:
True, ig that works.

BobIsMyManager:
yes this pog

seasnail8169:
what if my prefix is "IHateMinegame"???!?!!!!?? chat is gonna open when i type I 😞

seasnail8169:
this is stupid i hate it 
i hate you @MineGame159

C10udburst:
> what if my prefix is "IHateMinegame"???!?!!!!?? chat is gonna open when i type I 😞

You can disable this feature

seasnail8169:
No shit but thats not the point, its so retarded to use a character rather than a keybind. 

Character means it has to be the first char of your prefix which could require modifiers to enter (i.e captial letters or . or ) which all require shift which is just inconvenient. 

Keybind means it can be any key you want and wouldn't require modifiers which is just so much simpler and makes so much more sense to use.

C10udburst:
You can always just change the vanilla open chat keybind.

Also many other clients just similar method to this. Here are some random examples I found: [seppuku](https://github.com/seppukudevelopment/seppuku/blob/master/src/main/java/me/rigamortis/seppuku/impl/module/hidden/CommandsModule.java), [KAMI](https://github.com/zeroeightysix/KAMI/blob/fabric/src/main/kotlin/me/zeroeightsix/kami/feature/hidden/KamiPrefixChat.kt), [Gish](https://github.com/GishReloaded/Gish-Code-1.12.2/blob/master/src/main/java/i/gishreloaded/gishcode/managers/CommandManager.java), [postman](https://github.com/moomooooo/postman/blob/master/src/main/java/me/srgantmoomoo/postman/client/command/CommandManager.java), [bedroom](https://github.com/beach-house-development/bedroom/blob/fb13aa437d783e6bf1f84f843161448ea45b4eda/src/main/java/me/srgantmoomoo/bedroom/command/CommandManager.java), [fabric-client](https://github.com/SleepyNeptune/fabric-client/blob/1.16.3/src/main/java/dev/toastmc/client/ToastClient.kt)

BobIsMyManager:
> image
> 
> You can always just change the vanilla open chat keybind.
> 
> Also many other clients just similar method to this. Here are some random examples I found: [seppuku](https://github.com/seppukudevelopment/seppuku/blob/master/src/main/java/me/rigamortis/seppuku/impl/module/hidden/CommandsModule.java), [KAMI](https://github.com/zeroeightysix/KAMI/blob/fabric/src/main/kotlin/me/zeroeightsix/kami/feature/hidden/KamiPrefixChat.kt), [Gish](https://github.com/GishReloaded/Gish-Code-1.12.2/blob/master/src/main/java/i/gishreloaded/gishcode/managers/CommandManager.java), [postman](https://github.com/moomooooo/postman/blob/master/src/main/java/me/srgantmoomoo/postman/client/command/CommandManager.java), [bedroom](https://github.com/beach-house-development/bedroom/blob/fb13aa437d783e6bf1f84f843161448ea45b4eda/src/main/java/me/srgantmoomoo/bedroom/command/CommandManager.java), [fabric-client](https://github.com/SleepyNeptune/fabric-client/blob/1.16.3/src/main/java/dev/toastmc/client/ToastClient.kt)

also impact it updates very often

seasnail8169:
> You can always just change the vanilla open chat keybind.

How is that related?

PR #675, opened 2021-05-06
Allow ESC on HUD screen to go to previous tab

JFronny:
This simple change allows pressing ESC in the HUD tab to open the previous tab instead of closing the click GUI as you cant navigate from there using the tabs


BobIsMyManager:
yes

BobIsMyManager:
yus

PR #707, opened 2021-05-09
Added a delay check to killaura

### LINKED ISSUE #706: [Bug] The new killaura settings dont take weapon delay into account
**Describe the bug**
Both of the new killaura settings, 'only on click' and 'only on look' don't take weapon delay into account, even when smart delay is turned on. The result is essentially spam clicking that does little damage. Latest dev build, prob due to trigger being merged with ka

**To Reproduce**
1. Go to killaura settings
2. Enable either 'only on click' or 'only on look'
3. Try to attack an entity

StormyBytes:
Resolves #706


Wide-Cat:
ty ❤️

StormyBytes:
Np :D

PR #725, opened 2021-05-11
Improve KillAura

JFronny:
This contains the following improvements to KillAura from my fork:
- Fully disabled in spectator mode
- "Random Teleport" Option, similar to Wursts TP Aura
- ~Option to automatically rotate the clients viewport towards targets (useful in combination with onlyOnClick)~

I created a single PR for all of these because isolating changes would be bothersome. If you don't want some of these changes I can remove them.


MineGame159:
There is Aim Assist which is designed for rotating the view client side. In my opinion Kill Aura shouldn't do that.

PR #732, opened 2021-05-13
Word wrap book bot

mvrozanti:
Adds word-wrap functionality to book bot. Turns out [the word wrapping problem](https://www.tutorialspoint.com/Word-Wrap-Problem) is kinda tricky, so I just lazily added [this neat little library](https://github.com/davidmoten/word-wrap) which makes book bot behave exactly as expected.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,6 +52,7 @@ dependencies {
 	include "com.github.MeteorDevelopment:baritone:${project.minecraft_version}-SNAPSHOT"
 
 	// Other dependencies
+	dependency "com.github.davidmoten:word-wrap:${project.wordwrap_version}"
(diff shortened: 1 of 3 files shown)


mvrozanti:
I just noticed it crashes on single-page books, so I'll work on another commit.

MineGame159:
Word wrapping is not tricky at all and definitely does not need a library for it. I have done it in [WMultiLabel](https://github.com/MeteorDevelopment/meteor-client/blob/master/src/main/java/minegame159/meteorclient/gui/widgets/WMultiLabel.java). If you rewrite it without the librarby then feel free to reopen this issue or make a new one.

mvrozanti:
I was looking to avoid the greedy approach, but if you think that's okay, sure.

PR #741, opened 2021-05-15
Added anchor option for Surround

### LINKED ISSUE #695: [SUGGESTION] add anchor fot surround
https://i.imgur.com/I0QNQoX.png

https://imgur.com/I0QNQoX

StormyBytes:
Resolves #695

### DIFF
--- a/src/main/java/minegame159/meteorclient/systems/modules/combat/Surround.java
+++ b/src/main/java/minegame159/meteorclient/systems/modules/combat/Surround.java
@@ -8,6 +8,7 @@ package minegame159.meteorclient.systems.modules.combat;
 import meteordevelopment.orbit.EventHandler;
 import minegame159.meteorclient.events.world.TickEvent;
 import minegame159.meteorclient.settings.BoolSetting;
+import minegame159.meteorclient.settings.EnumSetting;
 import minegame159.meteorclient.settings.Setting;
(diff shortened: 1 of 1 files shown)


src/main/java/minegame159/meteorclient/systems/modules/combat/Surround.java
```diff
@@ -80,10 +87,10 @@
             .build()
     );
 
-    private final Setting<Boolean> useEnderChests = sgGeneral.add(new BoolSetting.Builder()
-            .name("use-ender-chests")
-            .description("Will surround with ender chests if they are found in your hotbar.")
-            .defaultValue(false)
+    private final Setting<Block> block = sgGeneral.add(new EnumSetting.Builder<Block>()
```
C10udburst, line 90: Wouldn't using `ItemListSetting` make more sense? Similar to the way it's used in [Trajectories](https://github.com/MeteorDevelopment/meteor-client/blob/dccdc2e8e154c2cecefe3bc8b4eb302eb191ea49/src/main/java/minegame159/meteorclient/systems/modules/render/Trajectories.java#L39)

seasnail8169:
Why not just use a block list setting? Also why are you trying to add the entire minecraft item registry as the default? Its literally 3 block types.

PR #751, opened 2021-05-16
.tps command

C10udburst:


seasnail8169:
maybe `.server tps` instead

PR #790, opened 2021-05-19
Added Luminance to Fullbright, is now the default mode

Walaryne:
See what you think
Hope you guys like it :heart:

Wala~


Wide-Cat:
👍

PR #812, opened 2021-05-23
No Push module

hashalite:
- Added a No Push module
- Removed NoPush option from Velocity (it does the same thing as 'Blocks' under No Push)


MineGame159:
Uh to me it just seems like 2 modules are doing exactly the same thing. NoPush stops you from getting pushed by blocks, entities and liquids and Velocity can do the same. Well I know some poeple might want to get pushed by entities but still take knockback so why not change the boolean settings in Velocity and replace them with double setting so they have separate sliders? In my opinion that would be better than having 2 modules do the same just one has more customizability than the other.

hashalite:
I could make a Push section in Velocity with toggles and sliders for each type.  Would that work?

hashalite:
I would have to rework the knockback sliders to not affect pushing though

hashalite:
Sick

PR #815, opened 2021-05-23
No Cave Culling

hashalite:
- Adds a no cave culling option under no render

### DIFF
--- a/src/main/java/minegame159/meteorclient/mixin/ChunkOcclusionDataBuilderMixin.java
+++ b/src/main/java/minegame159/meteorclient/mixin/ChunkOcclusionDataBuilderMixin.java
@@ -7,6 +7,8 @@ package minegame159.meteorclient.mixin;
 
 import minegame159.meteorclient.MeteorClient;
 import minegame159.meteorclient.events.world.ChunkOcclusionEvent;
+import minegame159.meteorclient.systems.modules.Modules;
+import minegame159.meteorclient.systems.modules.render.NoRender;
 import net.minecraft.client.render.chunk.ChunkOcclusionDataBuilder;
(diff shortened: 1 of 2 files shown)


src/main/java/minegame159/meteorclient/mixin/ChunkOcclusionDataBuilderMixin.java
```diff
@@ -20,5 +22,6 @@
     private void onMarkClosed(BlockPos pos, CallbackInfo info) {
         ChunkOcclusionEvent event = MeteorClient.EVENT_BUS.post(ChunkOcclusionEvent.get());
         if (event.isCancelled()) info.cancel();
+        if (Modules.get().get(NoRender.class).noCaveCulling()) {info.cancel();}
```
MineGame159, line 25: Just above is an event so why didn't you use the event instead of hardcoding the module?

MineGame159 (changes requested):

src/main/java/minegame159/meteorclient/mixin/ChunkOcclusionDataBuilderMixin.java
hashalite: I used the event in my addon which adds a No Cave Culling module, but when I went to add the functionality to NoRender I saw that most of the work for other settings is done outside of the module class.  It was just an attempt at keeping the No Render class more uniform.  I can change it back if necessary

MineGame159: Yeah, other stuff in No Render are not events just because they are not needed by anything else.

hashalite:
Fixed

PR #872, opened 2021-05-30
Merge death position into waypoints module

JFronny:
This PR merges the death positions module into the waypoints module and adds a new texture for waypoints titled "skull" (based on nf-mdi-skull), which is used instead of the default texture for death positions


src/main/java/minegame159/meteorclient/systems/modules/render/WaypointsModule.java
```diff
@@ -33,16 +35,104 @@
...
+
+import static minegame159.meteorclient.utils.player.ChatUtils.formatCoords;
+
 public class WaypointsModule extends Module {
     private static final Color GRAY = new Color(200, 200, 200);
+    
+    private final SettingGroup sgDeathPosition = settings.createGroup("death-position");
+
+    private final Setting<Integer> dpWpAmouunt = sgDeathPosition.add(new IntSetting.Builder()
+            .name("wp-amount")
```
MineGame159, line 57: Please name it something more descriptive

src/main/java/minegame159/meteorclient/systems/modules/render/WaypointsModule.java
```diff
@@ -33,16 +35,104 @@
...
+
+            Waypoints.get().add(waypoint);
+        }
+        cleanDeathWPs(dpWpAmouunt.get());
+    }
+
+    private void cleanDeathWPs(int max) {
+        int oldWpC = 0;
+
+        ListIterator<Waypoint> wps = Waypoints.get().iteratorReverse();
```
MineGame159, line 125: Can't it just use normal iterator that goes from the start?

MineGame159 (changes requested):

src/main/java/minegame159/meteorclient/systems/modules/render/WaypointsModule.java
JFronny: The point is to keep the most recent waypoints and remove earlier ones. The normal iterator starts with older ones and new ones would be removed. In order to use the normal iterator I would need to do 2 iterations, one to find all death waypoints, then get the older ones from a list generated from that and finally iterate through that list to delete the older waypoints. This is simpler

src/main/java/minegame159/meteorclient/systems/modules/render/WaypointsModule.java
JFronny: The new name ("max-death-positions") should be more descriptive

src/main/java/minegame159/meteorclient/systems/modules/render/WaypointsModule.java
MineGame159: Correct me if I am wrong but the oldest waypoints are at the start of the list. So wouldn't it make sense to loop from the start to remove the oldest first?

JFronny: I want to keep a number of death waypoints so I need to know how many death waypoints will follow the current waypoint if it is a death waypoint.

MineGame159: Oh right, sorry.

PR #940, opened 2021-06-05
Added SpawnProofer

### LINKED ISSUE #719: [Suggestion] Auto spawn proof
**Describe the feature**
A module that if enabled would continually check for solid block surfaces within placing range and if any were found would automatically cover them in slabs if there were some in the inventory

**Screenshots**
Screenshot 2021-05-10 142021

would be for spawn-proofing biomes

**Additional info**
Add any other info.

### LINKED ISSUE #868: [Suggestion] Illuminate
**Describe the feature**
Places torches or any other block that emits light to spawn-proof an area. It could take advntage of the Light Overlay code to find the best placement.

Would be better if it had settings like:
- a filter list for blocks to use
- the ability to place blocks always or only in a specified zone
- whether to place the light sources on walls or floors only

**Additional info**
Could be added to the World category. Useful to spawn-proof big areas, such as those around mob farms.

### LINKED ISSUE #916: [Suggestion] I want a feature/module to mob proof an area.
**Describe the feature**
I want a module to automatically put torches, slabs, or carpets on the top of blocks in a specific area if the user is holding the items.

**Screenshots**
I have no screenshots of it. I just came up with this idea now and I don't know any other mods implementing this feature.

**Additional info**
If you implement this feature in meteor, I want it so that it will take the data from the "Light Overlay" module and then place torches instead of placing torches everywhere. But slabs and carpets are needed to be put everywhere to mob proof the area.
I also want it not to place torches, slabs, or carpets over blocks like stairs or others where it can't be placed.

StormyBytes:
Resolves #868, resolves #719, resolves #916

Good for spawn proofing huge areas for mob farms.
(Also better with LightOverlay)

### DIFF
--- a/src/main/java/minegame159/meteorclient/systems/modules/Modules.java
+++ b/src/main/java/minegame159/meteorclient/systems/modules/Modules.java
@@ -517,6 +517,7 @@ public class Modules extends System<Modules> {
         add(new StashFinder());
         add(new Timer());
         add(new VeinMiner());
+        add(new SpawnProofer());
     }
 
     private void initMisc() {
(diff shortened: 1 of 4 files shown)


C10udburst:
> Resolves #868, #719, #916
> 
put `Resolves #868, resolves #719, resolves #916`, so it will auto close all of them not just the first one

Wide-Cat:
Why is the max range 4, and max slider range only 1? Or am I missing something

PR #984, opened 2021-06-07
Added SpawnProofer and moved isValidMobSpawn() to BlockUtils

StormyBytes:


StormyBytes:
Yikes, i need to update to java 16

PR #1014, opened 2021-06-11
Merge EntityLogger into Notifier

StormyBytes:
- Merged Entity Logger into Notifier because they were basically the same.

- Fixed a bug where `event.entity` would be null and cause entities to not be rendered.


seasnail8169:
bruh @StormyBytes this is literally just pasted from my code

PR #1067, opened 2021-06-17
Resolves #1063

### LINKED ISSUE #1063: [Suggestion] Better spam delay
**Describe the feature**
Improve the delay setting of 'spam'. It doesn't allow for decimals, it only allows integers which is quite annoying when trying to configure it to match the antispam limit as closely as possible. 
Either expand it so instead of running via seconds, run via ticks. (And adjust the slider to accomodate for higher values obviously) Or allow decimals.

a-couple-objects:
coding on da fone

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
@@ -33,10 +33,10 @@ public class Spam extends Module {
 
     private final Setting<Integer> delay = sgGeneral.add(new IntSetting.Builder()
             .name("delay")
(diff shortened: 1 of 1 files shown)


Wide-Cat:
ez coding on da fone

seasnail8169:
bad

seasnail8169:
wont work

supakeks:
why

a-couple-objects:
why not

seasnail8169:
check your code, see the obvious mistake

seasnail8169:
This will mean it will use tick delay for the first message and then seconds from then on

seasnail8169:
just dont pr from the phone please

PR #985, opened 2021-06-07
Added SpawnProofer and moved isValidMobSpawn() to BlockUtils

StormyBytes:
SpawnProofer.
Resolves #719 , resolves #916 

Useful for spawnproofing massive areas for farms.

PR #1071, opened 2021-06-17
Added Spawnproofer

StormyBytes:
A rewrite of #985 so it works with latest changes.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/SpawnProofer.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/SpawnProofer.java
@@ -16,12 +16,8 @@ import meteordevelopment.meteorclient.utils.world.BlockIterator;
 import meteordevelopment.meteorclient.utils.world.BlockUtils;
 import meteordevelopment.orbit.EventHandler;
 import net.minecraft.block.*;
-import net.minecraft.block.enums.BlockHalf;
-import net.minecraft.block.enums.SlabType;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.registry.Registry;
-import net.minecraft.util.shape.VoxelShapes;
-import net.minecraft.world.LightType;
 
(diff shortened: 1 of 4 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/world/SpawnProofer.java
```diff
@@ -0,0 +1,187 @@
...
+    );
+
+    private final Setting<Boolean> rotate = sgGeneral.add(new BoolSetting.Builder()
+        .name("rotate")
+        .description("Rotates towards the blocks being placed.")
+        .defaultValue(true)
+        .build()
+    );
+
+    private final Setting<Boolean> alwaysSpawns = sgGeneral.add(new BoolSetting.Builder()
```
seasnail8169, line 60: would you make this an enum setting
 
Spawns {
     Always,
     Potential
}

src/main/java/meteordevelopment/meteorclient/systems/modules/world/SpawnProofer.java
```diff
@@ -0,0 +1,187 @@
...
+            block instanceof TransparentBlock ||
+            block instanceof TripwireBlock;
+    }
+
+    private boolean isLightSource(Block block) {
+        return block.getDefaultState().getLuminance() > 0;
+    }
+
+    private static class Spawn {
+        public BlockPos.Mutable blockPos = new BlockPos.Mutable();
```
seasnail8169, line 179: this class is usless, just use the BlockPos.Mutable class for your list and pool

src/main/java/meteordevelopment/meteorclient/systems/modules/world/SpawnProofer.java
seasnail8169: Actually just use the enum BlockUtils.MobSpawn

PR #1075, opened 2021-06-18
AutoNametag Rewrite

### LINKED ISSUE #708: [Bug] Auto nametag uses multiple nametags when it doesn't need to.
modules like auto nametag and auto breeder often use nametags multiple times

StormyBytes:
- Clean up and rewrote AutoNametag.
- Fixes #708

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoNametag.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoNametag.java
@@ -49,6 +49,13 @@ public class AutoNametag extends Module {
             .build()
     );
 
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoNametag.java
```diff
@@ -45,61 +56,51 @@
...
+        if (!findNametag.found()) {
+            error("No Nametag in Hotbar");
+            toggle();
+            return;
         }
+
+
+        // Target
+        target = TargetUtils.get(entity -> {
+            if (entity.hasCustomName()) return false;
```
C10udburst, line 82: maybe compare the custom name against nametag's name, so you could rename already named entities

PR #1099, opened 2021-06-22
Don't use ASM for fov modification

JFronny:
I have confirmed the zoom functionality (the only place where this event seems to be used) to still work as expected.


MineGame159:
It uses ASM for Canvas compatibility.

RickyTheRacc:
Cope!

PR #1109, opened 2021-06-24
Add outlines to ActiveModules HUD

StormyBytes:
image


seasnail8169:
Have you fixed the gaps in some of the lines? Some have gaps or incomplete corners, if not could you fix that?

MineGame159:
When you have outline width set to 1 then the lines kinda overflow on the bottom like this
image


And you need to set sliderMin to be 1

PR #1123, opened 2021-06-25
Some changes with ca, noslow, autocity & anchoraura

supakeks:
CrystalAura: Attack Frequency (This is good on NCP servers that have a hard limit on how many PlayerInteractEntityC2SPacket packets you can send in 1 second)

NoSlow: Fixed web timer not resetting the timer in some cases

AutoCity: Auto Switch (I made it a setting since someone may wanna wait until the block can be broken to auto switch with packet mine, this can be good to make it look like you're not citying a player)

AnchorAura: Made it place in liquids (it can be broken inside water/lava)


blendifyClient:
add that seasnail

RedCarlos26:
We need this

MineGame159:
Why are you changing every setting

supakeks:
> Why are you changing every setting

in ca? since everytime you make a new setting it only indents once instead of twice like before so I just changed ca settings to do that since I assume every other module's settings are gonna be changed too

RickyTheRacc:
Please merge this I'm begging

RedCarlos26:
Yes

PR #1176, opened 2021-07-05
Add regex filtration to BetterChat

StormyBytes:
Wide suggestion


seasnail8169:
very wide

RickyTheRacc:
What is regex filtration

seasnail8169:
bebabebabeabe

MineGame159:
1. Look at any other list setting, you need to override the reset method and recreate the list.
2. In my opinion the string list setting should just be inline in the gui like it was before and a button that opens new screen.
3. While you at it can you also implement the list setting to other modules

PR #1185, opened 2021-07-06
Unmerged XRay and WallHack, added XRay opacity, added compat between both modules

Walaryne:
- XRay's opacity takes precedence if both XRay and WallHack are enabled
- WallHack's list of blocks to render transparent takes precedence over the "make everything else transparent" with XRay if enabled


seasnail8169:
what is the point of this? they both pretty much do the same thing and we already decided on it bruh

PR #1199, opened 2021-07-09
attempt to prevent people from being stupid

C10udburst:
- Fixed crash if someone set their prefix to nothing
- Added a button to HUD module that opens HUD tab
- Made buttons in HUD module into icons
- Added a prompt if someone set their prefix to nothing
- Added a prompt if someone set their prefix to something long
- Added a prompt if someone set their prefix to `/`
- Added a prompt if someone set their prefix to something which would conflict other keybinds

Added checkboxes to prompts if someone wants to ignore them

image


RedCarlos26:
This is great

MineGame159:
Can you make it save the list of prompts to not show as list of strings instead of integers? The overhead of strings vs integers doesnt matter at all in this case and is much clearer on what is happening.

PR #1237, opened 2021-07-15
Rewrote AutoSmelter

Thunderkill:
**AutoSmelter had a few issues:**
- Using unwanted items as fuel (for example enchanted bows)
- Smelting unwanted items (for example cobblestone)
- Not taking out items before changing smeltable item

These are the issues i planned to fix, but happened to rewrite almost the whole thing.

**New features:**
You select what items to smelt and what items to use as fuel
Will take out smelted items immediately
Ability whether to disable the module when you run out of items or not


src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSmelter.java
```diff
@@ -6,137 +6,125 @@
...
+    private final Setting<Boolean> disableWhenOutOfItems = sgGeneral.add(new BoolSetting.Builder()
+        .name("disable-when-out-of-items")
+        .description("Disable the module when you run out of items")
+        .defaultValue(true)
+        .build()
+    );
+
 
     public AutoSmelter() {
         super(Categories.World, "auto-smelter", "Automatically smelts all items in your inventory that can be smelted.");
```
timoreo22, line 55: This description should probably be changed

timoreo22 (changes requested):
Looks good ! the module's description should probably be changed tho

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSmelter.java
Thunderkill: I shall change it immediately

Thunderkill:
There seems to be some crashing on startup, i'll need to investigate that for a bit.

Thunderkill:
Also, the game will crash if you try to open the settings while in main menu, any ideas how i could fix that?

It happens because it is checking for all items that are smeltable, but that can't be done in the menu i assume.

MineGame159:
Just don't display anything while in main menu

Thunderkill:
> Just don't display anything while in main menu

Or then display all items?

MineGame159:
I think it would be better to not display anything

Thunderkill:
Closing PR in favor of a new one

PR #1238, opened 2021-07-15
Rewrote AutoSmelter (2)

Thunderkill:
This is an update for #1237

This PR now includes fixes for the settings saving issue, and contains an updated description


MineGame159:
Can you make it not display any items in main menu? Because now they can select for example stone as fuel item while in menu and when they go into a world and open the list the stone will be hidden by the filter but it will still remain selected in the list.

Thunderkill:
> Can you make it not display any items in main menu? Because now they can select for example stone as fuel item while in menu and when they go into a world and open the list the stone will be hidden by the filter but it will still remain selected in the list.

Well, i tried that; but then it wouldn't properly load the items at all

MineGame159:
Well then you did something wrong, I don't see a reason why it shouldn't work

Thunderkill:
> Well then you did something wrong, I don't see a reason why it shouldn't work

Well, i can't figure what i've done wrong. I mean it should work but it doesn't, it just won't load the items.

My assumption was that the item won't be loaded as it won't fit the filter on init

MineGame159:
The filter is ran every time the screen is opened

Thunderkill:
> The filter is ran every time the screen is opened

Actually the filter **is** run on startup:
ItemListSetting.java line 83


After some testing around, i settled on just showing everything. As the module itself checks the item again if it's smeltable or can be used as a fuel.

And after restarting the client it will remove the faulty items that should not be in the item lists.

MineGame159:
Yes what you are saying is correct but this filter is still ran every single time when you open the screen

Thunderkill:
> And after restarting the client it will remove the faulty items that should not be in the item lists.

Actually nevermind about this part

Thunderkill:
> Done. Is there anything you want to do or can I merge.

You can merge

PR #1243, opened 2021-07-16
Added adjustable cooldown to FastUse

### LINKED ISSUE #1200: [Suggestion] Fast Use adjustable cooldown
Fast Use sets the item use cooldown to 0 ticks which has gotten me kicked from servers by the anti-cheat and causes some blocks to disappear when placed.

A useful change to the Fast Use module would be adding a slider to adjust the cooldown ticks. For example I can get away with a 1 tick cooldown on some servers without getting kicked or encountering block glitches but only a 3 tick cooldown on other servers.

I'm not sure if all items have the same cooldown (for block placing it's either 4 or 5 ticks can't remember), but I think a good range would be from 0-5 ticks to allow room for adjustment depending on user preference and how aggressive a server's anti-cheat is.

masranber:
Closes #1200


MineGame159:
Removed the maximum value from the setting, I prefer to not use it unless higher values break something.

PR #1347, opened 2021-08-07
I added option to disable command suggestions in chat :,)

DarkReaper231:


src/main/java/meteordevelopment/meteorclient/mixin/CommandSuggestorMixin.java
```diff
@@ -71,4 +74,9 @@ public void onRefresh(CallbackInfo ci, String string, StringReader reader) {
         }
     }
 
+    @Inject(method = "render", at = @At(value = "HEAD"), cancellable = true)
+    public void render(MatrixStack matrices, int mouseX, int mouseY, CallbackInfo info) {
```
StormyBytes, line 78: Give this method a better name.
Something like `onRenderSuggestion` would be better.

DarkReaper231: Should be done now

src/main/java/meteordevelopment/meteorclient/systems/modules/render/NoRender.java
```diff
@@ -113,6 +113,13 @@
             .build()
     );
 
+    private final Setting<Boolean> onRenderSuggestion = sgOverlay.add(new BoolSetting.Builder()
```
StormyBytes, line 116: `onRenderSuggestion` makes no sense for a name.
`noCommandSuggestions` would be better.

src/main/java/meteordevelopment/meteorclient/systems/modules/render/NoRender.java
```diff
@@ -336,6 +343,10 @@ public boolean noEatParticles() {
         return isActive() && noEatParticles.get();
     }
 
+    public boolean onRenderSuggestion() {
```
StormyBytes, line 346: `noCommandSuggestions`

src/main/java/meteordevelopment/meteorclient/mixin/CommandSuggestorMixin.java
DarkReaper231: Sorry for so many commits. I do it in browser since I didn't setup git yet.

StormyBytes (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/render/NoRender.java
DarkReaper231: Okay now it should be done

PR #1381, opened 2021-08-17
Allow addons to share NBT data

StormyBytes:


src/main/java/meteordevelopment/meteorclient/MeteorAddon.java
```diff
@@ -5,8 +5,12 @@
...
 package meteordevelopment.meteorclient;
 
+import net.minecraft.nbt.NbtCompound;
+
 public abstract class MeteorAddon {
     public abstract void onInitialize();
 
     public void onRegisterCategories() {}
+
+    public abstract NbtCompound getAddonNBT();
```
C10udburst, line 15: Maybe dont make this abstract but return null instead

MineGame159:
I don't get whar is the point of this

seasnail8169:
neither

StormyBytes:
One of the uses I can think of,

Say for example 2 or more addons want to show credit on the title screen. It'll be a problem if they render the text on the same position. Rather than hardcoding a check for other addons, simply checking for addons with a nbt data like `createsTitleScreenCredit` would be better.

MineGame159:
Thats not really a good solution either. There is no standardised way to indicate its adding credit. Where is it rendering the credit? Where should my addon render it?

StormyBytes:
I can iterate through the list and count how many addons have the credit flag before my addon. I can use that to calculate the amount of y-offset I need.

This can be applied with other addons since there won't be repeating keys in the set, each addon will have a different y-offset and avoid conflicts

MineGame159:
And how do you know every addon will render it in the same corner?

StormyBytes:
For that I could use a more specific flag with corners specified too.

MineGame159:
Or have a more standardised way of adding credits

MineGame159:
Which would also mean addon developers wont need to care about if other addons add credits

StormyBytes:
That could work, but how about other purposes? Maybe 2 addons need to inform each other something else other than credits?

Hardcoding would work fine for just a few, but would escalate badly when multiple addons are taken into measure.

MineGame159:
Well I was specifically saying the credits approach is bad, not the concept as a whole. But I still can't see any real scenario where making a proper API won't be multiple times better than this approach.

PR #1383, opened 2021-08-17
Add scroll to sliders

### LINKED ISSUE #1382: [Suggestion] Scroll to de/increase values in options
**Describe the feature**
Every option that have a value also have a bar to increase/decrease the value of it, a great improvement would be able to use the mouse Scroll Wheel to increase and decrease those values while hovering with mouse over bar.

**Screenshots**
image
Above image shows a situation where we would be able to use our mouse scroll to change the value.

**Additional info**
I always try to use the scroll my mouse have the click failing and I'm used to Item Scroller mod.

maxsupermanhd:
Fixes #1382

Changes double values by 0.05 and integer by 1. Triggered only when mouse is over slider handle in order to stop random values from changing on casual menu scroll.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WDoubleEdit.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WDoubleEdit.java
@@ -108,7 +108,11 @@ public class WDoubleEdit extends WHorizontalList {
     }
 
     public void set(double value) {
-        this.value = Utils.clamp(value, min, max);
+		if(min != null && max != null) {
+			this.value = Utils.clamp(value, min, max);
+		} else {
+			this.value = value;
+		}
 
         textBox.set(valueString());
         if (slider != null) slider.set(value);
--- a/src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WDoubleEdit.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WDoubleEdit.java
@@ -108,15 +108,7 @@ public class WDoubleEdit extends WHorizontalList {
     }
 
     public void set(double value) {
-		double clmax = 0;
-		double clmin = 0;
-		if(max == null) {
-			clmax = Double::MAX_VALUE;
-		}
-		if(min == null) {
-			clmin = Double::MIN_VALUE;
-		}
-		this.value = Utils.clamp(value, min, max);
+		this.value = value;
 
         textBox.set(valueString());
         if (slider != null) slider.set(value);
--- a/src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WDoubleEdit.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WDoubleEdit.java
@@ -108,11 +108,15 @@ public class WDoubleEdit extends WHorizontalList {
     }
 
     public void set(double value) {
-		if(min != null && max != null) {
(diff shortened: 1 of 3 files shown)


MineGame159:
Can you also add an `else` block that will just increment the slider if its not part of an int edit or double edit. And the clamping before calling `set()` methods can be put inside the `set()` method itself, I dont know why its not there already.

maxsupermanhd:
> that will just increment the slider

As I tested it will desync slider from parent widget for some reason, I am new to codebase and widgets in general, slowly figuring it out. Draft for now but I want this feature so bad...

Yes, adding clamp to `WDoubleEdit` and `WIntEdit` already.

MineGame159:
Hmmm. You would need to call `WSlider.action` runnable callback after changing the individual slider's value. Bear in mind this cannot be put inside the set() method itself because its called from other callbacks.

src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WSlider.java
```diff
@@ -98,6 +98,14 @@ public boolean onMouseReleased(double mouseX, double mouseY, int button) {
 
         return false;
     }
+	
+	@Override
+	public void onMouseScrolled(double ammount) {
```
C10udburst, line 103: minor spelling mistake

MineGame159:
You just made it so it adds 0.05 even if its int edit. So it does nothing because it gets rounded.

maxsupermanhd:
Sorry for my stupidity, I am new to java in general but meteor just attracts pull requesting so good so I can not resist. (also my native language is Russian so sorry for that)

MineGame159:
One QOL feature you could add is that currently it only allows you to scroll when hovering above the handle, which means when scrolling you have to keep moving the mouse in that direction. It could remember the mouse position you started scrolling at and if you dont move your mouse then even if its not hovering above the handle it still allows you to scroll until you move your mouse. Hope that makes sense.

maxsupermanhd:
Yeah, any update of mouse will stop scrolling events, this is indeed annoying - thinking about how will be best to implement it.

maxsupermanhd:
After this commit it should be better

maxsupermanhd:
Well, I can not really understand how to pass sensitivity setting to slider here at all... To be clear, it is *designed* for really really fine control, if you want to move further fast you can just click or drag, kinda defeats purpose.

I don't understand what do you mean by that, can you tell me where or because of what action it crashes because it is running absolutely fine for me. Maybe traceback?

src/main/java/meteordevelopment/meteorclient/gui/widgets/input/WDoubleEdit.java
```diff
@@ -107,7 +108,7 @@ public double get() {
     }
 
     public void set(double value) {
-        this.value = value;
+        this.value = Utils.clamp(value, min, max);
```
MineGame159, line 111: Here `min` and `max` are boxed Double types and not primitives, that means they can be null and null means infinite. Same gues for int edit.

maxsupermanhd:
latest commit *should* fix crashing issue but I never successfully reproduced that crash...

MineGame159:
You didn't notice the crash because the methods are not called from the sliders, I crashed when clicking the reset button. Also what you did is wrong. You added a check if both are not null but it is very common that minimum value is set (not null) and maximum is not set (null).

MineGame159:
You know what just remove the clamping, its not needed and you just commited it with a compile time error.

PR #1389, opened 2021-08-19
Make packet mine block class public

maxsupermanhd:
Make `MyBlock` accessible so other modules and addons can use it too.

I am currently working on villager trade rolling addon and I don't really want to copypaste packet mine basically.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/PacketMine.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/PacketMine.java
@@ -251,7 +251,7 @@ public class PacketMine extends Module {
         return speed;
     }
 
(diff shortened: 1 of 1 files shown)


MineGame159:
You realise it depends on settings from Packet Mine.

MineGame159:
And now more specifically to your use case, use `BlockUtils.breakBlock()`, call it from `TickEvent.Pre` every tick until the block is gone. And if the user wants to packet mine they can just enable the module.

maxsupermanhd:
But @MineGame159, what is the best method to check that block is broken? Let's assume I spammed `BlockUtils.breakBlock()` but how can I stop and realize that it is broken?

MineGame159:
If its air

PR #1428, opened 2021-08-30
AutoEz

Ezzer29:
Bad AutoEz made by none other than Ezzer himself with a tf2 touch :)) Mongrel!
Has random text from list and dominations :) tf2 style

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
@@ -362,6 +362,7 @@ public class Modules extends System<Modules> {
         add(new AntiAnvil());
         add(new AntiBed());
         add(new ArrowDodge());
+        add(new AutoEz());
         add(new AutoAnvil());
         add(new AutoArmor());
         add(new AutoCity());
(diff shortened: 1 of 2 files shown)


Ezzer29:
oh, looks like my code is shit :|

Wide-Cat:
Based

RickyTheRacc:
EZZZZZZZ

MineGame159:
1. Move fields below settings and above constructor
2. Settings need to be named like this `ignore-friends`, not with spaces and uppercase characters
3. All of the fields can be `private finale`, one if public and some dont have anything
4. Best practive for declaring lists is using the List class, eg `List<UUID> list = new ArrayList<>()`
5. In `msgplayers` and `kills` you are using usernames which would be better to use UUIDs
6. In onTick you are assigning `newPlayers = players` and then at the end assigning it the other way around, this is useless and does nothing
7. `players` could be a Map (map from uuid to time) and `kills` could be a Map too (map from uuid to number of kills)
8. Please fix your formatting, especially in the `send` the send method, you have spaces around ( ) when in other places there are no spaces

Also I dont understand the whole sleeping the thread thing, cant you just immediatelly send the message when the player dies?

PR #1476, opened 2021-09-06
Added prerequisites for issue submitting

StormyBytes:
Before making sure your bug report
- [ ] This bug wasn't already reported.
- [ ] This is a valid bug and I have checked all settings.

---

Before making your suggestion
- [ ] This feature doesn't already exist in the client.
- [ ] This wasn't already suggested.
- [ ] This feature is a valid suggestion, according to the FAQ and Guides.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -3,18 +3,14 @@ description: Report a bug to help us improve Meteor.
 title: '[Bug] '
 labels: bug
 body:
-  - type: markdown
-    value: |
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -3,6 +3,13 @@ description: Report a bug to help us improve Meteor.
 title: '[Bug] '
 labels: bug
 body:
+  - type: markdown
```
C10udburst, line 6: maybe use `type: checkboxes` with `required: true`

Issue #1496, opened 2021-09-12
gui scale


confused101-1:
### Describe the bug

when you go into gui scale and change it to a bigger number then you wanted it turs the gui to big to revert back to its normal size and when you do .reset gui it doesn't do anything

### Steps to reproduce

```bash
you go into the gui scale and change it to a big number which make it your gui to big
```

### Meteor Version

1.17.1

### Minecraft Version

1.17.1

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported.
(I have checked every bug report on github)

- [X] This is a valid bug.
(I am able to reproduce this on the latest dev build)

ChrstnCrrnd:
.reset gui resets the positions not the scale, can easily be fixed

Wide-Cat:
aaca110960da1baaa3a6a6ec52db71c38e49d237

PR #1497, opened 2021-09-12
Fixed .reset gui not resetting scale

ChrstnCrrnd:
fixed issue #1496


seasnail8169:
this is just a bad way to do it

PR #1499, opened 2021-09-13
Fixed autotool break durability not using percentage

StormyBytes:


StormyBytes:
build failed? odd

seasnail8169:
because jitpack is always going down

PR #1508, opened 2021-09-14
Added Under Height and Render to Surround

Matejko06:
I made Surround render using Scaffold render because normal render was staying visible and blinked too fast. Under Height places blocks under the surround blocks first and then the normal ones to bypass a few anticheats. (Mainly happening on 1.12 servers)


seasnail8169:
This is just all around horrible to look at

PR #1530, opened 2021-09-23
Better NoInteract

StormyBytes:
- Wider range of interaction blocking.
- Wider range of hit blocking.
- Alot more settings for better customizability.


StormyBytes:
smh, Jitpack always down

PR #1544, opened 2021-09-29
Fix modulelist rainbow speed depending on FPS

Toshimichi0915:


MineGame159:
If there is a problem with the delta variable then that should be fixed instead of just fixing this one instance by keeping track of the time manually.

Toshimichi0915:
Sorry but what do you mean?
Delta variable is from vanilla and there's nothing I can do about it.
Also, tickDelta(partialticks) is not intended for measuring time.
We can use either real time(currentTimeMillis) or client ticks but in this case either will do

MineGame159:
I thought the delta time that meteor supplies is already in the correct format, guess I never did just thought about it

PR #1587, opened 2021-10-16
Baritone smart sprint

nothub:
This adds a setting to automatically enable and disable Baritones sprint based on food saturation.


nothub:
> We are not developers of Baritone.

?

MineGame159:
This should be suggested to Baritone, not us.

MineGame159:
Jesus christ im blind. Sorry about that. Either way I don't see how this is useful, surely it is better to make it eat, aka Auto Eat.

nothub:
I use it for the situation where the players food items are used up to automatically stop sprinting.

nothub:
just for reference, addon lives here now https://github.com/nothub/meteor-baritone-tweaks

PR #1589, opened 2021-10-16
KillAura ignore passive enderman and piglin

nothub:
This adds a setting to ignore non attacking enderman and piglin. tested with paper-1.17.1-333.jar

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java
@@ -26,10 +26,9 @@ import meteordevelopment.orbit.EventHandler;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityType;
 import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.Tameable;
 import net.minecraft.entity.mob.EndermanEntity;
 import net.minecraft.entity.mob.MobEntity;
-import net.minecraft.entity.mob.PiglinEntity;
-import net.minecraft.entity.mob.ZombifiedPiglinEntity;
 import net.minecraft.entity.passive.AnimalEntity;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.AxeItem;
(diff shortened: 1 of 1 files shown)


RickyTheRacc:
BASED

C10udburst:
do you know that there is already a setting to control which entities to attack? [right here](https://github.com/MeteorDevelopment/meteor-client/blob/996573df8cf2be84ca75b756e1731cb6b53b82a6/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java#L110)

nothub:
> do you know that there is already a setting to control which entities to attack? [right here](https://github.com/MeteorDevelopment/meteor-client/blob/996573df8cf2be84ca75b756e1731cb6b53b82a6/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java#L110)

Yeah but i could not find a way to separate passive and attacking enderman and piglin.

C10udburst:
ah yeah good point

nothub:
> I always wanted to do this but surely there is a better way to do it, cause there is other mobs that are passive unless you attack them so there is probably some flag somewhere to determine it. Maybe you could look for that to improve this?

Yeah there is `MobEntity.isAttacking()` but it could introduce some edge cases like targeting the players own dog that attacks enemies. I gonna change it and test if i can find any big problems.

arlomcwalter:
ok i found it 
image

there is a class called "Angerable" which those entities inherit

arlomcwalter:
yeah looks like its more complex for piglins

nothub:
Also `UUID target = ((Angerable) entity).getAngryAt();` returns null while the enderman is attacking the player.
So i guess Angerable is just useable in a server context.

nothub:
> > ok i found it image
> > there is a class called "Angerable" which those entities inherit
> 
> Oh yeah this looks relevant, but it looks like its missing the normal piglin. Gonna test.

I cant really find another way to check for the entities target in a sane manner.
Should i leave the logic like it is right now (enderman + piglin check only) or extend it to also check `(MobEntity) entity).isAttacking()` for the other entity types that are extending Angerable, to target them no matter if they are actually hostile towards the player?

necrophili:
Add pigmen and golems too?

MineGame159:
@nothub I think the best work around would to have a specific check for `EndermanEntity` using the `isAngry()` method and target all mobs using `MobEntity.isAttacking()` **_but_** if the entity also implements `Tameable` interface check if the owner is not the current player.

nothub:
> @nothub I think the best work around would be to have a specific check for `EndermanEntity` using the `isAngry()` method and target all mobs using `MobEntity.isAttacking()` **_but_** if the entity also implements `Tameable` interface check if the owner is not the current player.

Okay, i pushed a commit with the suggested changes.

Bluscream:
i merge this

nothub:
@MineGame159 anything else you want me to change?

PR #1597, opened 2021-10-20
Improve module search

JFronny:
- Ensure search contains modules only once (prevents the module search results from containing the same module multiple times)
- Use Set<Module> as return type for module searches as the integers are unused
- Count word occurences towards a modules ranking
- Require all words to be present in strings when searching


MineGame159:
Looks good but why create a new class `ValueComparableMap`? Can't you just create a new instance of `TreeMap`?

JFronny:
The order of a normal TreeMap is determined by a function comparing keys, not values. The class works around issues that would arise if I just used a comparator that looked up the corrosponding entry. It also makes the code cleaner.

PR #1600, opened 2021-10-21
Add keywords to modules for discoverability

JFronny:
When new users use meteor for the first time, they are likely to miss modules they would otherwise use because they didn't find them through the search feature.\
The goal of this PR is to add keywords used to describe modules elsewhere and respect them when searching.\
Example: new users commonly search for a "no push" module and fail to discover velocity, causing them to post in the meteor discord.

Separate from my previous search PR because this is probably a more controversial change


arlomcwalter:
this is shit

PR #1633, opened 2021-10-31
Add a waypoint command

Wide-Cat:
Added a command to help manage waypoints. Usage:

`.wp get [name]` gets information on all waypoints, or more detailed info on a specific one
`.wp add [name]` adds a waypoint to your current position with the specified name
`.wp delete [name]` deletes the specified waypoint
`.wp toggle [name]` toggles the specified waypoints visibility


MineGame159:
Move the WaypointArgumentType into a separate file next to the other argument types. And I think it would be better if `.wp get` was `.wp list`.

Wide-Cat:
should i move ProfileArgumentType into a seperate file while im at it

MineGame159:
Yes.

PR #1686, opened 2021-11-18
Add Anti Spam Bypass

Toshimichi0915:


RickyTheRacc:
Technically already a thing with the random suffix in better chat, but this looks easier to find and much more effective so I would actually vouch to add this and remove random suffix

RedCarlos26:
Same

PR #1727, opened 2021-11-28
add a module to delay chorus teleporting

Wide-Cat:


Wide-Cat:
I'll rewrite this to 1.18 and open a new pr when done

PR #1735, opened 2021-12-02
Toggle module on enter key in search text box

Toshimichi0915:
The only concern is that
WTextBox.java:178
```java
        else if (key == GLFW_KEY_ENTER || key == GLFW_KEY_KP_ENTER) {
            setFocused(false);
            return true;
        }
```

I removed the following line because setFocused method also invoke actionOnUnfocused. I think it was a bug so I removed the line but I'm not 100% sure so...
```java
if (actionOnUnfocused != null) actionOnUnfocused.run();
```


RickyTheRacc:
What a thrill!

Toshimichi0915:
don't merge yet i have to fix

Toshimichi0915:
That should be fine

MineGame159:
I don't think this is a good idea.

PR #1739, opened 2021-12-03
Added option for only showing exposed ores with xray

VolcanoCookies:
Added option to xray module to only show exposed blocks. Those that have at least one neighbor that is non-opaque.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/BlockModelRendererMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/BlockModelRendererMixin.java
@@ -39,7 +39,7 @@ public class BlockModelRendererMixin {
             }
 
             rewriteBuffer(vertexConsumer, alpha);
-        } else if (xray.isActive() && !wallHack.isActive() && xray.isBlocked(state.getBlock())) {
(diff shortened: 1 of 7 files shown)


MineGame159:
Isn't Xray's whole point that it displays hidden blocks?

VolcanoCookies:
I find this useful in situations where you for example cannot break blocks for any reason but can interact with them. 
With the new massive caves it can also be quicker to run to exposed ores and mine them rather than digging into deepslate to get hidden ones.

_Also, broken build due to wrong java version?_

arlomcwalter:
also would show ores with any non opaque block next to it, its not quite the same as being exposed

VolcanoCookies:
True, but in natural generation that almost always means its in a cave/surface, assumed doing more than a simple check like this would be rather detrimental to performance.
Should maybe rename it through?

MineGame159:
Could you change the `isExposed(BlockPos)` method to be more efficient and not create new `BlockPos` for every neighbour, basically have a global `BlockPos.Mutable` field that you just change the position of instead of calling `.offset()`.

VolcanoCookies:
This should do it?

PR #1748, opened 2021-12-04
Minor Improvements to KA

### LINKED ISSUE #1741: [Suggestion] Nuker; Interact
### Describe the feature

1. Add a setting to nuker to toggle between "break" and "interact".
Breaking would act is it normally does, however interact would right click on all whitelisted blocks within range. This has many misc uses such as spam opening chests and/or other blocks to lag servers or some other purpose.

Also, this has already been suggested but I'll mention it again. add goddamn custom shapes to nuker. x blocks left, y blocks right, z blocks up, etc.
and 'smash' mode explains nothing. still don't know what it's supposed to be doing

StormyBytes:
- Fixes #1741
- Improved random delay


StormyBytes:
oops wrong issue number

Issue #1753, opened 2021-12-04
[Bug]  impossible to put waypoints under y = 0


mijax:
### Describe the bug

waypoints under y= 0 are impossible, the value changes back to 0 when you put it on a negative value.

### Steps to reproduce

```bash
waypoints under y= 0 are impossible, the value changes back to 0 when you put it on a negative value.
```

### Meteor Version

0.4.7 build 1349

### Minecraft Version

1.18

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported.
(I have checked every bug report on github)

- [X] This is a valid bug.
(I am able to reproduce this on the latest dev build)

MineGame159:
Fixed in db3b567ca17eaf647d0e5e655677d6b01e130123.

PR #1756, opened 2021-12-05
[Bug Fix] Waypoint Y level

VolcanoCookies:
Fixed waypoints not allowing for use of new increased height in the overworld.

Original issue #1753


VolcanoCookies:
_No idea how to link this to issues so I just added it in the body._

PR #1772, opened 2021-12-08
Fix the light overlay on 1.18

### LINKED ISSUE #1762: [Bug] Light Overlay no longer accurate with 1.18 mob spawn changes
### Describe the bug

Mobs now only spawn at light level 0, should be an easy fix I think.

### Steps to reproduce

```bash
Mobs now only spawn at light level 0, should be an easy fix I think.
```

### Meteor Version

v0.4.7-1355

### Minecraft Version

1.18

### Operating System

Windows

- [X] This is a valid bug.

carlf:
This fixes the light overlay on 1.18. Mobs now only spawn at light level 0 rather than 7 as on 1.17.x. I'm unsure if there should be a version check to handle earlier versions. Happy for feedback on that or any other point.

This should fix #1762

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/LightOverlay.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/LightOverlay.java
@@ -60,6 +60,13 @@ public class LightOverlay extends Module {
             .build()
     );
 
+    private final Setting<Boolean> newMobSpawnLightLevel = sgGeneral.add(new BoolSetting.Builder()
+            .name("now-mob-spawn-light-level")
+            .description("Use the new (1.18+) mob spawn behavior")
+            .defaultValue(true)
+            .build()
+    );
+
     // Colors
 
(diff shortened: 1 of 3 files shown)


MineGame159:
There could be a setting to turn back on the old system. As for a version check you sadly can't reliably get the actual server version.

carlf:
Ok. I think I see how to do the settings. I'll update this shortly.

carlf:
Ok. So I added an integer setting for Light Overlay and Spawn Proofer. I could alternatively just make this a boolean to choose between 7 and 0. This seems more flexible but possibly more confusing for end users. Any preference?

MineGame159:
I don't think I have seen any server where the default mob spawning light level was changed so in my opinion a boolean should be enough. If you have heard of such a thing then sure, leave it in.

carlf:
Fair. I'll change it to a boolean.

MineGame159:
There is a typo in the setting names, now -> new.

carlf:
Good catch. Fixed.

PR #1773, opened 2021-12-09
add an option in inv tweaks to change the entity model in your inv

Wide-Cat:
@MineGame159 accept my pr or you're a coper for all eternity


MineGame159:
I really don't see why this should be in Meteor. If you want you can make a PR with just the entity setting and add it to some addon.

Wide-Cat:
😭

RickyTheRacc:
Enormous cope from mimgam

necrophili:
Can #24946bd be pulled at least 😭

PR #1787, opened 2021-12-12
Place Helper module

GenixZero:
This option checks to see if a block is next to the hit result. If enabled, it will loop through the max range to 1 to find the best hit result.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AirPlace.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AirPlace.java
@@ -14,13 +14,10 @@ import meteordevelopment.meteorclient.systems.modules.Module;
 import meteordevelopment.meteorclient.utils.render.color.SettingColor;
 import meteordevelopment.meteorclient.utils.world.BlockUtils;
 import meteordevelopment.orbit.EventHandler;
-import net.minecraft.block.Material;
 import net.minecraft.item.BlockItem;
 import net.minecraft.util.Hand;
 import net.minecraft.util.hit.BlockHitResult;
 import net.minecraft.util.hit.HitResult;
-import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.math.Direction;
 
 public class AirPlace extends Module {
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
@@ -31,13 +28,6 @@ public class AirPlace extends Module {
     private final Setting<Boolean> render = sgGeneral.add(new BoolSetting.Builder()
         .name("render")
         .description("Renders a block overlay where the obsidian will be placed.")
-        .defaultValue(true)
-        .build()
-    );
-
-    private final Setting<Boolean> onlyColliding = sgGeneral.add(new BoolSetting.Builder()
-        .name("collide-check")
-        .description("Only places if there is a block next to it.")
         .defaultValue(true)
         .build()
     );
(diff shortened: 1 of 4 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/world/AirPlace.java
```diff
@@ -99,4 +105,29 @@ private void onRender(Render3DEvent event) {
 
         event.renderer.box(((BlockHitResult) hitResult).getBlockPos(), sideColor.get(), lineColor.get(), shapeMode.get(), 0);
     }
+
+    private void setHitResult() {
+        final double r = customRange.get() ? range.get() : mc.interactionManager.getReachDistance();
+        for (int i = (int) r; i > 0; i -= 1D) {
+            hitResult = mc.getCameraEntity().raycast(Math.min(r, i), 0, false);
```
StormyBytes, line 112: Why do you need this part though? Doesn't this just ray cast but with decrementing lengths?

Also why do you need to use `Math.min(r, i)` when `i` will always be equal or less than `r`, just use `i` instead.

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AirPlace.java
```diff
@@ -99,4 +105,29 @@ private void onRender(Render3DEvent event) {
...
+        if (!onlyColliding.get()) {
+            return true;
+        }
+        for (Direction dir : Direction.values()) {
+            assert mc.world != null;
+            if (mc.world.getBlockState(pos.add(dir.getOffsetX(), dir.getOffsetY(), dir.getOffsetZ())).getMaterial() != Material.AIR) {
+                return true;
+            }
+        }
+        return false;
```
StormyBytes, line 131: Can't this be simplified to use `BlockUtils.getPlaceSide()` ?

Like,
```
return BlockUtils.getPlaceSide(pos) != null;
```

StormyBytes (commented):
Also isn't the whole point of the module named "AirPlace" to *air place* blocks without needing to care whether they have a side to place on or not?

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AirPlace.java
seasnail8169: yes

GenixZero:
Yea, forgot to remove the Math.min when I was testing something else. Also I don't know if BlockUtils.getPlaceSide works, the current method checks to see if a block is next to it if not, then hit result gets set to null. 
This works well for anticheats that check if a block that is placed is next to anything else.
The reason I loop through the range is to get the best placement. If you look straight down to the void while a block is under your feet, then the for loop will return the hit result of the block next to the one you're standing on.

StormyBytes:
Ah my bad, only now do I realize why the raycast needs to be in decrementing lengths. 

Though my point about the air placing still stands. In my opinion this would be better off as a sort of "place helper" module instead of being in the air place module as it just goes against the module's name.

GenixZero:
I agree with you, I made it into its own module.

seasnail8169:
i don't like this, it being in airplace would be better imo, you should just have a couple modes.

mode 1: *only* place if there is a block to place against
mode 2: try to find a side, if not then just place in the air

i will probably just write this myself tho tbh

StormyBytes:
> i will probably just write this myself tho tbh

10 months ago :trollface:

I-Got-You-Dead:
💀

PR #1789, opened 2021-12-12
Disable selection if inside block

MoMMde:
# Update
I disabled the Block Selection when you are inside a Block

# Config
It is configurable but i don't think it will be needed in future

# Example

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BlockSelection.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BlockSelection.java
@@ -5,6 +5,7 @@
 
 package meteordevelopment.meteorclient.systems.modules.render;
 
+import com.ibm.icu.impl.number.RoundingUtils;
 import meteordevelopment.meteorclient.events.render.Render3DEvent;
 import meteordevelopment.meteorclient.renderer.ShapeMode;
 import meteordevelopment.meteorclient.settings.*;
@@ -19,6 +20,8 @@ import net.minecraft.util.math.Box;
 import net.minecraft.util.math.Direction;
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.util.shape.VoxelShape;
+
+import java.math.RoundingMode;
 
 public class BlockSelection extends Module {
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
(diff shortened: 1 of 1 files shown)


MoMMde:
Fluid blocks not working

MoMMde:
Flying Underground currently has some issues

MoMMde:
Please don't ask why this takes soooo long

src/main/java/meteordevelopment/meteorclient/systems/modules/render/BlockSelection.java
```diff
@@ -65,6 +76,15 @@ public BlockSelection() {
     private void onRender(Render3DEvent event) {
         if (mc.crosshairTarget == null || !(mc.crosshairTarget instanceof BlockHitResult result)) return;
 
+        BlockPos playerEyePos = new BlockPos(mc.player.getEyePos());
+        int crossX = (int) Math.floor(mc.crosshairTarget.getPos().x);
+        int crossY = (int) Math.floor(mc.crosshairTarget.getPos().y);
+        int crossZ = (int) Math.floor(mc.crosshairTarget.getPos().z);
```
MineGame159, line 82: Why are you calculating all of this when it's not needed?

src/main/java/meteordevelopment/meteorclient/systems/modules/render/BlockSelection.java
```diff
@@ -65,6 +76,15 @@ public BlockSelection() {
...
+
+        if(hideInside.get() && result.isInsideBlock()) {
```
MineGame159, line 84: Can you follow our standards and put a space between `if` and `(`. Also the return could be on the same line but that's not important.

src/main/java/meteordevelopment/meteorclient/systems/modules/render/BlockSelection.java
```diff
@@ -17,8 +18,11 @@
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Box;
 import net.minecraft.util.math.Direction;
+import net.minecraft.util.math.Vec3d;
 import net.minecraft.util.shape.VoxelShape;
 
+import java.math.RoundingMode;
```
MineGame159, line 24: Why are you adding imports that are never used? There is another one at the top.

MineGame159 (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/render/BlockSelection.java
MoMMde: Forgot to remove the stuff

MoMMde: I'm new to Mod Development and i needed to figure out what each field does and i just started messing around with different fields i got and in the end i found my current solution but now forgot to remove the old things

Issue #1761, opened 2021-12-05
[Suggestion]  Hide tracer when pressing F1


naBBolethano:
### Describe the feature

As described in the title, nothing special, It's an inconsistency in the modules because everything gets hidden when pressing f1 except tracers.

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

RickyTheRacc:
Mans said what we're all thinking

DesiCow:
Someone forgot to close this issue 👀

Issue #997, opened 2021-06-09
[Suggestion] Hide Tracers in F1


MI2006007:
Add an option for tracers to hide them when pressing F1 key (or other key to hide HUD).

StormyBytes:
Implemented in #1790 2239edba

Issue #721, opened 2021-05-10
[Suggestion] Option to prevent render modules from rendering when in F1 mode


RetroDev256:
If possible, it would be nice to have an option in the "tracers" config (under appearence) that would disable rendering
tracers when F1 was pressed, allowing for cleaner and nicer screenshots.

cpj6789:
I agree with this

RoastedCheese:
Yeah

PR #1790, opened 2021-12-13
Tracers hide in F1

MoMMde:
Option to remove Tracers from beeing drawn.

## Related
https://github.com/MeteorDevelopment/meteor-client/issues/1761
https://github.com/MeteorDevelopment/meteor-client/issues/997
https://github.com/MeteorDevelopment/meteor-client/issues/721


RickyTheRacc:
Based

seasnail8169:
based = good

@MoMMde i dont think this should be a setting, i would just return if hideHud is true

MoMMde:
thanks for the review
i removed to option. should do its job now

PR #1838, opened 2021-12-25
mlg

necrophili:
ye mlg


MineGame159:
I cannot get it to work.

necrophili:
same

LilyRose2798:
Looks like you need a `&&` instead of `||` on line 139.

PR #1853, opened 2021-12-30
Added EntityTypeSetting 

Wide-Cat:
finally remembered to do it 🤦‍♂️


Wide-Cat:
wait i forgot that happened🤦‍♂️

PR #1863, opened 2021-12-31
TOTEM_OF_UNDYING added to items in Offhand.

Kiriyaga7615:
Because crystal-on-ca or crystal-on-mine now can be used for strict servers and prevent some totem fails


HoseanRC:
we have autotot

Kiriyaga7615:
try to read

PR #1868, opened 2022-01-01
Add Russian language support

N08I40K:


MineGame159:
Adding a support for one specific language (outside of english of course) is not worth it. Although if you have some source that documents character ranges for the most commonly used languages I might look at it.

spacewaretwo:
> Adding a support for one specific language (outside of english of course) is not worth it. Although if you have some source that documents character ranges for the most commonly used languages I might look at it.

Here is a list of ranges made.
If such a list does not fit, I can make another one.
[ranges.txt](https://github.com/MeteorDevelopment/meteor-client/files/9179229/ranges.txt)

PR #1872, opened 2022-01-01
added new feature to message aura and spam

HoseanRC:
hey
I added public chat to message aura (as somebody requested in discord)
and added a new bypass for spam (basically adds hidden unicode character to the message randomly so it doesn't effect the message and bypasses the anti spam plugins).

Note: I didn't check the client for any error, but it should work just fine.


RedCarlos26:
Yes please

PR #1912, opened 2022-01-09
Add More Blink Options

Theboiboi8:
Adds options for Trail and Fake Player to the "Blink" module.
Example


arlomcwalter:
all of the settings can go into sgGeneral

MineGame159:
I don't see the point why Blink should completely replicate Breadcrumbs. You can bind both to the same key to toggle them together.

Theboiboi8:
> 

ye but fakeplayer
also convenience

Theboiboi8:
and I mean you might be using breadcrumbs for smthn else and it might interfere if you just bind the same key for breadcrumbs

Theboiboi8:
also this way it's more accessible

Vitalasy:
based pr

PR #1927, opened 2022-01-12
MSTP HUD option

tokfrans03:
It would seem as if the hud folder has moved since I made this, so here's another PR


tokfrans03:
Hello, why was this PR closed? I figured the last one was because of the project restructuring, but this fixes that. 

I added this because I feel that the MSPT metric is more valuable than TPS because of the fact that it more accurately represents the servers performance. And while TPS is better "at a glance", MSTP is more insightful.

MineGame159:
Because it measures client's MSPT, not the server's MSPT.

tokfrans03:
So the normal getTickRate function just gets the clients TPS? MSTP is just the inverse of TPS*20 anyways

MineGame159:
Oh sorry, I kinda missed the part that it just uses the tick rate. So if it is just TPS * 1000 how is it useful, I know MSPT is useful when reported directly from the server because it measures the actual time and not just estimating it from the TPS.

tokfrans03:
While it is true that the accuracy of reported tick times by the client can be inaccurate, low tick rates displayed by a TPS number can only be between 0.0 and 20.0 while the MSTP can be between 50.0 and infinity. 

Because of this, it is also easier to estimate the impact of a machine when the TPS is under 20 since you are finding the difference between two MSTP readings instead of dividing two TPS readings. For example, If a machine adds 10ms to the tick time, two would add ~20ms.

> So if it is just TPS * 1000 how is it useful

Remember, it's the inverse of TPS: `TPS^-1 * 1000`

Since getting MSPT times below 50 is not possible by my current implementation, I will be adding that next if you think it is a good statistic.

arlomcwalter:
it's too niche, 99.9999% of players will never use it. just make a very basic addon that adds it

PR #1948, opened 2022-01-15
Fix and rework init dependency system

Paul1365972:
Not sure what kind of guardian angel this mod had beforehand, because it should have crashed consistently.
Well now its fixed (I think, worked when I tested it)
Also maybe add a cyclic dependency check or something in the future

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/MeteorClient.java
+++ b/src/main/java/meteordevelopment/meteorclient/MeteorClient.java
@@ -41,7 +41,8 @@ import java.io.File;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.Set;
+import java.util.*;
+import java.util.stream.Collectors;
 
 public class MeteorClient implements ClientModInitializer {
     public static final String MOD_ID = "meteor-client";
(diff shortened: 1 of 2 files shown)


arlomcwalter:
i don't understand what was broken about it

src/main/java/meteordevelopment/meteorclient/MeteorClient.java
```diff
@@ -144,38 +145,30 @@ private void openGui() {
...
-    private static void reflectInit(Method task, Init annotation) {
-        Class<?>[] preTasks = annotation.dependencies();
-
-        try {
-            if (preTasks == null || preTasks.length < 1) {
-                task.invoke(null);
-            }
-            else {
-                for (Class<?> aClass : preTasks) {
-                    Set<Method> preInits = new Reflections(aClass).getMethodsAnnotatedWith(Init.class);
```
Paul1365972, line 164: This line does not work, its just not how you use Reflections, it will always return null
Also if it did it would cause initialize to be called a second time

src/main/java/meteordevelopment/meteorclient/renderer/Renderer2D.java
```diff
@@ -28,7 +28,7 @@ public Renderer2D(boolean texture) {
         lines = new ShaderMesh(Shaders.POS_COLOR, DrawMode.Lines, Mesh.Attrib.Vec2, Mesh.Attrib.Color);
     }
 
-    @Init(stage = InitStage.Pre)
+    @Init(stage = InitStage.Pre, dependencies = Shaders.class)
```
Paul1365972, line 31: It was just never an issue, because dependencies were never used up until now, but Renderer2D definitely depends on Shaders, I probably missed some other init functions that have dependencies, but I guess it works for now

Issue #1883, opened 2022-01-04
[Suggestion] Vein Miner improvements


RickyTheRacc:
### Describe the feature

Allow the block selections to act as either a whitelist or a blacklist and add a delay between mining blocks

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

RickyTheRacc:
Please at least add the black/whitelist option I have to spend like 5 minutes spam clicking to remove all of the unnecessary blocks that I don't want it to mine and that's actually so annoying

RickyTheRacc:
The good ending 😌

Issue #1825, opened 2021-12-21
[Question] How to stop the zoom module from announcing it's been enabled & disabled


Decme:
### Describe the feature

If this isn't possible, could this be considered a feature request? I zoom in/out a lot and it's very spammy.

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

arlomcwalter:
gui > config tab > chat feedback

RedCarlos26:
@arlomcwalter I think he wanted to say an option to only disable the chat feedback for certain modules like zoom, since it becomes very spammy. The thing is having chat feedback on is useful, but the zoom module can make it spammy

Decme:
I see this has been closed. Has this been implemented?

RedCarlos26:
No

bluepanee:
> I see this has been closed. Has this been implemented?

I added this feature to my pull request. It will be implemented if everything is ok.

RedCarlos26:
Based thanks a lot

Issue #1800, opened 2021-12-15
Sneaking in Freecam via the Sneak module


Lence2b2t:
### Describe the feature

Currently, enabling Freecam while the Sneak module is enabled, you stop shifting; my suggestion is for there to be a change implemented that makes it so the Sneak module functions while in Freecam.

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

DesiCow:
Someone forgot to close this issue 👀

PR #1999, opened 2022-01-29
Smol changes and some improvements.

bluepanee:
Changes:
 - **VeinMiner** module has whitelist mode & delay slider now. #1883 
image

 - **Sneak** can works together with **Freecam**. #1800

 - Ability to disable chat feedback for certain modules. No more spam, Zoom-ers. #1825
image

 - In the **Zoom** module, you can now turn off the dumb hands rendering.
image


src/main/java/meteordevelopment/meteorclient/systems/modules/world/VeinMiner.java
```diff
@@ -139,7 +151,14 @@ private boolean isMiningBlock(BlockPos pos) {
     @EventHandler
     private void onStartBreakingBlock(StartBreakingBlockEvent event) {
         BlockState state = mc.world.getBlockState(event.blockPos);
-        if (state.getHardness(mc.world, event.blockPos) < 0 || blacklist.get().contains(state.getBlock())) return;
+
+        if (mode.get() == ListMode.Whitelist) {
+            if (!selectedBlocks.get().contains(state.getBlock()))
```
C10udburst, line 156: the `state.getHardness(mc.world, event.blockPos) < 0` check should be done regardless if the mode is on whitelist or blacklist

C10udburst (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/world/VeinMiner.java
bluepanee: But why? If I whitelist a block that has a hardness less than zero, doesn't that mean I intend to break it? Of course, if I'm in creative mode or have the ability to do so.

RedCarlos26:
Based pr

src/main/java/meteordevelopment/meteorclient/systems/modules/world/VeinMiner.java
bluepanee: Done.

RickyTheRacc:
please add this I beg you

RedCarlos26:
> please add this I beg you

Yes I beg as well

arlomcwalter:
i dont beg

arlomcwalter:
the state of meteor client

Bluscream:
I demand

RedCarlos26:
I mean why would it not be added. It's so useful

RickyTheRacc:
> the state of meteor client

I DEMAND AND REQUIRE that you merge this pr now. Also hello seasnail

arlomcwalter:
> > the state of meteor client
> 
> 
> 
> I DEMAND AND REQUIRE that you merge this pr now. Also hello seasnail

hello

Wide-Cat:
wtf is that the real beabail

MineGame159:
1. Why is the chat feedback field named `toggleChatFeedback` in NBT and not just `chatFeedback`
2. I think by default chat feedback should be enabled for modules
3. The check for module's chat feedback setting should be moved to the `Module.sendToggleMsg()` method instead it being checked manually every time that method is called

Apart from these things it looks fine

arlomcwalter:
> wtf is that the real beabail

possibly

bluepanee:
> 1. Why is the chat feedback field named `toggleChatFeedback` in NBT and not just `chatFeedback`
> 2. I think by default chat feedback should be enabled for all modules
> 3. The check for module's chat feedback setting should be moved to the `Module.sendToggleMsg()` method instead of it being checked manually every time that method is called
> 
> Apart from these things it looks fine

Done.

RedCarlos26:
Less go!!

arlomcwalter:
the state of meteor client

PR #2044, opened 2022-02-10
hand view: fix old settings and add new ones

Wide-Cat:
- Fix the old settings having invalid and conflicting names (e.g. "Scale X" -> "mainhand-scale-x")
- Add an option for your hands to rotate to follow your serverside rotations
- Add a seperate settings group for changing how your arm appears


Wide-Cat:
💪

necrophili:
so da settings doesn't have to be formatted in old metro tradition anymorr?

MineGame159:
? the pr literally fixed the names so they comply with the standard.

MineGame159:
OH you meant that. Yes, that is the tradition that should be generally followed but in this particular case I will allow it just because having hundreds of lines for settings is not really pleasant to work with.

PR #2055, opened 2022-02-15
Add autosave feature for profiles

SnivyLlama:
Ok this was like, the one thing bothering me about meteor so I decided to fix it myself
Lemme know if I should change it


SIMULATAN (changes requested):
👍 Very cool but i'd also add a scheduled task to save it every x seconds because most of the time the issue is that the game crashes without you saving the config. Sessions on Servers can be long (2b2t for example) and thus this function wouldn't be called too often

SnivyLlama:
Alright, I think I can get this done

SnivyLlama:
Alright, I got it

SnivyLlama:
Should I remake this pr with one pull request so that it's just one commit?

SnivyLlama:
yeah ill do that

SIMULATAN:
> Should I remake this pr with one pull request so that it's just one commit?

squash merge exists?

PR #2064, opened 2022-02-16
Add proxy importing, fix proxy issues (see desc.) - closes #1124

### LINKED ISSUE #1124: [Suggestion] Alt list/Proxy List
Add a feature to import proxies as a ip:port list, and alts as a username:password list.

Considering we have things like the swarm module, it's kind of absurd that we have to manually input proxies and alts.

SIMULATAN:
- fix proxy selection bug
- fix ability to add proxies with 0 as the port or no address
- fix showing empty proxy names

Damn, that's a huge update!

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
@@ -77,6 +77,6 @@ public class ProxiesImportScreen extends WindowScreen {
         }
         add(theme.horizontalSeparator()).expandX();
         WButton btnBack = add(theme.button("Back")).expandX().widget();
-        btnBack.action = () -> MeteorClient.mc.setScreen(parent);
+        btnBack.action = this::onClose;
     }
 }
--- a/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
@@ -1,0 +1,82 @@
+/*
+ * This file is part of the Meteor Client distribution (https://github.com/MeteorDevelopment/meteor-client/).
+ * Copyright (c) 2022 Meteor Development.
+ */
+
+package meteordevelopment.meteorclient.gui.screens;
+
+import meteordevelopment.meteorclient.MeteorClient;
+import meteordevelopment.meteorclient.gui.GuiTheme;
+import meteordevelopment.meteorclient.gui.WidgetScreen;
+import meteordevelopment.meteorclient.gui.WindowScreen;
+import meteordevelopment.meteorclient.gui.widgets.containers.WSection;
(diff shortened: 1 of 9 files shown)


src/main/java/meteordevelopment/meteorclient/utils/render/color/Color.java
```diff
@@ -10,7 +10,36 @@
 import net.minecraft.nbt.NbtCompound;
 import net.minecraft.util.math.Vec3d;
 
+@SuppressWarnings("unused")
 public class Color implements ICopyable<Color>, ISerializable<Color> {
+
+    public static final Color white = new Color(java.awt.Color.WHITE);
+    public static final Color WHITE = white;
+    public static final Color lightGray = new Color(java.awt.Color.LIGHT_GRAY);
+    public static final Color LIGHT_GRAY = lightGray;
```
C10udburst, line 19: is there a reason why there are two variables for each color?

SIMULATAN: It's to follow along the java.awt format

MineGame159: Leave just the uppercase ones.

SIMULATAN: Done in e6dd40f

SIMULATAN: If you're fine with how it looks, just squash merge it

src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
```diff
@@ -0,0 +1,82 @@
...
+                );
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        } else {
+            add(theme.label("Invalid File!"));
+        }
+        add(theme.horizontalSeparator()).expandX();
+        WButton btnBack = add(theme.button("Back")).expandX().widget();
+        btnBack.action = () -> MeteorClient.mc.setScreen(parent);
```
MineGame159, line 80: While this works to close the screen it is better practise to call the `onClose()` method.

Sorry for not pointing it out earlier.

MineGame159 (changes requested):

src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
SIMULATAN: `onClose` sounded like the close listener which is why I didn't use it. Also, there are other parts in the code where I found this method of switching screens which is why I used it.

MineGame159: Oh

SIMULATAN:
thanks mimgam, very cool

PR #2077, opened 2022-02-21
Clean up mixins.

Wide-Cat:
Reorganised the mixins from almost 150 files in one package to a structure like that of how the files are actually organised in minecraft themslef.


C10udburst:
This branch has conflicts that must be resolved. @Wide-Cat

PR #2079, opened 2022-02-22
breaking block hud element hide while not breaking

nothub:
Adds a setting to the `breaking-block` hud element to hide it while there is no block breaking activity.


MineGame159:
1. You should always reset the `visible` flag to true, not only when in editor because someone can change the setting while its false and it will never be set to true this way
2. Seeing as it is pretty old with no activity I will close it and implement it into my HUD rewrite which will hopefully come out soon

PR #2090, opened 2022-02-24
Various bugfixes and improvements

Wide-Cat:
- fixed small error in the world renderer mixin
- add an option in arrow dodge to ignore your own projectiles
- lowered alpha on surround rendering
- add an option in potion spoof to clear status effects on the module disable
- fixed break indicators not rendering the first stage of your block breaking progress

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BreakIndicators.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BreakIndicators.java
@@ -45,17 +45,17 @@ public class BreakIndicators extends Module {
 
 
     private final Setting<SettingColor> startColor = sgGeneral.add(new ColorSetting.Builder()
-            .name("start-color")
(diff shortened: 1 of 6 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/render/BreakIndicators.java
```diff
@@ -80,9 +80,24 @@ private void renderNormal(Render3DEvent event) {
...
+            double shrinkFactor = 1d - ownBreakingStage;
+            double progress = 1d - shrinkFactor;
+
+            renderBlock(event, box, orig, ownBreakingPos, shrinkFactor, progress);
+        }
+
         blocks.values().forEach(info -> {
             BlockPos pos = info.getPos();
             int stage = info.getStage();
+            if (pos == ownBreakingPos) return;
```
MineGame159, line 100: You should probably use `.equals()` even if it works currently just to be safe.

PR #2118, opened 2022-03-05
XRayBruteforce module

Nekiplay:
image


MineGame159:
I wonder how this can be used on anarchy 🤔

Nekiplay:
> I wonder how this can be used on anarchy 🤔
> 
> EDIT: Also the code is terrible

There are servers with XRay protection

Lovecrafts-Cat:
> There are servers with XRay protection

Then it's not anarchy, and it would go against our rules.

PR #2135, opened 2022-03-10
[ESP] Different colors for each entity

K-u00E4-t:


K-u00E4-t:
I'll get rid of the errors and try to beautify the code, i'll make another pr then since i can't change the files of this one :|

PR #2136, opened 2022-03-10
[ESP] Different colors for each entity

K-u00E4-t:
Fixed a bug and beautified some code this time


JFronny:
You can add multiple commits to a single PR to fix issues from the original commit btw

PR #2173, opened 2022-03-20
fixed bug in shader system

FlorianMichael:
When the ShaderSystem is used in screens, errors occur because the Rendering3D in the Shader System is not deactivated and therefore sometimes the rendering shifts up or down.


MineGame159:
I am not aware of any rendering problems in screens, can you show an example?

FlorianMichael:
It is not a real meteor issue because it does not occur in the original, but if you work with the render system in screens in a fork, this error occurs (e.g. drawing circles).

MineGame159:
If this "fix" is needed it means the code used to render it is broken, it should ALWAYS restore the projection and not depend on a weird fix that runs on tick.

PR #2211, opened 2022-04-06
Send messages before AutoLog does its thing

SnivyLlama:
Haven't tested and I don't think I will


arlomcwalter:
there are always two duplicate lines after calling sendFinalRemarks, you could just put them at the end of that method to clean up the duplicated code

PR #2220, opened 2022-04-11
Fix MineGame159

RickyTheRacc:
Cope seethe mald


arlomcwalter:
this is such a stupid way to do it

RickyTheRacc:
Well the current one doesn't work at all, so it's an objective improvement :/

Declipsonator:
Wasn't DeathMessageS2CPacket not reliable when looking for alternative ways to see if a player died

Declipsonator:
I thought that came up around notifier

RickyTheRacc:
I mean I put the same thing in rickyware for a toggle on death setting and it hasn't failed once so far

arlomcwalter:
> I mean, the current way just doesn't work at all, so this is objectively an improvement

that doesn't make it good

RickyTheRacc:
> that doesn't make it good

No, but it makes it better

SIMULATAN:
>Build failed

PR #2224, opened 2022-04-17
Implement .nbt paste command

JFronny:
When editing NBT data through the included command-based interface, it is common to run into the limitation on command length in vanilla.
This PR allow bypassing that limitation by copying the data to a separate editor through the preexisting `.nbt copy` command, modifying the data (and perhaps saving it for later reuse) and reapplying it to the item in question through the added `.nbt paste` command without needing to paste it into the chat input box which was required previously.
To achieve this, the existing NBT parsing in CompoundNbtTagArgumentType is reused via an instantiation of said class, however, moving it to a separate, static method is possible if this implementation is an issue.


arlomcwalter:
is this safe? it doesn't seem to validate the nbt but idk how .setNbt handles invalid nbt so it might be fine

JFronny:
I tested inputting some random java code i had in my clipboard and I got an error message as one would expect so validation seems to be done somewhere

arlomcwalter:
cool

PR #2264, opened 2022-05-04
Added more anticheats to plugins command

erenkarakal:


Lovecrafts-Cat:
Isn't there a list we could fetch to update these?

Also, you can add some more form here if you want:
https://www.spigotmc.org/wiki/anti-cheat-list-bukkit-and-spigot-1-8-x/
https://www.spigotmc.org/wiki/anti-cheat-list-bukkit-and-spigot-1-18/

erenkarakal:
nobody uses stuff other than these

PR #2278, opened 2022-05-09
Added powder Snow Bucket mode to No Fall

Hyped1XGeek:
Added a new mode to No Fall Module i.e. PowderSnow mode to mlg in Nether

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/NoFall.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/NoFall.java
@@ -25,6 +25,7 @@ import meteordevelopment.meteorclient.utils.player.PlayerUtils;
 import meteordevelopment.meteorclient.utils.player.Rotations;
 import meteordevelopment.meteorclient.utils.world.BlockUtils;
 import meteordevelopment.orbit.EventHandler;
+import net.minecraft.block.Blocks;
 import net.minecraft.fluid.Fluids;
 import net.minecraft.item.BlockItem;
 import net.minecraft.item.Items;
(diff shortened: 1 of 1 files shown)


arlomcwalter:
just include it in Mode.bucket and check which dimension you are in

Hyped1XGeek:
> just include it in Mode.bucket and check which dimension you are in

But someone might want it separately in cases like mlg on top half slab, top trapdoor, or stairs etc

In some cases when player has momentum it may fail to pick up water as soon as he lands so it may spread and may destroy some water-flushable blocks(like redstone

arlomcwalter:
> > just include it in Mode.bucket and check which dimension you are in
> 
> 
> 
> why? maybe someone wants to use powdered snow in overworld
> 
> 

well sure but at least don't place water in the nether. the reasoning is that you wouldn't want to swap modes when you change dim. instead it could check if you have water and you aren't in the nether and place water and then if not it would look for other fallbacks like bales or snow or whatever. my point is, it would be good to provide some kind of automation so you don't have to reconfigure when you change dimension

arlomcwalter:
> > just include it in Mode.bucket and check which dimension you are in
> 
> 
> 
> why? maybe someone wants to use powdered snow in overworld
> 
> 

well sure but at least don't place water in the nether. the reasoning is that you wouldn't want to swap modes when you change dim. instead it could check if you have water and you aren't in the nether and place water and then if not it would look for other fallbacks like bales or snow or whatever. my point is, it would be good to provide some kind of automation so you don't have to reconfigure when you change dimension

arlomcwalter:
> > just include it in Mode.bucket and check which dimension you are in
> 
> 
> 
> But someone might want it separately in cases like mlg on top half slab, top trapdoor, or stairs etc
> 
> 
> 
> In some cases when player has momentum it may fail to pick up water as soon as he lands so it may spread and may destroy some water-flushable blocks(like redstone) 

could just add a check to see if the water is still there a few ticks later to make sure it is successful

MineGame159:
@Hyped1XGeek are you gonna fix the code?

PR #2286, opened 2022-05-13
Add prototype of a custom stat hud element

maxsupermanhd:
TODO:
- Figure why `Stats.CUSTOM.getOrCreateStat` hits a null pointer even when identifier is correct (help wanted)


C10udburst:
> Send statistic requests (no idea how to do that, help wanted) 

Looking at wiki.vg [Client Status](https://wiki.vg/Protocol#Client_Status) packet might be what you want.
In fabric its called `net.minecraft.network.packet.c2s.play.ClientStatusC2SPacket`

maxsupermanhd:
> > Send statistic requests (no idea how to do that, help wanted)
> 
> Looking at wiki.vg, [Client Status](https://wiki.vg/Protocol#Client_Status) packet might be what you want. In fabric its called `net.minecraft.network.packet.c2s.play.ClientStatusC2SPacket`

Yes, but from what method/class do I send it? HudElement is not a module or something, it does not have onTick...

arlomcwalter:
it has an update method and you can just make it call it every like 5 seconds or something

maxsupermanhd:
Yes, how to update it is actually not a problem, I figured how to do that now. Real problem is how to select statistic that you want to display, really have no idea how to implement it.

Wide-Cat:
With the most recent changes to the hud and starscript, unfortunatley this has been broken and made redundant.

PR #2318, opened 2022-05-24
Capitalization and added collapsible text fields

CodeDoc537:


CodeDoc537:
just a proposal

MineGame159:
I think it is clearer with all the text displayed. There isn't so much of it that it would be too much.

CodeDoc537:
I agree, I was mostly just trying out things I found.

PR #2326, opened 2022-05-25
Update to match the new 1.18 height limit

CodeDoc537:
changed the max slider range in Anchor to match the new 128 blocks higher height limit in 1.18+

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Anchor.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Anchor.java
@@ -25,7 +25,7 @@ public class Anchor extends Module {
         .name("max-height")
         .description("The maximum height Anchor will work at.")
         .defaultValue(10)
-        .range(0, 255)
+        .range(0, 383)
         .sliderMax(20)
         .build()
     );


src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Anchor.java
```diff
@@ -25,7 +25,7 @@ public class Anchor extends Module {
         .name("max-height")
         .description("The maximum height Anchor will work at.")
         .defaultValue(10)
-        .range(0, 255)
+        .range(0, 383)
```
C10udburst, line 28: im pretty sure this is wrong.

1.18 coordinates are from `-64` to `320`

C10udburst (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Anchor.java
CodeDoc537: pre 1.18 was up to 256, so its the total height limit -1 so I just did the new height -1
320 + 64 = 384, 384 - 1 = 383

arlomcwalter: > im pretty sure this is wrong.
> 
> 
> 
> 1.18 coordinates are from `-64` to `320`

-64 to 320 and therefore its possible to be 383 (well more obviously) blocks above the bottom of the world

arlomcwalter: as in it is relative to the floor beneath you (iirc)

soshimee:
It really depends on the server you are playing on though, since it can be changed.

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Anchor.java
CodeDoc537: exactly

CodeDoc537:
> It really depends on the server you are playing on though, since it can be changed.

well yeah, but its nice just to have the option

offbeat-stuff:
```java
mc.world.getBottomY()
mc.world.getTopY()
```

arlomcwalter:
its literally nothing to do with world height

arlomcwalter:
> ```java
> 
> mc.world.getBottomY()
> 
> mc.world.getTopY()
> 
> ```
> 
> Can you use these two values

you can, but its nothing to do with this

arlomcwalter:
> It really depends on the server you are playing on though, since it can be changed.

literally this is a relative value, there is no way you'd ever want to have it over like 20 but anyway its nothing to do with world height

PR #2420, opened 2022-06-18
resolve #2395, #2396

### LINKED ISSUE #2395: [Bug] Banner preview crash
### Describe the bug

Holding left alt to preview this banner causes a client crash
image
No additional nbt tags, server 1.18.2 with VIA
[crash-2022-06-15_13.39.46-client.txt](https://github.com/MeteorDevelopment/meteor-client/files/8908559/crash-2022-06-15_13.39.46-client.txt)

Funny is that it likely a VIA bug because preview of same item works fine in singleplayer....

### Steps to reproduce

```bash
Preview a banner
```

### Meteor Version

dev 1501

### Minecraft Version

1.19

### Operating System

Linux

- [X] This is a valid bug.

### LINKED ISSUE #2396: [Suggestion] Make Autotool use swords on bamboo
### Describe the feature

Its just a tiny little thing but swords are the most effective tool to mine/cut bamboo with since swords instantly break bamboo

Wide-Cat:
resolves #2395
resolves #2396

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BetterTooltips.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BetterTooltips.java
@@ -80,7 +80,7 @@ public class BetterTooltips extends Module {
     );
 
     public final Setting<Boolean> alwaysShow = sgGeneral.add(new BoolSetting.Builder()
-        .name("bypass-'hideflag'-nbt")
+        .name("always-show")
         .description("Disables the HideFlags nbt tag.")
         .defaultValue(true)
         .build()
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/render/BetterTooltips.java
```diff
@@ -80,7 +80,7 @@ public class BetterTooltips extends Module {
     );
 
     public final Setting<Boolean> alwaysShow = sgGeneral.add(new BoolSetting.Builder()
-        .name("always-show")
+        .name("bypass-'hideflag'-nbt")
```
MineGame159, line 83: I think the old name is better, a normal user doesn't know what that means and in this case the description won't help either.

arlomcwalter: cap, the name "always show" is so ambiguous

Issue #2382, opened 2022-06-13
Experience Tweaks


Olioq:
### Describe the feature

I think the usage of Experience Bottles could be improved by adding XP Bottles as an option to Middle Click Extra, so it takes less time to mend armor (It can mend silently or just switchback). Another feature that would be nice is FeetXP. This setting will silently look towards your feet when throwing experience bottles so you will pick up all the xp and not have to look down manually each time.

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

Wide-Cat:
has been added

PR #2430, opened 2022-06-20
Do things

### LINKED ISSUE #2383: [Bug] Crash at startup
### Describe the bug

[crash-2022-06-14_13.28.10-client.txt](https://github.com/MeteorDevelopment/meteor-client/files/8897189/crash-2022-06-14_13.28.10-client.txt)

### Steps to reproduce

```bash
Launch the client
```

### Meteor Version

dev 1496

### Minecraft Version

19

### Operating System

Windows

- [X] This is a valid bug.

RacoonDog:
Fix #2383 
Adds the MiddleClickExtra part of #2382 
Update BetterChat for 1.19 compat + performance stuff

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/BedAura.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/BedAura.java
@@ -152,7 +152,7 @@ public class BedAura extends Module {
 
     private final Setting<Boolean> swing = sgRender.add(new BoolSetting.Builder()
         .name("swing")
-        .description("Whether to swing hand clientside clientside.")
+        .description("Whether to swing hand client-side.")
         .defaultValue(true)
         .build()
     );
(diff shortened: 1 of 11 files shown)


RacoonDog:
Ignore commits 1 and 3, they just github being github bc the fork i used is the same as my earlier pr

RacoonDog:
`meteordevelopment.meteorclient.systems.proxies.Proxies` has a regex pattern that starts with `^` and ends with `$`, this is redundant as it's matcher uses `Matcher.matches()` which tests against the whole string by default.

`meteordevelopment.meteorclient.utils.player.ChatUtils` stores custom prefixes as a list of pairs containing a string and the prefix, where when used it loops through each pair testing the package name against the pair's string.
This can easily be replaced with a `HashMap` where the key would be the string (package name) and value the prefix.
This would improve performance as looping through a list is O(n) while searching in a hashmap is O(1).
Only *downside* would be duplicate keys would no longer be accepted, but the code for the custom prefixes stops at the first match, so that is irrelevant.

MineGame159:
For the ChatUtils thing it uses `startsWith` and doesn't just check for equality.

RacoonDog:
gonna clean this up in another pr

PR #2439, opened 2022-06-21
new module and improvment

MaslovKK:
all commits has a comments


Talhoid:
i think you mean new module not "hack"

RacoonDog (changes requested):
Max jump is active even when Entity Control is not

MaslovKK:
> Max jump is active even when Entity Control is not

ur'e wrong, i'm checked it's now - everything is fine

katant (changes requested):
хуйня переделывай

ne-lanskoy:
а ты знал что твоя мама шлюха?

Wide-Cat:
With ignore border enabled, I can break blocks and such beyond the world border, but can't actually walk through it. Entity control seems to work much the same as before.

MaslovKK:
> With ignore border enabled, I can break blocks and such beyond the world border, but can't actually walk through it. ~Entity control seems to work much the same as before~ I think a desync bug with entity control is fixed, which is epic

U can walk through it, is u play vanilla or spigot?

Wide-Cat:
i was on singleplayer

MaslovKK:
5 months ignoring pr's. Best responsiveness!

ThebestkillerTBK:
thank god

PR #2442, opened 2022-06-22
Remove Blink

RickyTheRacc:
Should also have never been added


I-Got-You-Dead:
ik you think its retarded, but no ;-;
its good for trolling.

RacoonDog:
😐

RickyTheRacc:
:cope:

PR #2458, opened 2022-06-27
Fix freelook

Mersid:
This patch fixes an issue in the free look module wherein camera rotation in first person mode rotates the player instead. With this, the user can rotate their camera in first person mode, making it possible with some configuration possible to emulate ArmA 3's look-around option without having to enter third person mode.


Mersid:
Why are there so many junk commits

PR #2468, opened 2022-07-02
add armor storage

19MisterX98:


I-Got-You-Dead:
So… auto sneak but only when within the range of a sculk sensor?

MineGame159:
Just one small thing, could you make it work in creative inventory too?

MineGame159:
Oh also, fix the setting name to be `armor-storage`.

PR #2481, opened 2022-07-07
Fixed a wrong descripton in Notifier

RedCarlos26:


RedCarlos26:
How can I make it so it only requests the last commit which is [bd3f787](https://github.com/MeteorDevelopment/meteor-client/pull/2481/commits/bd3f787a429f2337be471b5d5c111cd08b452bcc)?

PR #2514, opened 2022-07-10
Fix PotionTimers render in world

WERooo:
Fixing a bug with potion timers when you open the hud editor in the world


Wide-Cat:
thanks for this fix, i noticed this error earlier yesterday but didnt get around to fixing it

Issue #2202, opened 2022-04-02
[Bug] Nametag does not working when forcing unicode


nosashuumai:
### Describe the bug

![スクリーンショット 2022-04-02 103403](https://user-images.githubusercontent.com/84297088/161360458-82f16be7-98b0-42c6-8f32-6a87d29a2b4b.png)

### Steps to reproduce

```bash
Enforce unicode and use name tags
```

### Meteor Version

0.4.8 1459

### Minecraft Version

1.18.2

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported.
(I have checked every bug report on github)

- [X] This is a valid bug.
(I am able to reproduce this on the latest dev build)

nosashuumai:
sorry for my bad english

nosashuumai:
@MineGame159 help me

PR #2518, opened 2022-07-11
little code clean up and fixes

WERooo:
fixes bugs like #2202
fixes outline bottom line not rendered when only one module enabled
cleaned up some modules code


RickyTheRacc:
Minegame?

MineGame159:
This PR breaks XRay with Fabric Api (no Sodium). Apart from that looks good.

WERooo:
in docs to mixin it says that remap is default to true

MineGame159:
Yes but if remap is set to false on the class itself the injections are false by default too.

WERooo:
i just seen previous commits and didnt check that. thanks.

PR #2540, opened 2022-07-16
RichPresence addon support

RacoonDog:
probably not the best code ever, am open to suggestions


MineGame159:
Put a `break;` statement in the loop when it finds a valid state. Apart from that looks good.

seasnail8169:
can you also move the static block to between the list and the constructor

Issue #2550, opened 2022-07-19
[Bug] Auto eat does not pause baritone unless actively pathing


pwnker:
### Describe the bug

[Auto eat](https://github.com/MeteorDevelopment/meteor-client/blob/3775c77ca0eb39029019ee1578602cb239b6075e/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoEat.java#L161-L167) pauses baritone but only if it is pathing which is not always the case. This is blatantly apparent when using baritone selection commands such as `#sel replace` where it doesnt appear to path much at all causing auto eat to fight with baritone for extended periods of time.
```js        
// Pause baritone
        wasBaritone = false;
        if (pauseBaritone.get() && BaritoneAPI.getProvider().getPrimaryBaritone().getPathingBehavior().isPathing()) {
            wasBaritone = true;
            BaritoneAPI.getProvider().getPrimaryBaritone().getCommandManager().execute("pause");
        }
    }
```

### Steps to reproduce

```bash
1. Enable auto eat and use a baritone selection command such as `#sel replace`.
2. Loose hunger.
3. If not pathing baritone will not be paused.
```

### Meteor Version

0.5.0

### Minecraft Version

Fabric 1.19

### Operating System

Windows

PR #2554, opened 2022-07-20
Pause on Baritone now also works when not pathing for KillAura and BowAimbot

RedCarlos26:


RedCarlos26:
Oops I think I forgot to update my gradle dependencies

RedCarlos26:
Why is there apparently issues not even related to my code changes 🤔

RedCarlos26:
Should I also rename the `wasPathing` variables to `wasBaritone` for more consistency?

pwnker:
> Should I also rename the `wasPathing` variables to `wasBaritone` for more consistency, especially that now it doesn't pause only when pathing?

Yeah go for it.

pwnker:
This can be merged, #2550 applies to these modules too.

pwnker (approved):
This can be merged, #2550 applies to these modules too.

MineGame159:
Why tho

RedCarlos26:
When using baritone to clear an area it's useful when it can actually pause to attack the entity then go back to work

MineGame159:
Then the check should be improved instead of fully removing it.

RedCarlos26:
Why improved? It's useless... Whtever try e process it we want to pause baritone when this is enabled like what?

pwnker:
> Then the check should be improved instead of fully removing it.

Yeah thats valid if the check isn't useless. The check checks wether baritone is on and pathing but why would would it make a difference if you are pathing. You don't want to be swinging your pickaxe trying to use aura or autoeat. This PR is an extension of #2551.

RedCarlos26:
@MineGame159 what do you mean by improving the check btw

PR #2602, opened 2022-08-03
Performance-related changes

RacoonDog:
### Use better hashmap impl
Benchmark
arrayMap_getOrDefault - 1578.984 ± 20.332  ops/ms
arrayMap_ifContainsGet - 1053.077 ±  5.827  ops/ms
openHashMap_getOrDefault - 2447.558 ± 69.816  ops/ms
openHashMap_ifContainsGet - 2290.297 ± 33.869  ops/ms

OpenHashMap using getOrDefault is **~132.42 ± 7.92%** faster than the current impl.

[Benchmark source](https://pastebin.com/brmTfScM)


Floof8952:
joe biden.

PR #2644, opened 2022-08-09
origins compat

RacoonDog:
kinda hate the fact i had to make 3 injects optional, but this makes origins compatible w meteor

worst case scenario: it crashes on startup
slightly better worst case scenario: it breaks reach and bettertooltips


seasnail8169:
also look at how existing origins compat is handled

Issue #1950, opened 2022-01-15
[Suggestion] Anti-Blindness


Theboiboi8:
### Describe the feature

based utility client feature fr fr also maybe add it to No Render settings because only nausea is in there rn

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

SIMULATAN:
Check the NoRender mod

Theboiboi8:
> Check the NoRender mod

I personally cannot find any such options in the no-render options

bluepanee:
> > Check the NoRender mod
> 
> I personally cannot find any such options in the no-render options

It's called fog.

bluepanee:
> Fog is enabled in norender and the blindness effect still occurs

Ah, I get it. Probably sodium issue if you have it installed. #1705 #1696

Wide-Cat:
this should be resolved

Issue #1705, opened 2021-11-24
Fix NoRender blindness for Sodium [Suggestion] 


Texbio:
### Describe the feature

Based on #1696 I was told I have to ask for this to be added as a suggestion.
The suggestion would be to make blindness and the water overlay work on sodium.
I am using a mac with the latest 1.17.1 dev build of sodium (https://github.com/CaffeineMC/sodium-fabric/actions/runs/1495670671). I tried it on other versions of sodium and cant get rid of blindness/water overlay. 
Also when distortion effects are set to 0%, the nausea overlay still appears but can be easily removed using a texturepack. (report as a bug?)

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

Wide-Cat:
this should be resolved

PR #2650, opened 2022-08-11
NoBlindness/NoDarkness in NoRender

### LINKED ISSUE #2242: No Blindness[Suggestion]
### Describe the feature

It would be a module that allows you to remove the blindness effect. So when you have it, there is no overlay.

ThebestkillerTBK:
Resolves #2242 #1950 #1705

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/BackgroundRendererMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/BackgroundRendererMixin.java
@@ -30,7 +30,7 @@ public class BackgroundRendererMixin {
         }
     }
 
-    @Inject(method = {"getFogModifier(Lnet/minecraft/entity/Entity;F)Lnet/minecraft/client/render/BackgroundRenderer$StatusEffectFogModifier;"}, at = @At("HEAD"), cancellable = true)
+    @Inject(method = "getFogModifier(Lnet/minecraft/entity/Entity;F)Lnet/minecraft/client/render/BackgroundRenderer$StatusEffectFogModifier;", at = @At("HEAD"), cancellable = true)
     private static void onGetFogModifier(Entity entity, float tickDelta, CallbackInfoReturnable<Object> info) {
         if (Modules.get().get(NoRender.class).noBlindness()) info.setReturnValue(null);
     }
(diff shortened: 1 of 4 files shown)


Wide-Cat:
correct me if im wrong but the fog setting in norender already removes blindness and darkness?

ThebestkillerTBK:
This PR fixes the black sky issue and sodium issue

src/main/java/meteordevelopment/meteorclient/mixin/BackgroundRendererMixin.java
```diff
@@ -27,4 +29,9 @@ private static void onApplyFog(Camera camera, BackgroundRenderer.FogType fogType
             }
         }
     }
+
+    @Inject(method = {"getFogModifier(Lnet/minecraft/entity/Entity;F)Lnet/minecraft/client/render/BackgroundRenderer$StatusEffectFogModifier;"}, at = @At("HEAD"), cancellable = true)
+    private static void onGetFogModifier(Entity entity, float tickDelta, CallbackInfoReturnable<Object> info) {
+        if (Modules.get().get(NoRender.class).noBlindness()) info.setReturnValue(null);
```
SIMULATAN, line 35: Why are there two settings if they both do literally the same?

SIMULATAN (changes requested):
Small weird design choice that imo needs clarification but otherwise LGTM

src/main/java/meteordevelopment/meteorclient/mixin/BackgroundRendererMixin.java
ThebestkillerTBK: It only removes blindness fog

SIMULATAN: Oh, accidentally overlooked these other things and commented on the wrong thing lol, sorry for the wrong review
Doing literally anything on phones is just so terrible :>(

SIMULATAN (approved):

seasnail8169:
can you make all of your mixin method targets into strings rather than arrays

ThebestkillerTBK:
> can you make all of your mixin method targets into strings rather than arrays

fixed

PR #2654, opened 2022-08-13
addons can have custom main menu credits

RacoonDog:
Also lets them add relevant info to other mod's credits by making Credit and Section public

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/player/TitleScreenCredits.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/player/TitleScreenCredits.java
@@ -16,9 +16,7 @@ import meteordevelopment.meteorclient.utils.network.MeteorExecutor;
 import meteordevelopment.meteorclient.utils.render.color.Color;
 import net.minecraft.client.util.math.MatrixStack;
 
-import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


C10udburst:
cant u just edit the `fabric.mod.json`?

RacoonDog:
> cant u just edit the `fabric.mod.json`?

main reason i wanted this is for the use case in the pr’s description where i can easily append info to other mods’ credits via mixin

MineGame159:
Why tho, what custom credits do you have in mind?

RacoonDog:
im basically just trying to make an addon that lets outdated meteor addons be used on current meteor, and similar to how it shows there's an update on the credit i'd use that system to show which addons are outdated and the user wouldn't get support for if an error happens.

so yeah technically for my use case the `customCredits` map is useless but i thought it'd be a nice addition

MineGame159:
About this pr:
1. From API point of view giving the users access to a field is not a good idea.
2. I don't think there is any need for completely custom credits, if someone has a valid usecase for it then we can make something better than this.

About your addon:
What the fuck? You realise the addons are outdated for a reason? Like for example some breaking change happened in Meteor thus the addon needed to be changed, how do you want to make them compatible?

PR #2713, opened 2022-09-03
more error handling

RacoonDog:
notifies the player if they're using an addon that is outdated
adds a button to turn of serverspoof on the disconnect screen
actually useful error reporting in ReflectInit

should lower the amount of useles issues

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/DisconnectedScreenMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/DisconnectedScreenMixin.java
@@ -49,11 +49,7 @@ public abstract class DisconnectedScreenMixin extends Screen {
                 button -> {
                     Modules.get().get(AutoReconnect.class).toggle();
                     time = Modules.get().get(AutoReconnect.class).time.get() * 20;
-                    }));
-            if (Modules.get().get(ServerSpoof.class).noSignatures()) addDrawableChild(new ButtonWidget(x, y + 44, 200, 20, Text.literal("Deactivate ServerSpoof & Reconnect"), button -> {
-                Modules.get().get(ServerSpoof.class).toggle();
-                reconnectBtn.onPress();
-            }));
+                }));
         }
     }
 
--- a/src/main/java/meteordevelopment/meteorclient/mixin/DisconnectedScreenMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/DisconnectedScreenMixin.java
@@ -5,8 +5,10 @@
 
 package meteordevelopment.meteorclient.mixin;
 
+import meteordevelopment.meteorclient.systems.modules.Module;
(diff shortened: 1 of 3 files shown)


src/main/java/meteordevelopment/meteorclient/mixin/DisconnectedScreenMixin.java
```diff
@@ -44,11 +45,15 @@ private void onRenderBackground(CallbackInfo info) {
...
                 ServerAddress.parse(Modules.get().get(AutoReconnect.class).lastServerInfo.address),
                 Modules.get().get(AutoReconnect.class).lastServerInfo)));
-            toggleAutoReconnectBtn = addDrawableChild(new ButtonWidget(x, y + 22, 200, 20, Text.literal("Toggle AutoReconnect"),
+            addDrawableChild(new ButtonWidget(x, y + 22, 200, 20, Text.literal("Toggle AutoReconnect"),
                 button -> {
                     Modules.get().get(AutoReconnect.class).toggle();
                     time = Modules.get().get(AutoReconnect.class).time.get() * 20;
                     }));
+            if (Modules.get().get(ServerSpoof.class).noSignatures()) addDrawableChild(new ButtonWidget(x, y + 44, 200, 20, Text.literal("Deactivate ServerSpoof & Reconnect"), button -> {
+                Modules.get().get(ServerSpoof.class).toggle();
```
C10udburst, line 54: maybe just toggle no signatures to off

C10udburst (changes requested):

src/main/java/meteordevelopment/meteorclient/mixin/DisconnectedScreenMixin.java
MineGame159: No signatures has been removed so the button can also be removed.

src/main/java/meteordevelopment/meteorclient/utils/ReflectInit.java
```diff
@@ -59,6 +65,10 @@ private static <T extends Annotation> void reflectInit(Method task, Class<T> ann
             task.invoke(null);
         } catch (IllegalAccessException | InvocationTargetException e) {
             e.printStackTrace();
+        } catch (NullPointerException e) {
+            RuntimeException exception = new IllegalStateException("Method \"%s\" using Init annotations from non-static context".formatted(task.getName()));
```
MineGame159, line 69: Wouldn't the catch clause also throw another null pointer exception because it is accessing the task?

RacoonDog: NPE comes from invoking a non-static method with a null instance, not from the method itself being null

This basically just improves the error message when someone uses PreInit or PostInit but without the static keyword

MineGame159: Oh yeah, sorry missed that

PR #2715, opened 2022-09-03
improve .server plugins

C10udburst:
this time it shouldn't be preventable by servers

image


MineGame159:
Just to not what has been said in discord: from my testing the old method is better than this pr in short.

PR #2725, opened 2022-09-06
Notebot Enhancements And Fixes

EpicPlayerA10:
### Features:
- Use all instruments corresponding to the song, not only 24 random instruments (https://www.youtube.com/watch?v=WJXuNnLWF5Y)
- Note Map where you can map notes to those to be played
- Auto Play Toggle
- Auto Rotate Toggle
- There are now two Notebot Modes: `AnyInstrument`, `ExactInstrument`
- Round Out Of Range Toggle
### Fixes:
- Fixed bug where song was not playing at the start

Ps. Tested on LiveOverflow's minecraft server (if someone knows this youtuber)

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/settings/SettingGroup.java
+++ b/src/main/java/meteordevelopment/meteorclient/settings/SettingGroup.java
@@ -39,6 +39,10 @@ public class SettingGroup implements ISerializable<SettingGroup>, Iterable<Setti
         return setting;
     }
 
+    public Setting<?> getByIndex(int index) {
+        return settings.get(index);
+    }
+
     @Override
     public Iterator<Setting<?>> iterator() {
         return settings.iterator();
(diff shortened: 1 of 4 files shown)


Wide-Cat:
From the looks of it, the 'Any' mode is essentially how the old notebot worked, and the 'One To One' mode actually makes sure it uses the right instruments, requiring building something like in the linked video? If so, I would change the names of those to be more descriptive, or maybe add some additional clarification in the setting description

I-Got-You-Dead:
Can someone merge this 💀

I-Got-You-Dead (approved):

InstallHappiness:
Amazing! Should 100% be merged :D

MineGame159:
Sorry for the late response but I have a few things:
1. When I tried to use Exact mode while I didn't have all the required instruments it crashed after tuning the noteblocks.
image
2. Could you format the names in the Note Map category from HARP to Harp.
3. Could you make the missing note message a bit nicer for example: `Missing note: <note> (<block>), <level>`.

Otherwise good job.

MineGame159:
Crashes on load.

EpicPlayerA10:
I know. Fixing it rn

EpicPlayerA10:
Okay, this commit works

PR #2727, opened 2022-09-07
small optimizations and changes

RacoonDog:
-  Replaces `InvUtils.findIn*` calls with `InvUtils.testIn*` when creating a `FIndItemResult` object is not needed
-  Optimizes a few if statement chains
-  Code style + readability changes
-  Optimizes `HoleFiller` predicate
-  Caches expensive `DataTracker` calls in `AutoLog`

Before merging, please fix `AnchorAura`'s 2 duplicate lines

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/AnchorAura.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/AnchorAura.java
@@ -249,9 +249,10 @@ public class AnchorAura extends Module {
         if (PlayerUtils.shouldPause(pauseOnMine.get(), pauseOnEat.get(), pauseOnDrink.get())) return;
         if (EntityUtils.getTotalHealth(mc.player) <= minHealth.get()) return;
 
-        if (TargetUtils.isBadTarget(target, targetRange.get()))
+        if (TargetUtils.isBadTarget(target, targetRange.get())) {
             target = TargetUtils.getPlayerTarget(targetRange.get(), targetPriority.get());
-        if (TargetUtils.isBadTarget(target, targetRange.get())) return;
+            if (TargetUtils.isBadTarget(target, targetRange.get())) return;
+        }
 
         FindItemResult anchor = InvUtils.findInHotbar(Items.RESPAWN_ANCHOR);
         FindItemResult glowStone = InvUtils.findInHotbar(Items.GLOWSTONE);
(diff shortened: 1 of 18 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoLog.java
```diff
@@ -94,19 +94,20 @@ public AutoLog() {
 
     @EventHandler
     private void onTick(TickEvent.Post event) {
-        if (mc.player.getHealth() <= 0) {
+        float playerHealth = mc.player.getHealth();
+        if (playerHealth <= 0) {
```
StormyBytes, line 98: This ignores the `toggleOff` and `smartToggle` completely.

src/main/java/meteordevelopment/meteorclient/systems/modules/combat/AnchorAura.java
```diff
@@ -313,7 +314,7 @@ private BlockPos findPlacePos(BlockPos targetPlacePos) {
                 else if (isValidPlace(targetPlacePos.add(1, 1, 0))) return targetPlacePos.add(1, 1, 0);
                 else if (isValidPlace(targetPlacePos.add(-1, -1, 0))) return targetPlacePos.add(-1, -1, 0);
                 else if (isValidPlace(targetPlacePos.add(0, 1, 1))) return targetPlacePos.add(0, 1, 1);
-                else if (isValidPlace(targetPlacePos.add(0, 0, -1))) return targetPlacePos.add(0, 0, -1);
+                else if (isValidPlace(targetPlacePos.add(0, 0, -1))) return targetPlacePos.add(0, 0, -1); //Duplicate of line 313
```
StormyBytes, line 317: As for the duplicate lines, why don't you just remove them yourself? You've commented on them anyways, imo its better to just delete it instead of waiting for someone else.

RacoonDog: because i feel like those duplicate lines were supposed to actually point to a unique block and it is currently being ignored
so im waiting on someone who actually knows how this module works to fix it

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoLog.java
RacoonDog: you sure? all i did was store the result of `mc.player.getHealth()` to a variable and use that instead of calling the method multiple times, i didnt change the structure of the code

StormyBytes: Ah mb

src/main/java/meteordevelopment/meteorclient/systems/modules/combat/AnchorAura.java
StormyBytes: Since this all happens within a single tick, they won't be pointing to a different block. Probably just a forgotten left over line from duplications during writing.

RacoonDog:
gonna completely remake this shithole pr

PR #2739, opened 2022-09-08
Add LavaBucket to Offhand

ChillerDragon:
When having a full inventory of lava buckets its annoying mouse time to get those into the hotbar.
Probably similar for water but since water has infinite sources I think its more common to walk around with a inventory full of lava.
Possible use case:

https://user-images.githubusercontent.com/20344300/189163595-61aff57e-0355-4247-9fac-692b7646f42a.mp4


seasnail8169:
this is way too niche

PR #2748, opened 2022-09-12
Random stuff

CodeDoc537:
Random shit I did before sleeping


src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BookBot.java
```diff
@@ -65,8 +65,8 @@ public class BookBot extends Module {
         .name("pages")
         .description("The number of pages to write per book.")
         .defaultValue(50)
-        .range(1, 100)
-        .sliderRange(1, 100)
+        .range(1, 50)
+        .sliderRange(1, 50)
```
MineGame159, line 69: The limit is 100, why hard limit to 50?

PR #2795, opened 2022-09-27
Add/remove individual values instead of changing the whole list setting

### LINKED ISSUE #418: [SUGGESTION] Command improvements
**Describe the feature**
Currently, if you want to add a block to the search menu it isn't very viable to do it via chat. Instead, you have to open the menu, locate the module, open it, go to the blocks list, search for the block you want, add it, then close the menu. It's a long process. and it's the same for every other module where you have a list of items. (Such as xray, wallhack, or even Item Highlight.)
The only way to do it via chat is by doing:
`.settings search blocks diamond_ore`
However, this also overrides all the other blocks you had in the menu. Which is why I suggest making it:
`.settings search blocks add diamond_ore`
`.settings search blocks remove diamond_ore`
However, it's still a very lengthy command which is why I suggest adding new commands for these modules. If this is added, then the settings cmd probably doesn't need to be altered.

`.search add diamond_ore`
`.xray add diamond_ore`
`.itemhighlight remove diamond_ore`
`.stashfinder list`
`.storagesp add diamond_ore`
`.nametags add item`

**Additional info**
This feature would be way faster than the current method of editing the list.
Would be cool if it also had cmd suggestions. For example, you would type '.blockesp remove' and diamond_ore would be the next suggested word as it's already in the list.

StormyBytes:
Closes #418

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/settings/BlockListSetting.java
+++ b/src/main/java/meteordevelopment/meteorclient/settings/BlockListSetting.java
@@ -36,16 +36,18 @@ public class BlockListSetting extends Setting<List<Block>> {
     @Override
     protected List<Block> parseImpl(String str) {
         String[] values = str.split(",");
-        List<Block> blocks = new ArrayList<>(values.length);
 
         try {
             for (String value : values) {
                 Block block = parseId(Registry.BLOCK, value);
-                if (block != null && (filter == null || filter.test(block))) blocks.add(block);
+                if (block != null && (filter == null || filter.test(block))) {
+                    if (!this.value.contains(block)) this.value.add(block);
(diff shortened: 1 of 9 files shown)


MineGame159:
I don't like how you made the "toggle" behaviour be the default. It goes against how every other setting works and IMO this should be done with sub commands like the issue says.

Also I am not a big fan of not creating a new ArrayList in the parse methods because they should not modify the current state of the setting as the caller is responsible for actually setting the value or rejecting the new value. But I guess it is fine because that can never happen with those settings.

StormyBytes:
Hmm. Guess I'll rewrite it with command argument builder instead.

PR #2799, opened 2022-09-29
More settings to BaritoneTab

StormyBytes:
- String settings
- Block list settings
- Item list settings

Because I got tired of configuring lists via commands


RedCarlos26:
So based

PR #2804, opened 2022-10-02
bukkit-cmd mode for server plugins

ThebestkillerTBK:
This mode helps you get less fake plugins and bypasses permission-based plugin hiding.


Wide-Cat:
Thank you for submitting this pr. There were some issues in here which I resolved in a seperate pr, and we figured that it would be easiest to close this and open a new one. Apologies for the delay.

Issue #2816, opened 2022-10-06
Copy paste color settings [Suggestion] 


duckyz1:
### Describe the feature

Copy paste hex or rgb settings easier. 
Or add hex so we can ctrl + c   ctrl + v  color settings from one to another. 
Or add a palette that we can choose the past colors used so we don't have to copy paste it manually. 

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

I-Got-You-Dead:
already a thing
































































































cope

ThebestkillerTBK:
> cope

nah you can only copypaste rgb not hex

I-Got-You-Dead:
¯\_(ツ)_/¯

PR #2818, opened 2022-10-07
[#2816] Better Hex support

ThebestkillerTBK:


src/main/java/meteordevelopment/meteorclient/mixin/BeaconScreenMixin.java
```diff
@@ -30,7 +30,7 @@
 @Mixin(BeaconScreen.class)
 public abstract class BeaconScreenMixin extends HandledScreen<BeaconScreenHandler> {
     @Shadow
-    protected abstract <T extends ClickableWidget> void addButton(T button);
+    protected abstract <T extends ClickableWidget & BeaconScreen.BeaconButtonWidget> void addButton(T button);
```
GreemDev, line 33: This change was previously undone because it causes a crash. Despite it fixing a mixin "issue", it makes it unable to find the target method.

PR #2825, opened 2022-10-10
Rewrite left and right clicks in autoclicker

### LINKED ISSUE #1269: [Suggestion] AutoMine and AutoPlace
**Describe the feature**
Auto Mine speaks for itself, it holds left click or mines the block without you holding it. Now, this might be useful since to mine blocks with baritone faster, you have to enable Auto Mine (which meteor doesn't have yet) or hold down left click, and it will be useful with Auto Place.

Auto Place will, as simple as auto mine is, place blocks that you're holding, which can be useful for bridging (lmao hell nah) and converting ores into items.

**Screenshots**
None available

**Additional info**
None

StormyBytes:
Allows more variable use of the two (now) separate click types. 
- Closes #1269

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoClicker.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoClicker.java
@@ -16,42 +16,41 @@ import meteordevelopment.meteorclient.utils.Utils;
 import meteordevelopment.orbit.EventHandler;
 
 public class AutoClicker extends Module {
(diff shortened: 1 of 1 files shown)


Wide-Cat:
Correct me if I'm wrong, but now both left and right click will always be active? Probably want add a third mode to turn off one of the keys

seasnail8169:
and also hide the related settings when the mode is not press or hold

StormyBytes:
mb, issues should be addressed now.

PR #2834, opened 2022-10-13
Loading a bunch of fonts isn't a good idea...

ThebestkillerTBK:
Loading system font can spam error when loading specific font(Failed to load system font) and slow down client launch


ThebestkillerTBK:
> So you literally just removed a feature we deliberately added

imo just copy paste the font into the font folder

and it took a LONG time to load the system fonts

MineGame159:
Then why not for example only load the fonts once the users opens the font selector for the first time? (ofc still load the currently selected one) I don't see a reason why the feature should be removed.

Issue #2693, opened 2022-08-26
[Suggestion] 4 Blink enchantments 


Affexe:
### Describe the feature

Add the cancel keybind that will prevent packets from sending to a server (#2104 closed bc author died, wasn't answered), would be really useful on 2b2t spawn where you have to carefully perform hard parkour jumps (can fall because of low speed or wrong angle even with parkour module)

The "anti bow-aimbot", that will cut off the time (suspend packets) if you gonna take damage from a projectile soon.

Also there could be auto blink restarter to not to get lagged back

As well as "second chance mode" which will work with blink canceler that was mentioned above and perform actions in real time but with some delay, for instance to make baritone's nether adventures much reliable or to prevent your oncoming mistakes using this wonderful time roll back. (These "several seconds of the future" (delay) will be only on client side until the time comes)

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

RickyTheRacc:
bruh

Affexe:
> bruh

Ain't it cool😭

Mathrandom7910:
no ;(

rfl890:
no ;)

PR #2838, opened 2022-10-14
Add a cancel keybind for blink

StormyBytes:
Idea from #2693


MineGame159:
Wouldn't it make sense if it also turned off the module? Maybe a setting that only shows up if the keybind is not none. Or maybe I am just saying bullshit.

StormyBytes:
> Wouldn't it make sense if it also turned off the module? Maybe a setting that only shows up if the keybind is not none. Or maybe I am just saying bullshit.

Whoops mb. Now it should toggle the module and reset packets & player position. As for a setting if the keybind is not none, well the point of "cancelling blink" is to not send the packets and reset your position. Doing only one and not the other as well would make it pointless.

PR #2843, opened 2022-10-15
Silent disconnect & Reconnect command

ThebestkillerTBK:


src/main/java/meteordevelopment/meteorclient/systems/modules/misc/PacketCanceller.java
```diff
@@ -48,4 +56,8 @@ private void onReceivePacket(PacketEvent.Receive event) {
     private void onSendPacket(PacketEvent.Send event) {
         if (c2sPackets.get().contains(event.packet.getClass())) event.cancel();
     }
+
+    public boolean silentDisconnect() {
+        return silentDisconnect.get();
```
GreemDev, line 61: wouldn't it be better to put an `isActive()` check in here, considering the code accessing this does not?

MineGame159:
I don't know. I don't see any point in both of those features. If you or some other peope have some use cases write them.

PR #2848, opened 2022-10-15
Added NoAutoJump to Strafe

vaxerski:
I am used to how the "strafe" module works in other clients, just allowing you to yeet physics out of the window and have no momentum, without having to jump all the time.

This adds a checkbox to the Speed module in Strafe mode to disable the autojump.


seasnail8169:
just use the default mode if you want to zoom around without momentum

seasnail8169:
strafe is designed to go a bit faster than sprint jumping without flagging ncp afaik

PR #2849, opened 2022-10-15
Added AutoDock to ElytraFly

vaxerski:
AutoDock was a very useful feature in Harakiri's ElytraFly module: it will automatically hover you .34 blocks off ground when you hold shift, allowing you to fly through 1x2 tunnels on certain servers (e.g. 9b9t)

https://user-images.githubusercontent.com/43317083/196003683-4df02178-7b1f-4b31-8fbb-2d9f46565e09.mp4

It's slowly docking (and setting pitch to 0) for a reason - to not "pop" you off the elytra and allow you to stay airborne.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/AbstractBlockMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/AbstractBlockMixin.java
@@ -15,6 +15,7 @@ import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.shape.VoxelShape;
 import net.minecraft.world.BlockView;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/ElytraFly.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/ElytraFly.java
@@ -334,13 +334,16 @@ public class ElytraFly extends Module {
             final boolean underCollidable = under.collidable || !underState.getFluidState().isEmpty();
             final boolean under2Collidable = under2.collidable || !under2State.getFluidState().isEmpty();
 
-            if(!underCollidable && under2Collidable) {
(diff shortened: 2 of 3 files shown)


MineGame159:
1. If I understood it correctly then it would be better to replace the air / water / lava checks with a check if the block has a collision block or not. So it can work properly with blocks.
2. Instead of forcing the pitch to be 0 you can clamp it between some range as the client doesn't land immediately. Don't know what the values are so you will just have to try and see what numbers work or find it in the code.
3. Move the settings out of the inventory group.

vaxerski:
1. makes sense
2. iirc anything below like 10 lands, but yeah clamping makes sense.
3. oops sorry

Mostly copied what was already made by me ages ago in harakiri - but yeah your points are completely valid. Will adjust tomorrow.

MineGame159:
Plus idk if docking is the best word here, maybe auto hover?

StormyBytes:
Docking to me sounds more like a "safe landing" type thing like NoFall

vaxerski:
All points addressed.

Renamed to autohover

Clamping between -90 and 20. >= 30 makes you pop off the elytra, but with lag that can be greater. It's impossible to predict the exact amount though, the server can be running 20TPS with you @ 5ms ping, or at 10TPS with you @ 200ms.

Widened AbstractBlock's collidable member and now using that.

Moved to general.

MineGame159:
Also before I look at the changes I will say this. Please when making changes to PRs just push separate commits, it's better for readability in the PR and when merging I can easily squash them if I want. 😅

vaxerski:
Sure, sorry, I come from projects where we squash everything into one commit.

MineGame159:
1. Forgot to rename the field for the setting.
2. Apparently fluids are not collidable so you will have to check for those separately with just a general "is fluid" check.
3. You can use our `Utils.clamp()` method instead of min + max.

Hopefully thats all.

vaxerski:
> Apparently fluids are not collidable so you will have to check for those separately with just a general "is fluid" check.

Yes, but we also have a "off in water" setting. Should we respect that?

MineGame159:
Hmmmm, one one side it makes sense to respect the setting but on the other you probably don't want to go into water if you are using the hover feature.

MineGame159:
I think it shouldn't respect it.

vaxerski:
done.

MineGame159:
Also just noticed that if your pitch is higher than some threshold (~ >60) then it fails to clamp the pitch and thus lands on the ground.

vaxerski:
fixed, also now clamping to 50 (because >50 will pop you off to, I noticed)

PR #2850, opened 2022-10-15
Added a NoGlitchBlocks Module

### LINKED ISSUE #818: [Suggestion] Anti Ghost Blocks
**Describe the feature**
Anti ghost block.

vaxerski:
Fixes #818, IMHO.

This adds a module that will prevent ghost blocks from breaking stuff too fast. 

I've tested the recommended "Anti Ghost" mod on 9b9t (multiconnect 1.12.2) and it doesn't do anything. This, however, works perfectly.

It will prevent the client from removing the block until the server acknowledges the break. It will still play the breaking animations and sounds, because being silent it is annoying and confusing.

https://user-images.githubusercontent.com/43317083/196008001-fe04d9a0-3ed5-4997-a73b-c05343598daf.mp4

I am unsure about the name but that's how it's been called in my client and I took the name from future, so it should be pretty fitting.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/events/entity/player/BreakBlockEvent.java
+++ b/src/main/java/meteordevelopment/meteorclient/events/entity/player/BreakBlockEvent.java
@@ -5,11 +5,12 @@
 
 package meteordevelopment.meteorclient.events.entity.player;
 
+import meteordevelopment.meteorclient.events.Cancellable;
 import net.minecraft.block.BlockState;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 
-public class BreakBlockEvent {
+public class BreakBlockEvent extends Cancellable {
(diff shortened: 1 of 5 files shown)


MineGame159:
Instead of adding the cancelled field manually make the event extend `Cancellable` class and call `INSTANCE.setCancelled(false);` inside the `BreakBlockEvent.get()` method and then `.cancel()` in the module.

vaxerski:
done because I am still awake for some reason and this was an easy fix

StormyBytes:
How about renaming it to "NoGhostBlocks"? Imo seems more fitting for the description "Attempts to prevent ghost blocks arising..."

vaxerski:
> I am unsure about the name but that's how it's been called in my client and I took the name from future, so it should be pretty fitting.

I will leave any naming decisions to y'all, I really don't mind as long as it actually is intuitive.

PR #2857, opened 2022-10-16
optimize HoleFiller predicate

RacoonDog:
removes duplicate instanceof check


RacoonDog:
grrr github acting like an average meteor client user

PR #2860, opened 2022-10-16
speed mine: prevent ghost blocks spawning

### LINKED ISSUE #1650: [Suggestion] New option for speed mine

ThebestkillerTBK:
closes #1650


MineGame159:
Wouldn't it be better if this feature was like a "speed limiter" setting? I can imagine that it might be useful to still go over this safe limit on some servers.

src/main/java/meteordevelopment/meteorclient/mixin/PlayerEntityMixin.java
```diff
@@ -48,8 +54,23 @@ private void onDropItem(ItemStack stack, boolean bl, boolean bl2, CallbackInfoRe
...
         if (!module.isActive() || module.mode.get() != SpeedMine.Mode.Normal) return;
+        float breakSpeed = cir.getReturnValue();
+        float breakSpeedMod = (float) (breakSpeed * module.modifier.get());
+
+        if (module.modifier.get() < 1) {
+            cir.setReturnValue(breakSpeedMod);
+            return;
+        }
+
+        HitResult result = mc.player.raycast(mc.interactionManager.getReachDistance(), 1f / 20f, false);
```
MineGame159, line 65: Also you can use `mc.crosshairTarget` here instead of making a raycast.

PR #2865, opened 2022-10-16
NoBreakDelay -> BreakDelay, now you can modify the break cooldown

ThebestkillerTBK:


seasnail8169:
wouldn't there be a conflict when break delay and nuker both set that field

PR #2874, opened 2022-10-18
Flight Packet AntiKick Improvements and Quick Notebot Fixes

EpicPlayerA10:
### Flight Changes
- Added 'Fly Down Delay' option which controls a delay to fly down a bit
- Changed the default 'Fly Down Delay' to 1500ms to avoid kicking while player is not moving
- Added better check if player is in the air to avoid kicking from the server
### Quick Notebot Fixes
- Don't autoplay when song preview has been ended
- Don't play song when there aren't any valid noteblock nearby


seasnail8169:
please separate these into different pull requests, unrelated changes should always be in separate prs

seasnail8169:
also the new setting added in flight shouldn't exist, just change the off-time setting to work for both modes

PR #2879, opened 2022-10-19
Use levenshtein distance for module searcher

StormyBytes:
Works better with typos and small inaccuracies. 
Perhaps the ins/del/sub costs may need to be tweaked, idk 1 3 5 worked good enough for me.


MineGame159:
Wouldn't it be better to replace the current `Utils.search()` with some default costs?

StormyBytes:
Should I change `Utils.search()` to utilize levenshtein and have it use some default costs? Kinda confused as to what you're saying.

MineGame159:
Well thats what I am asking if it wouldn't be better. Isn't this supposed to be a generally better search algorithm?

StormyBytes:
> Well thats what I am asking if it wouldn't be better. Isn't this supposed to be a generally better search algorithm?

Alright, levenshtein distance is now applied to other text search usages.

PR #2895, opened 2022-10-22
Quality of life modules

yorik100:
NoJumpDelay
KnockbackPlus
AutoClicker use weapon delay option
AutoClicker only when clicking option


I-Got-You-Dead:
is there not already multiple addons with nojumpcooldown?

Wide-Cat:
Could you open another pr with just the NoJumoCooldown module? The other changes would better fit to be an addon.

PR #2906, opened 2022-10-27
Refactor ticking block render into RenderUtils

StormyBytes:
Idea originated from #donator-chat 
https://discord.com/channels/689197705683140636/713429344135020554/1035102740982018069


Wide-Cat:
might want to take a look at nuker as well because i think it uses a similar rendering mode

StormyBytes:
> might want to take a look at nuker as well because i think it uses a similar rendering mode

done

MineGame159:
Have you even tried running it? Lol. All of the methods you added need to be static.

MineGame159:
For the love of god please fucking test your changes.

PR #2907, opened 2022-10-27
Have !swapBack set prevSlot to swap slot

StormyBytes:


MineGame159:
Shouldn't it set it to -1?

StormyBytes:
actually yeah

Issue #2917, opened 2022-10-30
Notebot breaks note blocks when switching to creative while playing a song


niclan051:
### Describe the bug

Notebot breaks note blocks when switching to creative while playing a song

### Steps to reproduce

```bash
1. Start playing a song using Notebot
2. Switch to creative mode
```

### Meteor Version

0.5.1

### Minecraft Version

1.19.2

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported (I have searched bug reports on GitHub).

- [x] This is a valid bug (I am able to reproduce this on the latest dev build).

C10udburst:
Use `.gamemode survival` <img src="https://cdn.discordapp.com/emojis/888951267873923093.png?size=48" width="24" alt=":verycool:">

PR #2918, opened 2022-10-30
fixed https://github.com/MeteorDevelopment/meteor-client/issues/2917

niclan051:


niclan051:
why did merge commits appear here

PR #2934, opened 2022-11-06
adding estimated damage to combathud

circularsprojects:
this shows the estimated damage dealt to the target based on held item and the target's armor


circularsprojects:
??

PR #2941, opened 2022-11-09
Added distance colors to esp

### LINKED ISSUE #2938: Distance colors ESP
### Describe the feature

Meteor client developers, I have a suggestion that could add an overall a better user experience with the client and its customizability. I thought it would be great to add Distance colors to ESP because this can add a more appealing look to the tracer + ESP combination due to only being able to enable Distance colors for tracers which would result in an inconsistent color scheme, which some users might find unappealing, including me. I have been using meteor client for almost a year and a half now and donated, and after using it, this is the only thing I've wanted added to the client. Thank you for taking to the time to read this and have a good day.  (I've seen other utility clients with distance colors for esp as well.)

I-Got-You-Dead:
closes #2938


MineGame159:
Doesn't work with fade distance.
Also I don't know why you implemented it so complexly. All you have to do is just modify `getEntityTypeColor(Entity)` to add `if (distance.get()) return getColorFromDistance(entity);` at the top of the method. That way you don't have to change the calls to `getColor(Entity)` or mess with the rest of `getEntityTypeColor(Entity)`. Well unless I am missing something.

I-Got-You-Dead:
Cuz I’m a smooth brain

I-Got-You-Dead:
Should be fixed now

PR #2942, opened 2022-11-10
Document some methods in Notebot

EpicPlayerA10:
### Changes
- Documentate some methods
- Make `Stage` class public
- Make `loadFileToMap` method public
- Rename `stop()` to `onSongEnd()`
- Rename `forceStop()` to `stop()`

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Notebot.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Notebot.java
@@ -518,7 +518,7 @@ public class Notebot extends Module {
     }
 
     /**
-     * Set up a tune hits map which tells how many times player need to
+     * Set up a tune hits map which tells how many times player needs to
      * hit noteblock to obtain desired note level
      */
     private void setupTuneHitsMap() {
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Notebot.java
```diff
@@ -447,6 +447,9 @@ else if (mc.player.getAbilities().creativeMode) {
         }
     }
 
+    /**
+     * Setups a map of noteblocks positions
```
RacoonDog, line 451: `Sets up` instead of `Setups`

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Notebot.java
```diff
@@ -513,6 +516,10 @@ private void setupNoteblocksMap() {
         }
     }
 
+    /**
```
RacoonDog, line 519: `Sets up`*

RacoonDog (changes requested):
Also `document` instead of `documentate`

PR #2945, opened 2022-11-12
allow registering setting widget factories

RacoonDog:


MineGame159:
I think it would be better if the custom factories field was moved to the `SettingsWidgetFactory` base class and then instead of it being copied into the normal factories map in the constructor it would inside `group()` first check if there is a custom factory for it and use that, if not use the normal one.

Also it might be a good idea to make `SettingsWidgetFactory` an abstract class instead of an interface and add a protected method like `getFactory(Setting)` which would do that automatically.

src/main/java/meteordevelopment/meteorclient/gui/utils/SettingsWidgetFactory.java
```diff
@@ -5,10 +5,49 @@
...
+    /** {@code SettingsWidgetFactory.registerCustomFactory(SomeSetting.class, (theme) -> (table, setting) -> {//create widget})} */
+    public static void registerCustomFactory(Class<?> settingClass, Function<GuiTheme, Factory> factoryFunction) {
+        customFactories.put(settingClass, factoryFunction);
+    }
+
+    public static void unregisterCustomFactory(Class<?> settingClass) {
+        customFactories.remove(settingClass);
+    }
+
+    protected abstract void registerDefaultFactories();
```
MineGame159, line 40: Why is this an abstract method? Just register them in the constructor of DefaultSettingsWidgetFactory

RacoonDog: i originally added it so that it was possible to extend `DefaultSettingsWidgetFactory` while also modifying the factories themselves, but then i moved all the non-factory code out of it and forgot to remove that lol

PR #2955, opened 2022-11-13
Search module limit

### LINKED ISSUE #1780: [Suggestion]
### Describe the feature

A culling feature for the module:search it would be nice because u can lag if u are trying to find something which is kinda alot in the area and culling would be very nice to have in that case

ThebestkillerTBK:
no more lags lol (closes #1780)

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/events/world/UnloadChunkEvent.java
+++ b/src/main/java/meteordevelopment/meteorclient/events/world/UnloadChunkEvent.java
@@ -1,0 +1,20 @@
+/*
+ * This file is part of the Meteor Client distribution (https://github.com/MeteorDevelopment/meteor-client).
+ * Copyright (c) Meteor Development.
+ */
+
+package meteordevelopment.meteorclient.events.world;
+
+import meteordevelopment.meteorclient.events.Cancellable;
+import net.minecraft.world.chunk.WorldChunk;
+
+public class UnloadChunkEvent extends Cancellable {
+    private static final UnloadChunkEvent INSTANCE = new UnloadChunkEvent();
+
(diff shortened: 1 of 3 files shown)


RacoonDog:
w pr

I-Got-You-Dead:
> w pr

fr

MineGame159:
There is a very big problem with this. In the screenshot I am searching for diamonds and set my limit to 1000 (or even the default 10000 does the trick) and you can see it searched the chunks from furthest to closest which it should be the other way around. Also when I move around it should rescan the closest chunks because right now even if I move those 1000 blocks will stay there until they are unloaded and no new diamonds will be shown.
image

I think a better way to approach this is to search every chunk as it does right now and limit the amount of chunks it renders.

ThebestkillerTBK:
> There is a very big problem with this. In the screenshot I am searching for diamonds and set my limit to 1000 (or even the default 10000 does the trick) and you can see it searched the chunks from furthest to closest which it should be the other way around. Also when I move around it should rescan the closest chunks because right now even if I move those 1000 blocks will stay there until they are unloaded and no new diamonds will be shown.
> I think a better way to approach this is to search every chunk as it does right now and limit the amount of chunks it renders.

It seems hard to render the closest chunks and if you check the limit on render it will cause lag

MineGame159:
Why would it be hard or slow? In the render method before it loops over the chunks you just need to sort all of the chunks and then have a loop with an if statement to break out of the loop if the number of blocks rendered increased over the limit. And for the lag part, sorting the chunks is just a matter of calculating the distance to the camera, and since you are only comparing the values you can use the squared version to avoid square root operations.

I-Got-You-Dead:
Sadage

PR #2971, opened 2022-11-17
Rename Search to Block ESP

RacoonDog:
There are many reasons why I feel this is a necessary change.

1. The name `Search` is ambiguous. 
When someone asks:

> Which module has ESP for blocks?  

And is replied with "Search", that person might think that we are asking them to search it up, not knowing that it is, in fact, the name of the module in question.
This change will hopefully lower the amount of people asking about `Block ESP` on the Discord.

2. The name `Search` is already used in the client, as in the modules tab search bar.

3. The name `Block ESP` is much more indicative of this module's use.
The term `ESP` is already used multiple time in the client (`ESP`, `Hole ESP`, `City ESP`, etc.) and is used to designate the exact same feature as the `Search` module. Not only that, but `Search` doesn't specify the target to be searched, whilst `Block ESP` clearly indicates the module's intended use just by the name.


I-Got-You-Dead:
Agreed

ehylo:
Agreed

RacoonDog:
gangshit 😈

PR #2975, opened 2022-11-18
add color util methods

RacoonDog:
I personally needed the `SettingColor(int packed)` constructor and the `Color.toTextColor()` method for addon reasons and while I was at it I figured why not add the rest

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/render/color/Color.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/render/color/Color.java
@@ -8,7 +8,11 @@ package meteordevelopment.meteorclient.utils.render.color;
 import meteordevelopment.meteorclient.utils.misc.ICopyable;
 import meteordevelopment.meteorclient.utils.misc.ISerializable;
 import net.minecraft.nbt.NbtCompound;
+import net.minecraft.text.Style;
+import net.minecraft.text.TextColor;
+import net.minecraft.util.Formatting;
 import net.minecraft.util.math.Vec3d;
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/gui/screens/settings/ColorSettingScreen.java
```diff
@@ -58,14 +58,12 @@ public boolean fromClipboard() {
...
 
         if ((parsed = parseRGBA(clipboard)) != null) {
-            setting.get().rainbow(false);
             setting.set(parsed);
             setting.get().validate();
             return true;
         }
 
         if ((parsed = parseHex(clipboard)) != null) {
-            setting.get().rainbow(false);
```
MineGame159, line 68: This was done on purpose, when you are pasting a color into the color picker you don't want to have rainbow enable from before, that doesn't make sense as the pasted color is never gonna be used anyways.

RacoonDog: the `Setting`’s value isn’t mutable, running `setting.set(parsed)` completely overwrites the previous value, so setting a value prior to overwriting it is useless

I think you’re confusing
`setting.set(parsed)` with `setting.get().set(parsed)` in which case that line would be required as it doesn’t take the rainbow parameter into account

MineGame159: Oh yeah, sorry my bad

PR #2984, opened 2022-11-21
RideStats module and add a filter to EntityTypeListSetting

Declipsonator:
Ridestats is a module that I made a while back that shows info over rideable mobs' heads (like speed, jump height, and slots for llamas). I changed the onlyAttackable variable in EntityTypeListSetting to a filter for it (which was marked by a todo anyways). Only merge this if it's a good feature, I understand if it's not. \
\

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/screens/settings/EntityTypeListSettingScreen.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/screens/settings/EntityTypeListSettingScreen.java
@@ -16,7 +16,6 @@ import meteordevelopment.meteorclient.gui.widgets.input.WTextBox;
 import meteordevelopment.meteorclient.gui.widgets.pressable.WCheckbox;
 import meteordevelopment.meteorclient.settings.EntityTypeListSetting;
 import meteordevelopment.meteorclient.utils.Utils;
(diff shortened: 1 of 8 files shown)


RacoonDog:
Good idea for a module, but imo should be merged with `Entity Owner` into some kind of `Entity Nametags` module

seasnale:
also i think these stats should, by default, only render on the crosshair target entity to avoid clutter

I-Got-You-Dead:
^

Declipsonator:
I could see merging this with entity owner, and at that point just merging that into nametags. I'll leave this open for another day in case anyone else has any comments. The only thing I'm not sure about is @RacoonDog's idea of showing stats for every entity which seems like a large undertaking.

RacoonDog:
> I could see merging this with entity owner, and at that point just merging that into nametags. I'll leave this open for another day in case anyone else has any comments. The only thing I'm not sure about is @RacoonDog's idea of showing stats for every entity which seems like a large undertaking.

Would it be though? Isnt most of the interesting stuff located in the `LivingEntity` class? If so then we dont need to hardcode every mob’s stats, just a few outliers like the horse entity

Declipsonator:
I think a lot of entities have their own special traits in their own class but I see what you're saying. I'll close this for now.

PR #2986, opened 2022-11-22
Added ignoreFriends to tracers

### LINKED ISSUE #1255: [Suggestion] Option to disable tracers for friends
Would be very appreciated
Because chasing someone with friends is complicated and it could be way easier

thebest9178:
Closes #1255
Moved settings in sgAppearance to their own section like other modules do


I-Got-You-Dead:
I’ll work on adding friends to distance colors tomorrow so if you have that planned, cope

PR #2988, opened 2022-11-22
add addons to crash report

RacoonDog:


MineGame159:
Is this really needed? The crash report already contains all mods.

RacoonDog:
> Is this really needed? The crash report already contains all mods.

yeah but it also contains all 50+ fapi modules
good to have imo

src/main/java/meteordevelopment/meteorclient/mixin/CrashReportMixin.java
```diff
@@ -30,6 +32,31 @@ private void onAddStackTrace(StringBuilder sb, CallbackInfo info) {
             sb.append("Dev Build: ").append(MeteorClient.DEV_BUILD).append("\n");
         }
 
+        if (!AddonManager.ADDONS.isEmpty()) {
+            sb.append("\n[[ Addons ]]\n");
+            for (var addon : AddonManager.ADDONS) {
+                sb.append(addon.name);
+
+                List<String> authors = new ArrayList<>(List.of(addon.authors));
+                authors.removeIf(s -> s.equals("seasnail") || s.equals("seasnail8169"));
```
C10udburst, line 41: :trollface:

RacoonDog: id rather die standing than live kneeling

Wide-Cat: o7

C10udburst: o7

PR #2991, opened 2022-11-24
Add Back CustomFOV

lapoliciarobomiquesofrances:
Uses some code from https://github.com/seasnale/custom-fov
(Rule 1 is: always give seasnail credit where credit is due)


seasnale:
its already a command, you don't need the module

but the mixin can stay

PR #2993, opened 2022-11-24
Nuker blacklist

### LINKED ISSUE #2256: [Suggestion]
### Describe the feature

Maybe make the nuker have a block blacklist not only whitelist so you can blacklist blocks you would like to not mine

thebest9178:
Closes #2256
Added blacklist in nuker
Minor formatting


thebest9178:
A problem occurred configuring root project 'meteor-client'.
> Could not resolve all dependencies for configuration ':modCompileOnly'.
   > Could not resolve io.vram:canvas-fabric-mc119:1.0.+.

I don't think that's the fault of this pr

MineGame159:
Why did you remove the bounding box and broken blocks comments?

thebest9178:
I don't see it as being useful just clutter and a potential cause for confusion when someone assumes it's a separate category because no other module does that

thebest9178:
Want me to put it back?

PR #2996, opened 2022-11-25
EntityControl REVAMPED

ThebestkillerTBK:
BoatFly, EntitySpeed -> EntityControl
ReverseStep support vehicles


seasnale:
can you remove your formatting changes to Modules

MineGame159:
1. Can you put speed and fly into separate groups as they expand into a bunch of other settings and it makes it a bit confusing to look at.
2. Speed doesn't seem to work for pigs, actually it only slows them down no matter the speed I set in the module.
3. Saddle spoof works even if the entity is not selected.
4. Would it be possible to make it so when fly is enabled and you press space to go up with a horse to cancel the normal jump method?
5. The speed you set in the module doesn't seem to match what the HUD is showing (which we know is correct based on other modules).

Otherwise looks alright, good job.

ThebestkillerTBK:
> 2. Speed doesn't seem to work for pigs, actually it only slows them down no matter the speed I set in the module.

use carrot stick

MineGame159:
> use carrot stick
I am using it but it just slows the pig down no matter what speed value I use.

Also did you find something about the speed being different than what HUD shows?

ThebestkillerTBK:
> Also did you find something about the speed being different than what HUD shows?

the entity itself is moving smh

seasnale:
why close?

Issue #2963, opened 2022-11-14
AutoTool avoid sword option


moritz-t-w:
### Describe the feature

I don't use AutoTool because it uses up my sword, would be nice if there was an option to exclude any given tool(s).

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

I-Got-You-Dead:
“Am I a joke to you?” - Mending

moritz-t-w:
🤯

MineGame159:
Added in cd739b56188b61f08cf2e3398b2457e4664b7b52.

PR #2998, opened 2022-11-27
added blacklist to auto tool

TaxEvasiqn:
description

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
@@ -23,15 +23,21 @@ import net.minecraft.block.BlockState;
 import net.minecraft.block.Blocks;
 import net.minecraft.enchantment.EnchantmentHelper;
 import net.minecraft.enchantment.Enchantments;
-import net.minecraft.item.ItemStack;
-import net.minecraft.item.ShearsItem;
-import net.minecraft.item.SwordItem;
-import net.minecraft.item.ToolItem;
+import net.minecraft.item.*;
 
+import java.util.List;
 import java.util.function.Predicate;
 
 public class AutoTool extends Module {
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
+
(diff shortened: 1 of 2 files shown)


TaxEvasiqn:
fixes https://github.com/MeteorDevelopment/meteor-client/issues/2963 or something i guess

src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
```diff
@@ -173,12 +180,11 @@ public static double getScore(ItemStack itemStack, BlockState state, boolean sil
...
             score += 9000 + (item.getMaterial().getMiningLevel() * 1000);
 
-
         return score;
     }
 
-    public static boolean isTool(ItemStack itemStack) {
-        return itemStack.getItem() instanceof ToolItem || itemStack.getItem() instanceof ShearsItem;
+    public static boolean isTool(Item item) {
+        return item instanceof ToolItem || item instanceof ShearsItem;
```
MineGame159, line 187: Why did you change it to take in `Item`? From my perspective a useless change that only breaks compatibility with third party code.

src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
```diff
@@ -123,6 +129,7 @@ private void onStartBreakingBlock(StartBreakingBlockEvent event) {
 
         for (int i = 0; i < 9; i++) {
             double score = getScore(mc.player.getInventory().getStack(i), blockState, silkTouchForEnderChest.get(), prefer.get(), itemStack -> !shouldStopUsing(itemStack));
+            if (blacklist.get().contains(mc.player.getInventory().getStack(i).getItem())) score = -1;
```
MineGame159, line 132: Can you move this before the `getScore()` call and return early so it doesn't need to run if it is already blacklisted. And while you are at it put the item inside a variable just so it is easier to read than twice repeating `mc.player.getInventory().getStack(i)`.

MineGame159:
Closed in favor of cd739b56188b61f08cf2e3398b2457e4664b7b52.

PR #2999, opened 2022-11-27
Auto Login

ThebestkillerTBK:


seasnale:
we are never gonna add stuff that is specifically for cracked servers
imo cracked shouldn't be in the account manager but whatever, this definitely isn't something meteor is ever gonna have

PR #3004, opened 2022-11-29
addons can have no authors

RacoonDog:
[as per the docs](https://fabricmc.net/wiki/documentation:fabric_mod_json_spec#:~:text=0.x.-,Mandatory%20fields) the `authors` field is optional, which would make the title credits fucky
this makes it unfucky

(the `name` field is also optional, but it already defaults to `id` if not present)


MineGame159:
Honestly I have to disagree with fabric's decision on this, in my opinion it would be better if meteor itself enforced that addons have the authors field present and crash with a nice message explaining what happened.

@seasnale your opinion?

seasnale:
yeah i agree

PR #3011, opened 2022-12-01
better baritone descriptions

Cr7mbl3:
baritone has way more settings than I thought...

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/tabs/builtin/BaritoneTab.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/tabs/builtin/BaritoneTab.java
@@ -205,6 +205,7 @@ public class BaritoneTab extends Tab {
         descriptions = new HashMap<>();
         addDescription("acceptableThrowawayItems", "Blocks that Baritone is allowed to place (as throwaway, for sneak bridging, pillaring, etc.)");
         addDescription("allowBreak", "Allow Baritone to break blocks");
+        addDescription("allowBreakAnyway", "Blocks that baritone will be allowed to break even with allowBreak set to false");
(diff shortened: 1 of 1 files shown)


ehylo:
You are missing the descriptions for:
`allowBreakAnyway` - Blocks that baritone will be allowed to break even with allowBreak set to false
`randomLooking113` - How many degrees to randomize the yaw every tick. Set to 0 to disable
`mobSpawnerAvoidanceRadius` - Distance to avoid mob spawners.
`mobAvoidanceRadius` - Distance to avoid mobs.
`doBedWaypoints` - allows baritone to save bed waypoints when interacting with beds
`doDeathWaypoints` - allows baritone to save death waypoints
`verboseCommandExceptions` - Print out ALL command exceptions as a stack trace to stdout, even simple syntax errors

Some other things:
`containerMemory` does not exist (anymore)
`minYLevelWhileMining` description is wrong, and it should be "Sets the minimum y level whilst mining - set to 0 to turn off. if world has negative y values, subtract the min world height to get the value to put here"
`buildValidSubstitutes` description is also missing some stuff, but it's just syntax, so it's not as important. 

These are all of the ones I could find, but I'm sure I missed some. I would suggest looking at [Meteor's baritone settings](https://github.com/MeteorDevelopment/baritone/blob/1.19/1.19.2/src/api/java/baritone/api/Settings.java) to get the actual list, as the settings list on [Leijurv's website](https://baritone.leijurv.com/baritone/api/Settings.html) is for 1.2.15, not 1.9/Meteor's baritone

Cr7mbl3:
thank you @ehylo for pointing this out, last commit contains descriptions for all settings

PR #3012, opened 2022-12-02
Friend color override

I-Got-You-Dead:
Added the option to override distance colors with friend colors.


MineGame159:
I think the setting should work even if distance colors is disabled. That would require `PlayerUtils.getPlayerColor()` to take in a parameter as to if it should use friend color but make sure to add an overload that defaults that parameter to true.

PR #3013, opened 2022-12-02
Auto Tool exclude swords

I-Got-You-Dead:
Added the option to exclude swords from Auto Tool.


MineGame159:
I would rather have a general whitelist / blacklist setting instead of a specific one for swords. There was an attempt in #2998 but looks like the author does not want to update it.

MineGame159:
Closed in favor of cd739b56188b61f08cf2e3398b2457e4664b7b52.

PR #3022, opened 2022-12-06
BetterTooltips

ThebestkillerTBK:
Modifiable hidden elements
Rename entities -> buckets to prevent confusion

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BetterTooltips.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/BetterTooltips.java
@@ -136,8 +136,8 @@ public class BetterTooltips extends Module {
         .build()
     );
 
-    private final Setting<Boolean> entities = sgPreviews.add(new BoolSetting.Builder()
-        .name("buckets")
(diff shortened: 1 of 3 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/render/BetterTooltips.java
```diff
@@ -171,6 +165,57 @@ public class BetterTooltips extends Module {
...
+
+    private final Setting<Boolean> canPlaceOn = sgHideFlags.add(new BoolSetting.Builder()
+        .name("can-place-on")
+        .description("Show \"CanPlaceOn\" tag.")
+        .defaultValue(false)
+        .build()
+    );
+
+    private final Setting<Boolean> etc = sgHideFlags.add(new BoolSetting.Builder()
+        .name("etc")
```
I-Got-You-Dead, line 206: Can I suggest changing the setting name to misc instead of etc.

MineGame159:
To even better prevent confusion it should be named with multiple words like `entities-in-buckets`. And for the sections I think the default values should match vanilla.

PR #3023, opened 2022-12-07
About Compatibility

XC071224:
Hello, I want to use both the Meteor client and the Wurst client in my world version 1.19, but adding the Mods folder to start the game at the same time will cause an error. Can you solve it?


RacoonDog:
they are incompatible

PR #3026, opened 2022-12-07
Updated method of NoRender entities. 

chiteroman:
Drops spawn packets of selected entities.


Wide-Cat:
If you were to disable no render after blocking some entities, how long does it take them to reappear?

chiteroman:
> If you were to disable no render after blocking some entities, how long does it take them to reappear?

That is a problem, the purpose of this change was to avoid FPS drops when there are a lot of entities, for example, near lag machines. The only way to see it again would be to disable the module and reconnect to the server.

Wide-Cat:
In that case, it might be best to add another setting to control whether you want to use the old system or new one, to potentially prevent you from needing to reconnect if you want to disable it

chiteroman:
Updated it!! I added an option to enable or disable that.

I-Got-You-Dead:
Why do people add things that give an advantage against other clients in free clients where it just gets skidded into every client

chiteroman:
> Why do people add things that give an advantage against other clients in free clients where it just gets skidded into every client

Lambda client also has it. It's very useful for building lag machines that have a lot of entities like the last one of The Horizon.

I-Got-You-Dead:
> > Why do people add things that give an advantage against other clients in free clients where it just gets skidded into every client
> 
> Lambda client also has it. It's very useful for building lag machines that have a lot of entities like the last one of The Horizon.

Yea exactly, don’t give it to absolutely everyone

chiteroman:
> Let’s say you want to crash people’s clients on a server, you want to crash them and not yourself. If they have this they won’t crash. With the amount of downloads meteor has, half the script kiddies you want to crash just won’t.

XDXDXD. If you don't want to merge it it's ok, I will make an addon.

Wide-Cat:
😐 don't close this just because he told you, wait to see what snail or minegame says

I-Got-You-Dead:
Yea don’t

I-Got-You-Dead:
I’m just a bit annoyed that its public in a popular client

MineGame159:
> I’m just a bit annoyed that its public in a popular client

@I-Got-You-Dead So we should not add features just because a paid client has them? The whole point of Meteor is that you don't need to spend money on a damn block game client.

I-Got-You-Dead:
> > I’m just a bit annoyed that its public in a popular client
> 
> @I-Got-You-Dead So we should not add features just because a paid client has them? The whole point of Meteor is that you don't need to spend money on a damn block game client.
> 
> Also with the way you say "gets skidded into every single other client" you make it sound like if the feature takes longer than 5 minutes to code for someone who has atleast some idea of what they are doing.

That isn’t what I said but ok

I-Got-You-Dead:
What I’m saying is that by adding it to meteor, it’s giving script kiddies the power to avoid being crashed

MineGame159:
So let me rephrase, we should not improve Meteor anymore because it got popular?

I-Got-You-Dead:
You know what yes, you shouldn’t

I-Got-You-Dead:
:trollface:

PR #3036, opened 2022-12-10
Allow  Rotation Priority w/o Callbacks

RickyTheRacc:
No more self referencing when you don't want to define a callback but need a rotation to be done first


MineGame159:
No more what now? You just pass in null as you did in the commit.

PR #3048, opened 2022-12-12
Highlight currently active .modules

DesiCow:
Highlights currently active modules with yellow color in chat on usage of `.modules` command


seasnale:
can you just make inactive ones grey instead

PR #3050, opened 2022-12-13
Improve Auto City

RickyTheRacc:
**Misc**
- Moved getBreakDelta and getBreakProgress methods from PacketMine to Blockutils
- Improved the methods to find cityblocks and surroundblocks

**Auto City**
- Added different target and break ranges
- Improved the breaking of the block
- Added silent switch
- Added rendering
- Added the ability to choose if your hand swings
- Added chat info setting
- Targeting now works based off closest angle instead of lowest distance


RickyTheRacc:
What more would you want added?

RickyTheRacc:
Should be more clear now I think

seasnale (changes requested):

PR #3055, opened 2022-12-14
Optimize validHole check

RickyTheRacc:
Uses mc.world.getOtherEntities to check for entities (also fixes bug with ignoring collisions when force fill was pressed)

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/HoleFiller.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/HoleFiller.java
@@ -26,9 +26,11 @@ import meteordevelopment.orbit.EventPriority;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.block.Blocks;
+import net.minecraft.entity.TntEntity;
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/combat/HoleFiller.java
```diff
@@ -312,23 +317,32 @@ private void onRender(Render3DEvent event) {
...
+        if (mc.player.getBlockPos().equals(testPos)) return false;
+        if (distance(mc.player, testPos, false) > placeRange.get()) return false;
+        if (mc.world.getBlockState(testPos).getBlock() == Blocks.COBWEB) return false;
+
+        if (((AbstractBlockAccessor) mc.world.getBlockState(testPos).getBlock()).isCollidable()) return false;
+        testPos.add(0, 1, 0);
+        if (((AbstractBlockAccessor) mc.world.getBlockState(testPos).getBlock()).isCollidable()) return false;
+        testPos.add(0, -1, 0);
+
+        Box box = new Box(pos);
```
seasnale, line 331: this is not optimisation

seasnale (changes requested):

PR #3072, opened 2022-12-17
Texture rotations module

DesiCow:
Obfuscates texture rotations client-side by randomizing the rendering seed.


DesiCow:
💀

MineGame159:
But why?

RickyTheRacc:
So your coords can't be reversed by using grads rotations etc. while screensharing, this would actually be pretty useful imo qnd is definitely unique

DesiCow:
> So your coords can't be reversed by using grass rotations etc. while screensharing, this would actually be pretty useful imo qnd is definitely unique

Thanks, thats the purpose of this module.
Also, It uses system.currentTime() to make it impossible to crack the obfuscation.

seasnale:
> > So your coords can't be reversed by using grass rotations etc. while screensharing, this would actually be pretty useful imo qnd is definitely unique
> 
> 
> 
> Thanks, thats the purpose of this module.
> 
> Also, It uses system.currentTime() to make it impossible to crack the obfuscation.

well it would be more impossible if you just used a random number lol

theoretically if there was timestamp data on a screenshot it would be easier to find than a random number lol

DesiCow:
> > > So your coords can't be reversed by using grass rotations etc. while screensharing, this would actually be pretty useful imo qnd is definitely unique
> > 
> > 
> > Thanks, thats the purpose of this module.
> > Also, It uses system.currentTime() to make it impossible to crack the obfuscation.
> 
> well it would be more impossible if you just used a random number lol
> 
> theoretically if there was timestamp data on a screenshot it would be easier to find than a random number lol

But the random number generators are also just based on current time

seasnale:
you can just use @Unique to create a method in a mixin

idk why you used an interface

also why even make it a method? its only used once

DesiCow:
Closed in favor of #3192

Issue #2922, opened 2022-11-01
Ingoing and outgoing packet logger. 


aRandomMenno:
### Describe the feature

Module; packet logger
Feature; Add a module that logs incoming/outgoing packets. You also add an element in the HUD maybe. 

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

RacoonDog:
> add to Starscript compatibility for ingoing and outgoing packets or add a module for that.
how would that even work?

JustABiscuit:
Then make a module for it?

I-Got-You-Dead:
Retard

RacoonDog:
??? HOW DO YOU STARSCRIPT A PACKET

JustABiscuit:
THEN MAKE A FUCKING MODULE BASTARD.

JustABiscuit:
> > THEN MAKE A FUCKING MODULE BASTARD.
> 
> already exists
> 
> TELL ME HOW YOU THINK STARSCRIPT WOULD FUNCTION WITH PACKETS THOUGH

Which module in meteorclient then? to see how many packtes are incoming and outgoing?

JustABiscuit:
> > THEN MAKE A FUCKING MODULE BASTARD.
> 
> already exists
> 
> TELL ME HOW YOU THINK STARSCRIPT WOULD FUNCTION WITH PACKETS THOUGH

Bro, you stupid.
Then make a module with java or js smth.

RacoonDog:
- starscript is a text formatting language
- packets are complex structures containing a ton of info and there are hundreds incoming and outgoing per second

how the FUCK would you fuse the two

JustABiscuit:
> * starscript is a text formatting language
> 
>     * packets are complex structures containing a ton of info and there are hundreds incoming and outgoing per second
> 
> 
> how the FUCK would you fuse the two

Idk you are the programmer here, and otherwise you could also make it so the ingoing and outgoing packets have their own .txt file.

JustABiscuit:
> * starscript is a text formatting language
> 
>     * packets are complex structures containing a ton of info and there are hundreds incoming and outgoing per second
> 
> 
> how the FUCK woul

> * starscript is a text formatting language
> 
[...]

PR #3077, opened 2022-12-19
add packet logger module

Cr7mbl3:
Implements [Ingoing and outgoing packet logger. #2922](https://github.com/MeteorDevelopment/meteor-client/issues/2922)


I-Got-You-Dead:
Why not use addons

Cr7mbl3:
makes sense
see https://github.com/Cr7mbl3/meteor-packet-logger

I-Got-You-Dead:
👍

PR #3082, opened 2022-12-19
`PooledList` util class

RacoonDog:
Replace usages of separate `Pool` and `ArrayList` into a fused `PooledList`
+ fix usage of `Pool` in VoidESP, PacketMine & VeinMiner
+ fuse 2 loops in RenderUtils


src/main/java/meteordevelopment/meteorclient/utils/misc/PooledList.java
```diff
@@ -0,0 +1,84 @@
...
+        pool = new Pool<>(producer);
+    }
+
+    public T get() {
+        T obj = pool.get();
+        list.add(obj);
+        return obj;
+    }
+
+    public T peekList() {
```
MineGame159, line 30: Can you rename this method to `getFirst()` to be more clear. Also make it return null if the list is empty since right now it will throw an exception. Also while you are at it, why not add a `getLast()` method too.

src/main/java/meteordevelopment/meteorclient/utils/misc/PooledList.java
```diff
@@ -0,0 +1,84 @@
...
+        list.clear();
+    }
+
+    public void free(T obj) {
+        pool.free(obj);
+        list.remove(obj);
+    }
+
+    /** Required instead of {@link PooledList#free(Object)} when in an iterator loop. {@link PooledList} does not support freeing in foreach loops. */
+    public void free(T obj, Iterator<T> iterator) {
```
MineGame159, line 49: Instead of having this method what you can do is make a custom `private class PooledListIterator implements Iterator<T> { ... }` which will basically be a wrapper over the normal iterator returned from `list.iterator()` but it will override the `remove()` method to call `super.remove();` and `free(obj);`. Then in the `iterator()` method you can just return a new instance of the custom iterator and code using this class can just normally call `it.remove()`.

MineGame159 (changes requested):

src/main/java/meteordevelopment/meteorclient/utils/misc/PooledList.java
RacoonDog: Problem with that is the other `get` method removes an item from the pool and adds it to the list, while this one peeks the top of the list only. I think it'd be clearer to rename `get` to something else

MineGame159: Well yeah, I also made a comment for that.

src/main/java/meteordevelopment/meteorclient/utils/misc/PooledList.java
RacoonDog: I genuinely have no clue how to implement it, currently it uses `list.iterator()` which is a private implementation that I can’t just extend so I’d have to reimplement it or something

PR #3085, opened 2022-12-20
KillAura Improvements

RickyTheRacc:
- Can now sync with TPS
- Can pause when the server is lagging
- Uses hitboxes in range calcs
- Auto shield breaker (so epic)
- Remove settings that are only useful on smps (cope)

Also adds new TimerUtil object for measuring passed MS and ticks, will be also used in CA etc in later PRs :muscle:


RickyTheRacc:
Give an example

I-Got-You-Dead:
Only on look, hit chance and so on

I-Got-You-Dead:
An example of the use case would be protecting a baritone bot or afk mob farming

RickyTheRacc:
Neither of those are flagged by anticheat lmfao, rotate is the same thing as only on look except it doesn't require making you look "legit", hit chance has never been checked against, and "so on" means you ran out of examples

I-Got-You-Dead:
💀

I-Got-You-Dead:
Maybe not on a strict ac

I-Got-You-Dead:
Although I agreed random tp should be removed, that’s retarded af

RickyTheRacc:
*On any anticheat

I've tested this stuff out on every 1.19 server I've got knowledge of, and it works everywhere I could find lmao. There has never to my knowledge been a check for how successful your hits are on an anarchy server, and only on look is just worse rotations

I-Got-You-Dead:
> *On any anticheat
> 
> I've tested this stuff out on every 1.19 server I've got knowledge of, and it works everywhere I could find lmao. There has never to my knowledge been a check for how successful your hits are on an anarchy server, and only on look is just worse rotations

I mean igs, if this gets merged I might just make my own ka or something then

RickyTheRacc:
Mmmmm some oddness rn gimme asec

PR #3094, opened 2022-12-22
Improve Crystal Aura

RickyTheRacc:
Redo how pausing works and add more pausing settings, add several new render modes, more auto switch settings, ignore naked players


I-Got-You-Dead:
:trollface:

PR #3096, opened 2022-12-22
Improve Nametags

RickyTheRacc:
Reorganize settings, allow customization of colors, show totem pops from Notifier


I-Got-You-Dead:
@RickyTheRacc bad code <3

PR #3097, opened 2022-12-22
Middle Click Extra -> Keybind Action

RickyTheRacc:
No longer requires the item to be in your hotbar 
Merged Middle Click Friend
Renamed to "Bind Click Extra"
Made the action key configurable


seasnale:
the name is retarded, makes no sense

seasnale:
also remove those comments

RickyTheRacc:
I couldn't think of anything better ig, if you have ideas

Also rip

seasnale:
also i think middle click friend should be its own module

RickyTheRacc:
I would disagree, I mean if you're using both at the same time then like 
Why?

RickyTheRacc:
As for the name Lemme think then ig

PR #3098, opened 2022-12-22
Improve Nametags

RickyTheRacc:
Reorganize settings, allow customization of all the colors, does not show pops :T


MineGame159:
Why are items and other nametags that are not for players using the ping color? Also the description for ignore self is wrong.

PR #3127, opened 2022-12-28
use try-with-resources

RacoonDog:
From `Files.list()`'s javadocs:
> This method must be used within a try-with-resources statement or similar control structure to ensure that the stream's open directory is closed promptly after the stream's operations have completed.


seasnale:
why use var

ThebestkillerTBK:
this take too long lol

PR #3142, opened 2023-01-02
DiscordRPC NPE fix

RacoonDog:


RacoonDog:
Before someone asks, yes there is a reproducible situation where during a tick `Utils.canUpdate()` returns false, yet `mc.currentScreen` is null

Issue #1122, opened 2021-06-25
[Suggestion] TeleportReach


a-couple-objects:
**Describe the feature**
A module which teleports you in range of your target, attacks, and teleports you back. Inertia has this for an example

mmvanheusden:
This is build into kill aura

RedCarlos26:
This is already a thing in the kill-aura module @mmvanheusden

PR #3146, opened 2023-01-02
TP-Hit mode for reach module

DesiCow:
As suggested in #1122 , this setting teleports(in multiple segments) the player closer to target, attacks, and returns back to original location.


Wide-Cat:
you need to do rotation stuff in the Pre tick because of how theyre applied, otherwise it'll be a tick behind, also `new Thread` and `Thread.sleep()` 😩

I-Got-You-Dead:
Bruh no more trolling small children :l

I-Got-You-Dead:
Now we gotta deal with Skript kiddies using tp range 💀

I-Got-You-Dead:
Sadage

DesiCow:
> you need to do rotation stuff in the Pre tick because of how theyre applied, otherwise it'll be a tick behind, also `new Thread` and `Thread.sleep()` 😩
How would you delay sending the packets?

DesiCow:
bad code

PR #3162, opened 2023-01-09
Use better bootstrap check for Registry

ThebestkillerTBK:
:trollface:


MineGame159:
How is this better.

ThebestkillerTBK:
because you said `TODO: Probably extremely retarded but seems to work` :trollface:

Wide-Cat:
Not wrong

MineGame159:
Well in my opinion this is even more retarded

PR #3163, opened 2023-01-10
Implement Alphabetical sorting

### LINKED ISSUE #1958: [Suggestion] Alphabetically sort module list
### Describe the feature

As said in the title, I think it'd be nice to have the option to sort the module list alphabetically.

It's easier to find what you need in a sorted list.
The sorting name could be `Alphabetic` or `Alphasort`.
We already have a drop down menu so it could be added to it easily.

- [x] This feature doesn't already exist in the client.

Lovecrafts-Cat:
This PR introduces an alphabetical sorting mode for the `ActiveModulesHud` element.
image

While it may not be as pleasant to look at as the Biggest/Smallest sorting, it makes it easier to locaate modules.
Also, this closes #1958 made long ago by me.


Lovecrafts-Cat:
Just noticed this way it's not using `getModuleWidth`, but the modes behave just like before.
If this is not intended, you can revert the `e.title.length()` calls with the previously mentioned getter.

PR #3192, opened 2023-01-19
Better TextureRotations

DesiCow:
This pr presents improvement upon my previous pr #3072. Instead of randomizing the rendering seed, **same seed** is now returned for all blocks regardless of their position. This eliminates the possibility of brute forcing the original orientation and thus, helps spoofing coordinates of the player.


StaidChannel708:
perhaps this should be integrated within the no render module under a name of block rotations

PR #3194, opened 2023-01-19
Proxies minor patch

Lovecrafts-Cat:
Minor cleanup, replaced string concat with .format for readability, corrected equals method.

NOTE: the current check used to differentiate between SOCKS v4 and v5 is not that effective.
Whenever you get proxies from any online list, they don't have the tag that the current regex filter tries to match.

A possible solution would be trying to connect with v5, and fallback to v4 if that doesn't work.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
@@ -26,21 +26,20 @@ public class ProxiesImportScreen extends WindowScreen {
     public ProxiesImportScreen(GuiTheme theme, File file) {
         super(theme, "Import Proxies");
         this.file = file;
-        this.onClosed(() -> {
-            if (parent instanceof ProxiesScreen screen) {
-                screen.reload();
-            }
+        onClosed(() -> {
+            if (parent instanceof ProxiesScreen screen) screen.reload();
         });
     }
 
     @Override
     public void initWidgets() {
         if (file.exists() && file.isFile()) {
-            add(theme.label("Importing proxies from " + file.getName() + "...").color(Color.GREEN));
+            add(theme.label("Importing proxies from %s...".formatted(file.getName())).color(Color.GREEN));
             WVerticalList list = add(theme.section("Log", false)).widget().add(theme.verticalList()).expandX().widget();
             Proxies proxies = Proxies.get();
             try {
-                int pog = 0, bruh = 0;
+                int success = 0;
+                int failure = 0;
                 for (String line : Files.readAllLines(file.toPath())) {
                     Matcher matcher = Proxies.PROXY_PATTERN.matcher(line);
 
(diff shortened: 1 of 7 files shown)


Lovecrafts-Cat:
After a long battle with the inner workings of the SOCKS protocol, the multiple ways to detect its version, and the general mess involved with that, I can say that the current implementation within Meteor is at least useable.

I'll close this PR, but before that, let me list some improvements I'd love to see someone else tackle:

- The proxy import screen should not add proxies if connection to them is refused. If the user can't connect to the given proxy, it's useless to add it in the first place.
- The protocol version should be calculated through the connection request response (keeping in mind that the versions are `v4`, `v4a` and `v5`). Proxies that use v4a are usually treated as v5 since they can use domains instead of IPs, but don't have support for password auth. Also, it's worth noting that some proxy servers may be configured to present themselves as a different protocol version than they actually are, so even using a library may not give you the correct information in all cases.
- The protocol version shouldn't be derived the way it currently is. Proxy lists available online never append the `@socks4` or `@socks5` marker to them, so the regex match for that section is unuseable.

I leave here an example of a test class to verify protocol versions, though it fails to discriminate between `v4a` and `v5`:
```java
import java.io.*;
import java.net.InetSocketAddress;
import java.net.Socket;
import java.nio.file.Files;
import java.nio.file.Paths;
import java.util.concurrent.*;

class Scratch {
    private static final String path = "PATH_TO_PROXY_TXT";
    public static class SocksProxyChecker {
    private static final ExecutorService threadPool = Executors.newFixedThreadPool(10);

    public static String checkProxy(String ip, int port) {
        try {
            CompletableFuture<String> future = CompletableFuture.supplyAsync(() -> {
                try (Socket socket = new Socket()) {
                    socket.connect(new InetSocketAddress(ip, port), 2000);
                    InputStream in = socket.getInputStream();
                    OutputStream out = socket.getOutputStream();
                    out.write(new byte[]{0x05, 0x01, 0x00});
                    out.flush();
                    byte[] response = new byte[2];
                    in.read(response);

                    if (response[0] != 0x05) return "Invalid response from proxy";
                    return switch (response[1]) {
                        case 0x00 -> "SOCKS v5";
                        case 0x01 -> "SOCKS v4";
                        default -> "Unsupported SOCKS version";
                    };
                } catch (IOException e) {
                    return "Unable to connect to proxy";
                }
            }, threadPool);
            return future.get(3, TimeUnit.SECONDS);
        } catch (InterruptedException | ExecutionException | TimeoutException e) {
            return "Unable to connect to proxy";
        }
    }
}


    public static void main(String[] args) {
        try (BufferedReader br = new BufferedReader(new InputStreamReader(Files.newInputStream(Paths.get(path))))) {
            String line;
            while ((line = br.readLine()) != null) {
                String ip = line.split(":")[0];
                int port = Integer.parseInt(line.split(":")[1]);
                System.out.println("Connecting to %s:%d with result: %s".formatted(ip, port, SocksProxyChecker.checkProxy(ip, port)));
            }
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

As for the proxies used to thest this, they can be found at [this](https://github.com/TheSpeedX/PROXY-List) GitHub repo.

PR #3209, opened 2023-01-23
Added bogus-accept resource-packs option to ServerSpoof module

DesiCow:
Automatically sends fake packets to the server that client accepted their resource pack.


ThebestkillerTBK:
Why use while bro

DesiCow:
> Why use while bro

uhh mc.getNetworkHandler() is sometimes null when joining a server

ThebestkillerTBK:
> > Why use while bro
> 
> uhh mc.getNetworkHandler() is sometimes null while joining a server, that's why I do it in MeteorExecutor service

Have you tested it? The network handler already received the resource packet tho, it cannot be null.

DesiCow:
> > > Why use while bro
> > 
> > 
> > uhh mc.getNetworkHandler() is sometimes null while joining a server, that's why I do it in MeteorExecutor service
> 
> Have you tested it? 
yes

PR #3212, opened 2023-01-24
AutoTool : Prefer Fortune for ores and crops option

### LINKED ISSUE #1796: Auto Tool should have "Fortune for Ores" option[Suggestion]
### Describe the feature

Sometimes you want to mine some stone with Silktouch but the ores with Fortune. So it would be nice if it could auto switch for u.

DesiCow:
Mines ores and crops only with the fortune enchantment. Closes #1796


ManInMyVan:
We need this

PR #3213, opened 2023-01-24
Improve server plugin search method and compact code

LLKawi:
Everything just works fine.


ThebestkillerTBK:
I re-added MassScan because CommandBlocker plugin.

Wide-Cat:
conflicts

PR #3216, opened 2023-01-25
No insta break

ThebestkillerTBK:


Wide-Cat:
would this not be better as an option in speedmine, because i feel like its more relevant there

ThebestkillerTBK:
This is different. This adds delay when you can instant break the block, so you wouldnt misbreak other blocks.

PR #3217, opened 2023-01-25
Fix a trolling method

ThebestkillerTBK:


MineGame159:
Can you rename the methods to be more descriptive? I assume the `P` there stands for prefix so just do `infoPrefix`. Also while you are at it can you replace the `ChatUtils.infoP();` calls to the module's `this.info();` methods so the prefix is not passed there every time.

PR #3218, opened 2023-01-25
Added Macro execute command

DesiCow:
Syntax:
`.macro <macro-name>` to execute the macro

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/Systems.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/Systems.java
@@ -40,9 +40,9 @@ public class Systems {
         config.load();
 
         add(new Modules());
+        add(new Macros());
         add(new Commands());
         add(new Friends());
-        add(new Macros());
         add(new Accounts());
         add(new Waypoints());
         add(new Profiles());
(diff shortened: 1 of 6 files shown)


arlomcwalter:
macros triggering macros :trollface:

DesiCow:
> macros triggering macros :trollface:

That's the motivation behind this pr :trollface: 

> why did you re order the tab bar

So MacroCommand can access Macros, but it doesn't affect the order of tabs ingame, i.e. tabs are still displayed in [same order](https://github.com/MeteorDevelopment/meteor-client/blob/master/src/main/java/meteordevelopment/meteorclient/gui/tabs/Tabs.java#L19-L26).

MineGame159:
I am failing to understand the reason behind the reorder.

DesiCow:
> I am failing to understand the reason behind the reorder.
> 
> EDIT: Oh I think I see why, when the command is created the `MacroArgumentType` is loaded for the first time, so the static fields are initialized, which accesses the `Macros` class. Anyways, the examples should probably be recomputed every time it is called, or atleast cached because for first time users they won't have any macros.

Thanks for the review. And yes, thats the reason for reorder.

PR #3243, opened 2023-01-30
Allow meteor client to work without baritone

MonkeySaint:
This should make updating to 1.20 easier since you wont need to wait for baritone to update. All changes have no real effect on performance or the end user.


ehylo:
If baritone isn't updated by the time meteor is (unlikely, wag is pretty fast on updating baritone) minegame already comments out all of baritone, so this pr is not needed

RacoonDog:
i don't think wrapping each baritone usage with a try-catch is the best way to achieve this anyways

ThebestkillerTBK:
Using try catch is kinda stupid lol

MonkeySaint:
This is more or less a proof of concept to show that it can be done relatively easily, and in my opinion should be done just for easy updates. Not meant to actually remove baritone. Try catch is pretty dumb though, was the only way I could think without using fabric api though. That being said should have just on init check if they had baritone loaded instead of every usage of it

RacoonDog:
> was the only way I could think without using fabric api though

You can check whether baritone is present or not using `FabricLoader.getInstance().isModLoaded()` with requiring fapi, and I think the best way to implement such a system would be using a dedicated util class which would basically wrap the baritone api? idk

MonkeySaint:
> > was the only way I could think without using fabric api though
> 
> You can check whether baritone is present or not using `FabricLoader.getInstance().isModLoaded()` with requiring fapi, and I think the best way to implement such a system would be using a dedicated util class which would basically wrap the baritone api? idk
> 

Didn't know that existed, what if instead of each occurrence of baritone calls, only load the used modules if baritone is loaded.

ThebestkillerTBK:
Also the way you deal with the infinity miner seems bad, just automatically toggle on active if it doesnt exist
Better rewrite these code

MonkeySaint:
> Also the way you deal with the infinity miner seems bad, just automatically toggle on active if it doesnt exist
> Better rewrite these code 

Yeah I'm going to redo it with some changes that I mentioned in my previous comment

MonkeySaint:
Maybe a new category for modules that use baritone?

MonkeySaint:
On the latest commit it just disables the modules and adds a slight change ti the waypoints class (to make sure the modules exists) and to the ClientPlayNetworkHandlerMixin to get the baritone prefix.

MonkeySaint:
Using this for I updated to 23w04a to show it works. 
https://github.com/MonkeySaint/compatable-meteor-client/tree/23w04a

The main change is that packet is moved into a different package.
To use without baritone you need to keep baritone as a library and hardcode the game version.
I think this will make it easier to maintain and allow for people to commit minecraft version updates without directly being on the meteor team.

ThebestkillerTBK:
You also did not check mixin dependency, should add baritone to MixinPlugin

ThebestkillerTBK:
Maybe just let baritone become an dependency and should be installed separately? cuz addons uses baritone quite often.

MonkeySaint:
> You also did not check mixin dependency, should add baritone to MixinPlugin
> 
> We should not totally disable the module and the baritone should be modImplementation.
> 
> the Meteor-baritone should not be used if this is done. 

Yeah I don't think it matters to check the mixin dependency since as that is it runs fine without baritone. It's not for the end user to use it without baritone it's for developers.

MineGame159:
This won't even work. Sure it makes it possible to load Meteor without Baritone (if someone manually removed it from the Meteor jar) but it doesn't allow us to update faster than Baritone. The dependency on Baritone is still there, it still has to be present there when building Meteor so the compiler can reference the methods.

PR #3247, opened 2023-01-30
Improve Hide HUD

### LINKED ISSUE #2075: [Suggestion] Improve Hide Hud
### Describe the feature

Make it hide the hud when in inventory screens and the pause menu, and make it show the hud again while editing the hud

DesiCow:
Hides **Meteor** hud when user is in **inventory** or **pause menu**. Closes #2075

Note : This feature should not be confused with the [Hide HUD in GUI Settings](https://github.com/MeteorDevelopment/meteor-client/blob/b67f0a599333d407284f5f6c3fcdf256a4182506/src/main/java/meteordevelopment/meteorclient/gui/GuiTheme.java#L273) option, which hides both the Minecraft and Meteor HUD when the user is in the **Meteor GUI**.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/hud/Hud.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/hud/Hud.java
@@ -105,7 +105,7 @@ public class Hud extends System<Hud> implements Iterable<HudElement> {
     );
 
     private boolean resetToDefaultElements;
-    private boolean wasMenuScreen, wasHudHidden;
+    private boolean wasHudHidden;
 
     public Hud() {
         super("hud");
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/hud/Hud.java
```diff
@@ -234,6 +246,17 @@ private void onCustomFontChanged(CustomFontChangedEvent event) {
...
 
+    @EventHandler(priority = EventPriority.LOWEST)
+    private void onOpenScreen(OpenScreenEvent event) {
+        if (hideInMenu.get()) {
+            if (!wasMenuScreen) wasHudHidden = !this.active;
+
+            if (event.screen instanceof GameMenuScreen || event.screen instanceof HandledScreen) this.active = false;
+            else if (!wasHudHidden) this.active = true;
+        }
+        wasMenuScreen = event.screen instanceof GameMenuScreen || event.screen instanceof HandledScreen;
```
MineGame159, line 257: Why are there 2 fields `wasMenuScreen` and `wasHudHidden` screen? If I am not missing anything then inside this event handler you can check if the new screen is `GameMenuScreen` or `HandledScreen` and if yes then set `this.active` to false and also some other new field to true if the HUD was active. And if the screen is not one of the two then check if the new field is set to true and re-enable the HUD.

RickyTheRacc:
I am a BIG SUPPORTER of this PR

src/main/java/meteordevelopment/meteorclient/systems/hud/Hud.java
DesiCow: This was meant for some odd cases when for example player is already in inventory/menu screen and somehow this or toggle hud option was used using code, macros, etc.

Wide-Cat:
if you open the creative menu it hides the hud, but when you close the menu it doesnt show the hud it again

PR #3253, opened 2023-01-31
Anti packet kick update

### LINKED ISSUE #3242: Antiban
### Describe the feature

Anti ban would allow you to not get bookbanned or chunkbanned by either canceling large amounts of packets (like packet canceler) Or you do what you want to do but this would would also allow you to make an chunk ban and turn it on while your still in the chunk. Or simply as an protection from bookbans. Also antipacketkick does not cancel bookbans, At least for me.

ThebestkillerTBK:
closes #3242 by dropping corrupted packets.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/ClientConnectionMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/ClientConnectionMixin.java
@@ -5,11 +5,14 @@
 
 package meteordevelopment.meteorclient.mixin;
 
+import io.netty.channel.ChannelHandlerContext;
 import meteordevelopment.meteorclient.MeteorClient;
 import meteordevelopment.meteorclient.events.packets.PacketEvent;
 import meteordevelopment.meteorclient.events.world.ConnectToServerEvent;
 import meteordevelopment.meteorclient.systems.modules.Modules;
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/mixin/ClientConnectionMixin.java
```diff
@@ -55,4 +59,12 @@ private void onSendPacketHead(Packet<?> packet, CallbackInfo info) {
     private void onSendPacketTail(Packet<?> packet, CallbackInfo info) {
         MeteorClient.EVENT_BUS.post(PacketEvent.Sent.get(packet));
     }
+
+    @Inject(method = "exceptionCaught", at = @At("HEAD"), cancellable = true)
+    private void exceptionCaught(ChannelHandlerContext context, Throwable throwable, CallbackInfo ci) {
+        if (throwable instanceof IOException && Modules.get().get(AntiPacketKick.class).catchExceptions()) {
+            if (Modules.get().get(AntiPacketKick.class).logExceptions.get()) ChatUtils.warning("Caught exception: %s", throwable);
```
MineGame159, line 66: This should use `Modules.get().get(AntiPacketKick.class).warning(...);` to better highlight it is coming from this module.

PR #3264, opened 2023-02-01
ESP but faster

RacoonDog:
1. `ESP.getColor()` is a fairly expensive method and it was being called for every `Entity` regardless of whether ESP was even enabled, so I turned it into a `Supplier<Color>` that's only called if it can render the `Entity`.
2. `ESP.getColor()` could always return `null`, so I marked it as `@Nullable`
3. `ESP.shouldSkip()` used to return `true` if `ESP.getFadeAlpha() == 0`, I replaced that with a null check, since `ESP.getColor()` already returned `null` if `ESP.getFadeAlpha() == 0`
4. `TextUtils.getColoredCharacterCount()` now returns an `Object2IntMap<Color>` instead of a `Map<Color, Integer>`
5. Rewrote `TextUtils.getMostPopularColor()` to not use streams
6. `TextUtils.toColorTextList()`, replaced `str.equals("")` with `str.isEmpty()`


Wide-Cat:
that was quick

PR #3268, opened 2023-02-02
add getCustomPrefix(String packageName)

tympanicblock61:


src/main/java/meteordevelopment/meteorclient/utils/player/ChatUtils.java
```diff
@@ -57,6 +57,14 @@ public static void registerCustomPrefix(String packageName, Supplier<Text> suppl
     public static void unregisterCustomPrefix(String packageName) {
         customPrefixes.removeIf(pair -> pair.getLeft().equals(packageName));
     }
+    
+    public static Pair<String, Supplier<Text>> getCustomPrefix(String packageName) {
+        for (Pair<String, Supplier<Text>> pair : customPrefixes) {
+            if (pair.getLeft().equals(packageName)) {
+                return pair;
```
GreemDev, line 64: Is there a need to return the pair directly? The left side of the pair is just the same package name that you input to the method

tympanicblock61: no, your correct mb

tympanicblock61:
Why close there's nothing wrong with it and it's useful

PR #3282, opened 2023-02-03
Remove FakeClientPlayer due to bad implementation and it is broken

ThebestkillerTBK:
Bad implementation + broken after 1.19.3


arlomcwalter:
what's wrong with the implementation?
as far as i can tell it just needs to be fixed, not removed

ThebestkillerTBK:
new fix delivered

ThebestkillerTBK:
new fix delivered

PR #3287, opened 2023-02-03
Added onlyOnWeapon option to Hitboxes module

### LINKED ISSUE #2456: [Suggestion] item or slot specific hitbox
### Describe the feature

I like to use hitboxes a lot in my personal world or smps as it just makes fighting mobs less boring and lets you focus on building, but i don't really use them much at all for anything else, would be nice if you had an option for hitboxes to expand only while having some hand picked items picked or being on a specific slot

DesiCow:
Only modifies hitbox when holding a weapon in hand. Closes #2456


arlomcwalter:
should add bows too

Wide-Cat:
does that actually work? surely arrows are handled by the server

arlomcwalter:
maybe idk, worth a test

DesiCow:
Tested: All the projectiles do NOT work with client-side enlarged hitboxes.

PR #3295, opened 2023-02-04
Settings Code modification and finish a todo

LLKawi:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Make all setting constructors are not accessible so the Builder is always called

also finish a todo

## Related issues

N/A

# How Has This Been Tested?

No crash on loading and quitting.

# Checklist:


arlomcwalter:
why did you make some of the classes final

ThebestkillerTBK:
> why did you make some of the classes final

Because they should never be inherited

RacoonDog:
i can say i have definitely extended multiple setting classes in the past

RacoonDog:
but yeah seansail private constructor stops you from extending a class even if its not final

arlomcwalter:
> but yeah seansail private constructor stops you from extending a class even if its not final

well ok but i don't think that's necessarily a reason to add final to those classes in this pr

arlomcwalter:
the constructors should all be protected, the ones that aren't are mistakes, not deliberate 

@ThebestkillerTBK could you make all the classes and constructors consistent in this pr?

LLKawi:
closing due to inactivity

PR #3423, opened 2023-03-09
Removed useless square-rooting

DAMcraft:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

Calculating sqrt is an operation that requires a lot of resources. As Double.compare only compares two numbers, we don‘t need to calculate the sqrt, as the square of whatever that number is is always bigger as well. 

(example)
sqrt(4) > sqrt(9) 
can be converted into
4 > 9

in other words,
if sqrt(x) > sqrt(y), then x > y
if sqrt(x) < sqrt(y), then x < y
if sqrt(x) == sqrt(y), then x == y

So, dont sqrt if it isnt needed, as computers need to calculate a lot

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/entity/TargetUtils.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/entity/TargetUtils.java
@@ -101,6 +101,6 @@ public class TargetUtils {
         double e1pitch = Math.abs(Rotations.getPitch(e1) - mc.player.getPitch());
         double e2pitch = Math.abs(Rotations.getPitch(e2) - mc.player.getPitch());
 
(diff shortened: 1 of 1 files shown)


DAMcraft:
Made a typo in the commit message, meant sqrt(9), not sqrt(0)

DAMcraft (commented):
Made a typo in the commit message, meant sqrt(9), not sqrt(0)

RacoonDog:
instead of casting and using double compare, you can just use int/long compare now

DAMcraft:
> instead of casting and using double compare, you can just use int/long compare now

honestly i edited this on my phone, just wanted to add casting to ensure nothing gets bronked

RacoonDog:
> > instead of casting and using double compare, you can just use int/long compare now
> 
> honestly i edited this on my phone, just wanted to add casting to ensure nothing gets bronked
> 
> there is a good possibility this doesn’t even work 👍

54ED3E1F-301B-45C8-ADB4-5E8D9C661D7E
:shock:

PR #3439, opened 2023-03-11
Fixed auto gap not eating normal gapples

### LINKED ISSUE #3188: auto gap
### Describe the bug

when auto gap is enabled and health slider was put to 6, when my hp dipped below 6 it did not gap (normal gaps not enchanted)

### Steps to reproduce

```bash
turn on auto gap set health slider to any value and it will not gap
```

### Link to crash report/log if applicable (https://mclo.gs)

_No response_

### Meteor Version

0.52-1753

### Minecraft Version

1.19.3

### Operating System

Windows

- [X] This bug wasn't already reported (I have searched bug reports on GitHub).

- [X] This is a valid bug (I am able to reproduce this on the latest dev build).

DesiCow:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Fixed Auto Gap not eating normal gapples if `preferEGap` was enabled and no e-gapples were found in Hotbar.

## Related issues

Resolves #3188

# How Has This Been Tested?

Tested on dedicated paper server.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoGap.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoGap.java
@@ -276,11 +276,8 @@ public class AutoGap extends Module {
     }
 
     private int findSlot() {
-        boolean preferEGap = this.preferEGap.get();
-        if (requiresEGap) preferEGap = true;
-
+        boolean preferEGap = this.preferEGap.get() || requiresEGap;
         int slot = -1;
-        Item currentItem = null;
 
         for (int i = 0; i < 9; i++) {
             // Skip if item stack is empty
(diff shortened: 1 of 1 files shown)


MineGame159:
Well you introduced the opposite bug. If I understand it correctly then now it won't eat EGAPs is `preferEgap` is false. Which shouldn't happen, it should still eat them just not prioritize them.

DesiCow:
> Well you introduced the opposite bug. If I understand it correctly then now it won't eat EGAPs if `preferEgap` is false. Which shouldn't happen, it should still eat them just not prioritize them.

It was an intentional feature, I treated that setting as "allow egap" to only allow eating normal gaps as enchanted ones are expensive.

MineGame159:
Well then the name / description should be changed as right now it makes no sense

DesiCow:
> Also I don't see how this fixes the original issue. To me it looks like it will only eat normal apples if `requiresEGap` is false which can be simplified for this example to just mean if the setting `allow-egap` is false. But based on the description it should fallback to eating normal golden apples if there are no enchanted ones.

Only Egap provides fire resistance and resistance effects, that's the reason it doesn't fallback to normal gapple if those effects run out.
However, it does fallback to normal gapple in other cases such as when player's health reaches below threshold.

PR #3452, opened 2023-03-17
Fix Build Hight to work on 1.17+ worlds

Ompyl:
I fixed the bug that in 1.17+ worlds blocks would be placed under the block you clicked on if you are on y level 255 or higher

I had to use an If because mc.world.getDimension().height() for some reason returns 382 on 1.18 worlds


Ompyl:
sorry if I made somethin wrong with the pull request, it's my first time

MineGame159:
You are assuming the world always has the height of 320 on new versions which is simply not true. Servers can customise it. If the height method is returning wrong values on old servers then it either means the plugin the server uses to support new clients has a bug or viafabric / multiconnect has a bug.

MineGame159:
Also just realised what are you going on about. That is the correct height 1.18 worlds should have.

PR #3456, opened 2023-03-18
Fix Scaffold FastTower

headzz06:
If you used scaffold fasttower without any blocks in your hotbar and tried to tower upwards, it would make you start flying in the air

## Type of change

- Bug fix

## Description

Fixes it by checking if the player has any blocks in their inventory before attempting to fast tower

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
@@ -306,7 +306,7 @@ public class Scaffold extends Module {
                     mc.player.setVelocity(0, -0.28f, 0);
             }
         } else {
-            if (mc.options.jumpKey.isPressed() && !mc.options.sneakKey.isPressed() && fastTower.get()) {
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
```diff
@@ -306,7 +306,7 @@ private void fastTower(boolean down, BlockPos checkBlock) {
                     mc.player.setVelocity(0, -0.28f, 0);
             }
         } else {
-            if (mc.options.jumpKey.isPressed() && !mc.options.sneakKey.isPressed() && fastTower.get()) {
+            if (mc.options.jumpKey.isPressed() && !mc.options.sneakKey.isPressed() && fastTower.get() && InvUtils.testInHotbar(stack -> stack.getItem() instanceof BlockItem)) {
```
DesiCow, line 309: Suppose I have only sand or any unselected block in my hotbar, wouldn't it still start flying in air?

headzz06: okay i changed it to use the validItem method instead of checking if its an instance of BlockItem

TheAtomicPunk: balls

PR #3477, opened 2023-03-21
use intended method for proxies

FlorianMichael:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

This pull request moved the code to add the proxies to the new method added with Minecraft 1.19.4, this fixes the incompatibility with the new [ViaFabricPlus](https://github.com/FlorianMichael/ViaFabricPlus) version, because of the Bedrock support I have to overwrite the complete connect

# Checklist:


RacoonDog:
Should check if address is valid, cause it looks like it could throw an unchecked exception otherwise

FlorianMichael:
@MineGame159 I forgot to check if the addHandlers is called by the ClientConnection, because it is used in the ServerNetworkIo as well, you could add an if for it

MineGame159:
Thanks, done

PR #3479, opened 2023-03-21
Revert "use intended method for proxies"

BDARmech:
Reverts MeteorDevelopment/meteor-client#3477


MineGame159:
what

FlorianMichael:
Why do you want to revert that? It's better than the old code and makes the mod compact with my mod

PR #3524, opened 2023-03-26
Fix crash when GameRenderer client is null but normal isnt

xyzaurora:
## Type of change

- [X ] Bug fix
- [ ] New feature

## Description

Weird edge case for 1.19.4, dont know why it happens but lead me to crash everytime i join a world.

## Related issues

None

# How Has This Been Tested?

Works on my machine:tm:

# Checklist:


xyzaurora:
Doesn't seem to be a valid fix, seemed to work at first but now doesn't, pretty sure its one of my other mods interfering with meteor

RacoonDog:
anyways if `client` is null then `Utils.canUpdate()` throws an `NPE`

PR #3540, opened 2023-03-31
Feat: Emit sound on Notifier's player enter and exit

vaxerski:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds a (toggleable) option to emit a sound when a player enters/exits the render distance.

Without this, it's really easy to miss a player when going fast (at least for me)

# How Has This Been Tested?

Joined 9b, works alright.

# Checklist:


Wide-Cat:
Using playSoundFromEntity might have unintended side affects. It sounds like the audio is cut off when the entity moves out of your render distance, presumably because it gets unloaded, and can be very quiet when entering because it's fairly far away. It might be best to use mc.player for both of the entities, so it should always work properly.

vaxerski:
sure, initial thought was about directional audio or something, but fair point.

PR #3545, opened 2023-04-03
Don't send meteor agent to Mojang

FlorianMichael:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

This pull request removes Meteor sending the Meteor UserAgent in API requests to Mojang, which is pretty stupid because you could easily detect and block Meteor users that way.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/screens/CommitsScreen.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/screens/CommitsScreen.java
@@ -32,7 +32,7 @@ public class CommitsScreen extends WindowScreen {
 
         MeteorExecutor.execute(() -> {
             GithubRepo repo = addon.getRepo();
-            Response res = Http.get(String.format("https://api.github.com/repos/%s/compare/%s...%s", repo.getOwnerName(), addon.getCommit(), repo.branch()), true).sendJson(Response.class);
+            Response res = Http.get(String.format("https://api.github.com/repos/%s/compare/%s...%s", repo.getOwnerName(), addon.getCommit(), repo.branch())).sendJson(Response.class);
 
             if (res != null) {
                 commits = res.commits;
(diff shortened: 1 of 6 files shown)


arlomcwalter:
??? why would mojang block requests with that user agent

Talhoid:
agree with seasnail

allinkdev:
> ??? why would mojang block requests with that user agent

Imagine how easy it would be for Mojang/Microsoft, if they so wanted to, to permanently ban every Meteor client user who has ever logged into their Microsoft account with the built-in alt manager, by using the conveniently provided "Meteor Client" user agent you currently use to communicate with their authentication server.

FlorianMichael:
I just don't think it's good that a cheat mod tells the official servers of the game creator when it makes API requests.

arlomcwalter:
> > ??? why would mojang block requests with that user agent
> 
> 
> 
> Imagine how easy it would be for Mojang/Microsoft, if they so wanted to, to permanently ban every Meteor client user who has ever logged into their Microsoft account with the built-in alt manager, by using the conveniently provided "Meteor Client" user agent you currently use to communicate with their authentication server.

ok but literally what reason do they have to do that

arlomcwalter:
> I just don't think it's good that a cheat mod tells the official servers of the game creator when it makes API requests.

ok but meteor is not a cheat mod so mojang wouldn't care

FlorianMichael:
It's your choice, I already see Meteor as a cheat client because it has modules for it, but if you see it differently, that's fine.

FlorianMichael:
I just don't think it's a good idea to tell Mojang that you're using Meteor, I mean Meteor is also spoofing the ClientBrand to "vanilla", so you could make the same argument.

Talhoid:
:bruh:

arlomcwalter:
new information has come to light :troll:

FlorianMichael:
That's what I originally wanted to do, but I wasn't sure if the MeteorClient user agent had any function.

FlorianMichael:
done @arlomcwalter

PR #3546, opened 2023-04-03
Feat: Add offscreen style to tracers

vaxerski:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds an offscreen mode to tracers, which is a less ugly and more efficient method of doing them.
Same feature as I have made in my client, and IMHO it's really neat.

image

# How Has This Been Tested?

Joined 9b9t, works. Singleplayer - works.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
@@ -353,12 +353,12 @@ public class Tracers extends Module {
             else
                 pitch = 90;
         } else {
-            yaw = (float)(Math.atan2(forward.y, forward.x) * 180 / 3.1415f);
+            yaw = (float)(Math.atan2(forward.y, forward.x) * 180 / Math.PI);
             if(yaw < 0)
                 yaw += 360;
 
             tmp = (float)Math.sqrt(forward.x * forward.x + forward.y * forward.y);
-            pitch = (float)(Math.atan2(-forward.z, tmp) * 180 / 3.1415f);
+            pitch = (float)(Math.atan2(-forward.z, tmp) * 180 / Math.PI);
             if(pitch < 0)
                 pitch += 360;
         }
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
@@ -291,9 +291,9 @@ public class Tracers extends Module {
             final Vec2f SCREEN_CENTER = new Vec2f(mc.getWindow().getFramebufferWidth() / 2.f, mc.getWindow().getFramebufferHeight() / 2.f);
 
             Vector3d projection = new Vector3d(entity.prevX, entity.prevY, entity.prevZ);
-            NametagUtils.to2D(projection, 1, false, false);
+            boolean projSucceeded = NametagUtils.to2D(projection, 1, false, false);
 
-            if (!(projection.x < 0.f || projection.x > mc.getWindow().getFramebufferWidth() || projection.y < 0.f || projection.y > mc.getWindow().getFramebufferHeight()))
+            if (projSucceeded && projection.x > 0.f && projection.x < mc.getWindow().getFramebufferWidth() && projection.y > 0.f && projection.y < mc.getWindow().getFramebufferHeight())
                 continue;
 
             projection = new Vector3d(entity.prevX, entity.prevY, entity.prevZ);
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
@@ -277,9 +277,6 @@ public class Tracers extends Module {
         if (mc.options.hudHidden || style.get() != TracerStyle.Offscreen) return;
         count = 0;
 
-        MatrixStack matrices = RenderSystem.getModelViewStack();
-        matrices.push();
-
         Renderer2D.COLOR.begin();
 
         for (Entity entity : mc.world.getEntities()) {
@@ -324,7 +321,6 @@ public class Tracers extends Module {
         }
 
         Renderer2D.COLOR.render(null);
-        RenderSystem.getModelViewStack().pop();
     }
 
     private void rotateTriangle(Vector2f[] points, float ang) {
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
@@ -136,7 +136,7 @@ public class Tracers extends Module {
         .build()
     );
 
-    private final Setting<Double> blinkOffscreenSpeed = sgAppearance.add(new Double.Builder()
+    private final Setting<Double> blinkOffscreenSpeed = sgAppearance.add(new DoubleSetting.Builder()
         .name("blink-offscreen-speed")
         .description("Offscreen's blink speed.")
         .defaultValue(4)
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
@@ -136,7 +136,7 @@ public class Tracers extends Module {
         .build()
     );
 
-    private final Setting<Integer> blinkOffscreenSpeed = sgAppearance.add(new IntSetting.Builder()
+    private final Setting<Double> blinkOffscreenSpeed = sgAppearance.add(new Double.Builder()
         .name("blink-offscreen-speed")
         .description("Offscreen's blink speed.")
         .defaultValue(4)
@@ -363,7 +363,7 @@ public class Tracers extends Module {
     }
 
     private float getAlpha() {
-        float speed = blinkOffscreenSpeed.get() / 4.f;
+        double speed = blinkOffscreenSpeed.get() / 4.0;
         double duration = Math.abs(Duration.between(Instant.now(), initTimer).toMillis()) * speed;
 
         return (float)Math.abs((duration % 1000) - 500) / 500.f;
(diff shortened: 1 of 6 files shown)


I-Got-You-Dead2:
Sexy

I-Got-You-Dead2:
first good new feature in a while

Wide-Cat:
This looks pretty good. One thing - could you hide the relevant setting for linear tracers if offscreen ones are enabled? Aside from that, this is ready to merge.

vaxerski:
How do I do that without damaging the structure? I can't reference offscreen before it's initialized, and the default tracer opts are above the offscreen ones. Adding a `private boolean` seems like trashing

Wide-Cat:
You could maybe add an enum setting above the other settings to choose between linear and offscreen tracers, and then hide the ones not relevant to the selected option

vaxerski:
Done, new enum setting "style"

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
         }
     }
 
+    @EventHandler
+    public void onRender2D(Render2DEvent event) {
+        if (mc.options.hudHidden || style.get() != TracerStyle.Offscreen) return;
+        count = 0;
+
+        for (Entity entity : mc.world.getEntities()) {
+            if (!PlayerUtils.isWithin(entity, maxDist.get()) || (!Modules.get().isActive(Freecam.class) && entity == mc.player) || !entities.get().getBoolean(entity.getType()) || (ignoreFriends.get() && entity instanceof PlayerEntity && Friends.get().isFriend((PlayerEntity) entity)) || (!showInvis.get() && entity.isInvisible()) | !EntityUtils.isInRenderDistance(entity)) continue;
```
MineGame159, line 271: Maybe put this into a method so it can be re-used with the lines mode

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+
+            Color color;
```
MineGame159, line 273: Same with the color here

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+            Vector2f point = new Vector2f(points[i].x, points[i].y);
+            point = subVec2f(point, TRIANGLE_CENTER);
+
+            Vector2f newPoint = new Vector2f(point.x * COS - point.y * SIN, point.x * SIN + point.y * COS);
+
+            points[i] = addVec2f(newPoint, TRIANGLE_CENTER);
+        }
+    }
+
+    private Vector2f VectorAngles(final Vector3d forward) {
```
MineGame159, line 369: Follow proper naming please

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+        float tmp, yaw, pitch;
+
+        if(forward.x == 0 && forward.y == 0) {
+            yaw = 0;
+            if(forward.z > 0)
+                pitch = 270;
+            else
+                pitch = 90;
+        } else {
+            yaw = (float)(Math.atan2(forward.y, forward.x) * 180 / 3.1415f);
```
MineGame159, line 379: You can use the `Math.PI` constant here

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+                TRIANGLE_POINTS[2].y, finalColor);
+            Renderer2D.COLOR.render(null);
+
+            count++;
+
+            RenderSystem.getModelViewStack().pop();
+        }
+    }
+
+    private Vector2f addVec2f(Vector2f a, Vector2f b) {
```
MineGame159, line 342: `Vector2f` already contains these methods, either `a.add(b, new Vector2f())` to create a new one or if you omit the second parameter then it will store the result inside `a` for better performance

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+                    case AMBIENT -> ambientColor.get();
+                    default -> miscColor.get();
+                };
+            }
+
+            Color finalColor = new Color(color.r, color.g, color.b, color.a);
+            if (blinkOffscreen.get())
+                finalColor.a *= getAlpha();
+
+            final Vec2f SCREEN_CENTER = new Vec2f(mc.getWindow().getFramebufferWidth() / 2.f, mc.getWindow().getFramebufferHeight() / 2.f);
```
MineGame159, line 298: Naming convention

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+                color = switch (entity.getType().getSpawnGroup()) {
+                    case CREATURE -> animalsColor.get();
+                    case WATER_AMBIENT, WATER_CREATURE, UNDERGROUND_WATER_CREATURE, AXOLOTLS -> waterAnimalsColor.get();
+                    case MONSTER -> monstersColor.get();
+                    case AMBIENT -> ambientColor.get();
+                    default -> miscColor.get();
+                };
+            }
+
+            Color finalColor = new Color(color.r, color.g, color.b, color.a);
```
MineGame159, line 294: You can just modify the existing color object, no need to create a new one

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+    private Vector2f subVec2f(Vector2f a, Vector2f b) {
+        return new Vector2f(a.x - b.x, a.y - b.y);
+    }
+
+    private Vector2f divVec2f(Vector2f src, float fac) {
+        return new Vector2f(src.x / fac, src.y / fac);
+    }
+
+    private void rotateTriangle(Vector2f[] points, float ang) {
+        final Vector2f TRIANGLE_CENTER = divVec2f(addVec2f(addVec2f(points[0], points[1]), points[2]), 3);
```
MineGame159, line 355: Naming conventions

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+
+            projection = new Vector3d(entity.prevX, entity.prevY, entity.prevZ);
+            NametagUtils.to2D(projection, 1, false, true);
+
+            Vector2f angle = VectorAngles(new Vector3d(SCREEN_CENTER.x - projection.x, SCREEN_CENTER.y - projection.y, 0));
+            angle.y += 180;
+
+            MatrixStack matrices = RenderSystem.getModelViewStack();
+            matrices.push();
+            matrices.translate(0, 0, 0);
```
MineGame159, line 314: You don't need the translate here, you are moving the matrix by 0

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -192,6 +262,141 @@ else if (entity instanceof PlayerEntity) {
...
+            Vector2f TRIANGLE_POINTS[] = {
+                new Vector2f(NEW_POINT.x - sizeOffscreen.get(), NEW_POINT.y - sizeOffscreen.get()),
+                new Vector2f(NEW_POINT.x + sizeOffscreen.get() * 0.73205f, NEW_POINT.y),
+                new Vector2f(NEW_POINT.x - sizeOffscreen.get(), NEW_POINT.y + sizeOffscreen.get())
+            };
+
+            // rotate
+            rotateTriangle(TRIANGLE_POINTS, angle.y);
+
+            Renderer2D.COLOR.begin();
```
MineGame159, line 331: You can call `begin()` before the whole loop and `render(null)` after it to increase performance. Along with that you will also need to move the matrix push and pop calls outside the loop.

src/main/java/meteordevelopment/meteorclient/utils/render/NametagUtils.java
```diff
@@ -52,15 +56,22 @@ public static boolean to2D(Vector3d pos, double scale, boolean distanceScaling)
         vec4.mul(model, mmMat4);
         mmMat4.mul(projection, pmMat4);
 
-        if (pmMat4.w <= 0.0f) return false;
+        final boolean BEHIND = pmMat4.w <= 0.f;
```
MineGame159, line 59: Naming again

MineGame159 (commented):
Also can you please use the standard formatting everywhere?
 - For example sometimes you do `if(` and sometimes you do `if (`. The latter is what we use.
 - In Java using `final` for parameters and local variables basically does nothing apart from the error checker. (you also sometimes use it and sometimes not)
 
And there are also multiple places where you can re-use the vector objects instead of creating a new one every time but I didn't feel like writing a comment every time.

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
vaxerski: unfortunately doesn't seem to be the case as color becomes a reference and editing it changes the setting itself.

vaxerski:
I think I've addressed all the comments, the color has to be new because otherwise it's a reference and we edit it.

PR #3564, opened 2023-04-09
Fix .vclip not working in downward direction

DesiCow:
## Type of change

- [X] Bug fix

## Description

Fixed vclip lagging back the player when teleporting >10 blocks downward on Paper servers.

The problem was caused by incorrect calculation of number of onGround move packets required, where `Math.ceil(blocks/10)` would return a `negative` value, resulting in skipped onGround move packets. This fix involves taking the `absolute value` of the steps before finding the ceil, ensuring that the correct number of onGround move packets are sent when teleporting downward.

## Related issues

N/A

# How Has This Been Tested?

Tested on dedicated Paper server 🙏 

# Checklist:


Saturn5Vfive:
so real... meteor best client

7n7o:
I was just about to make a pr for this exact issue ahaha

PR #3646, opened 2023-05-09
reverting sodium 0.4.11 too 0.4.10 to fix sodiums broken ass api thing

master7720:


master7720:
closing this cause @RacoonDog has a better solution.

PR #3663, opened 2023-05-15
EntityTypeListSetting Patch

RacoonDog:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Make `EntityTypeListSetting` use a fastutil `ObjectOpenHashSet` instead of a fastutil `Object2BooleanOpenHashMap`

Benefits:
- Lower memory usage (No need for the map's value array)
- HashMap value was always `true`
- Simpler to use, before modules used both `entityTypeList.get().containsKey()` and `entityTypeList.get().getBoolean()` for the same usage, it has been changed to `entityTypeList.get().contains()`. Looping is now `for (var type : entityTypeList.get())` instead of `for (var type : entityTypeList.get().entrySet())`

# How Has This Been Tested?

workey no crashey

# Checklist:


BDARmech:
I kinda want to be your bitch.

PR #3664, opened 2023-05-15
InputCommand Memory Leak Fix

RacoonDog:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

`KeypressHandler` objects were not getting gc'd, resulting in a slight memory leak with every usage.

# How Has This Been Tested?

Spamming the command until out of memory freeze, or lack thereof

# Checklist:


BDARmech:
Balls level. +9-9 on my end.

RacoonDog:
Note: A better fix would be to fix the underlying issue with orbit, in which case it'd be better to not merge this pr

PR #3676, opened 2023-05-23
Mixin extras

Wide-Cat:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Replace a broad amount of mixin injection with some of those from [MixinExtras](https://github.com/LlamaLad7/MixinExtras). These have the benefit that they shouldn't crash due to redirect conflicts with another mod, they should instead chain. I replaced most injections where it was sensible to do. There are a few more instances where I could have replaced an `@Inject` annotation in favour of `@ModifyReturnValue`, but chose not to since setting the return value via `@Inject` also cancels the target method. Finally, increases the jar size by ~200kb.

## Related issues

Probably some issues regarding mod compatibility, but I haven't tested against different mods thoroughly enough to confirm which ones.

# How Has This Been Tested?

Development and production on a local paper server. Due to the number of changes and finicky nature of mixins it isn't unlikely I missed something, but I attempted to be thorough.

# Checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,7 @@ dependencies {
     library "org.reflections:reflections:${project.reflections_version}"
     library("io.netty:netty-handler-proxy:${project.netty_version}") { transitive = false }
     library("io.netty:netty-codec-socks:${project.netty_version}") { transitive = false }
-    include(implementation("com.github.LlamaLad7.MixinExtras:mixinextras-fabric:${project.mixin_extras_version}"))
+    library "com.github.LlamaLad7.MixinExtras:mixinextras-fabric:${project.mixin_extras_version}"
     annotationProcessor "com.github.LlamaLad7.MixinExtras:mixinextras-fabric:${project.mixin_extras_version}"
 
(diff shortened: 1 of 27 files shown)


build.gradle
```diff
@@ -68,7 +68,7 @@ dependencies {
     library "org.reflections:reflections:${project.reflections_version}"
     library("io.netty:netty-handler-proxy:${project.netty_version}") { transitive = false }
     library("io.netty:netty-codec-socks:${project.netty_version}") { transitive = false }
-    include(implementation("com.github.LlamaLad7.MixinExtras:mixinextras-fabric:${project.mixin_extras_version}"))
+    library "com.github.LlamaLad7.MixinExtras:mixinextras-fabric:${project.mixin_extras_version}"
```
Soda5601, line 71: Use ModInclude instead of library is better I think

Wide-Cat: Minegame recommended to use library, to keep consistencey with the rest of them. Is modInclude different or better to library?

PR #3678, opened 2023-05-23
Verbose addon crash

RacoonDog:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Makes the `java.lang.NoClassDefFoundError: meteordevelopment/meteorclient/systems/commands/Command` crash reason point to the addon causing it.

# How Has This Been Tested?

:trollswagcat:

# Checklist:


BDARmech (approved):
nice end.

MineGame159:
You should check if it message of the exception contains the word `meteorclient` (I don't think you can get the missing Class from it) since this might not always be the reason for that exception.

PR #3722, opened 2023-06-11
Always return response body, not only with an HTTP code of 200

DAMcraft:
## Description

In the old version the response body was for some reason only returned, if the status code is 200. If an API (like mine) uses special status codes and gives a detailed error in the response, you cannot fetch it

# Checklist:


Paddyk45:
This kinda breaks the EasyMC screen because it uses this to verify if a token is valid or not, not a big deal though

MineGame159:
Wdym not a big deal

DAMcraft:
Yeah whatever, I use my own small requests thingy now

arlomcwalter:
what, all you have to do is actually handle the status codes

PR #3778, opened 2023-06-25
Build Optimizations

Fiz-Vector:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Use modInclude instead of shadow for compatibility and jar size
Launcher now depends on Java 17

## Related issues

N/A

# How Has This Been Tested?

Successfully launching Minecraft in the launcher.

# Checklist:


MineGame159:
The launcher is using Java 8 for a reason. It is because people that are not tech savvy dont just install new Java versions as they are released but they most probably have Java 8 since it was so popular. We were getting a lot of "bug reports" from those people because they couldnt open the jar, the reason was that they had Java 8.

And how is using `modInclude` better for compatibility?

Fiz-Vector:
> And how is using modInclude better for compatibility?

That's cuz I think other mods with some classes will override meteor shadowed jar's class and might break other mods.

build.gradle
```diff
@@ -38,10 +37,6 @@ configurations {
     // include mods
     modImplementation.extendsFrom(modInclude)
     include.extendsFrom(modInclude)
-
-    // include libraries
-    implementation.extendsFrom(library)
```
RacoonDog, line 43: non-minecraft dependencies should use `implementation`, switching them from `library` to `modInclude` uses `modImplementation` and adds a redundant remap step

Fiz-Vector: But If you use implementaion, the dependencies won't be added into the meteor jar

RacoonDog: > But If you use implementaion, the dependencies won't be added into the meteor jar

yes, that’s intended gradle behaviour. `include` is the configuration which adds the dependency into the jar

PR #3799, opened 2023-07-01
Fixed Jesus On Blocks Other Than Liquid Source

etianl:
Jesus was murdering me when I was walking on any lava other than source blocks, by causing the character to dip into it. This smol change fixes that

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Jesus.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Jesus.java
@@ -19,6 +19,7 @@ import meteordevelopment.meteorclient.systems.modules.Module;
 import meteordevelopment.meteorclient.utils.entity.EntityUtils;
 import meteordevelopment.orbit.EventHandler;
 import net.minecraft.block.BlockState;
+import net.minecraft.block.Blocks;
 import net.minecraft.enchantment.ProtectionEnchantment;
 import net.minecraft.entity.effect.StatusEffects;
 import net.minecraft.fluid.Fluids;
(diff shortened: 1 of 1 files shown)


BDARmech:
\no?

etianl:
Yes, if you attempt to use jesus on blocks other than the source no matter what setting you choose it will dip you in over and over. This change fixes that

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Jesus.java
```diff
@@ -215,9 +215,9 @@ else if ((event.fluidState.getFluid() == Fluids.LAVA || event.fluidState.getFlui
     private void onFluidCollisionShape(CollisionShapeEvent event) {
         if (event.state.getFluidState().isEmpty()) return;
 
-        if (event.state.getFluidState().isOf(Fluids.WATER) && !mc.player.isTouchingWater() && waterShouldBeSolid()) {
+        if (event.state.getBlock() == Blocks.WATER && !mc.player.isTouchingWater() && waterShouldBeSolid()) {
```
BDARmech, line 218:
```suggestion
        if (event.state.getBlock() == Blocks.WATER && !mc.player.isTouchingWater() && waterShouldBeSolid()) {
```
Is the fluid tag/function obsolete, will it not be used?

etianl: No idea if it's obsolete but it doesn't work properly for the portions of liquid that are flowing other than the source block

etianl: This is a video showing how it acts without the code change. (Broken)
https://github.com/MeteorDevelopment/meteor-client/assets/115842502/eb3c48ba-9197-4e3b-86db-fe7429483cda
This is how it acts, after that smol change. (Fixed)
https://github.com/MeteorDevelopment/meteor-client/assets/115842502/bbef313e-dd35-428a-82a0-1ada241227c6

etianl:
"maybe just add water and lava in collisions ("
Yeah this is a bypass to make Jesus work, but all the options in Jesus do nothing to change that it's deadly on lava and annoying on water (not the source blocks, just the blocks that flow from them) without the fix I provided. With that smol change to the if statements it behaves how it used to again before the update to 1.20.1

etianl:
I had to fix the codes again because my change was not accounting for the waterlogged blocks as being water blocks

Issue #3788, opened 2023-06-27
Reconnect button reconnects to wrong server


CRD716:
### Describe the bug

When I lose connection to a server and press the Reconnect button, it connects me to the first server in my server list instead of the server I was on.

### Steps to reproduce

```bash
Have multiple servers on your server list
Connect to server on latest dev build
Lose connection through some means (unplug your ethernet if you want)
Attempt to reconnect
Watch as you connect to a server that might not like your killaura kicking in
```

### Meteor Version

0.5.4 Dev

### Minecraft Version

1.20.1

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported (I have searched bug reports on GitHub).

- [X] This is a valid bug (I am able to reproduce this on the latest dev build).

ManInMyVan:
Yeah it's kinda annoying.

RedCarlos26:
I can confirm this

Fiz-Victor:
I created a [pr ](https://github.com/MeteorDevelopment/meteor-client/pull/3438/files#diff-c99db43bdb906d4edcc7fe113d7f88a2ef9ea934f3b010fc249c839c9719ae23)  to fix this but this was closed

Wide-Cat:
should be fixed as of 3e3f2964da06addf50f1b654041a78e7182a8110

PR #3827, opened 2023-07-12
Auto Reconnect fixed the joining the wrong server bug

Fiz-Victor:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Added visibility toggle option
Fix AutoReconnect unable to reconnect to a server if you failed to join

## Related issues

#3788

# How Has This Been Tested?

Joining an nonexist server and it will attempt to reconnect now.

# Checklist:


MineGame159:
1. Why did you rename AutoReconnect to Reconnect?
2. Why did you add an @Unique field for the AutoReconnect instance when you don't even use it and instead get the instance from the module manager anyways?
3. Why didn't you just change from where the even is being fired instead of removing it and hardcoding the mixin?

Fiz-Victor:
> 1. Why did you rename AutoReconnect to Reconnect?
> 4. Why did you add a setting that enables / disables the sole function of the module? Makes zero sense.

I wanted to make the "Reconnect" and "AutoReconnectToggle" button toggleable. (this made meteor more customizable imo)

> 2. Why did you add an @unique field for the AutoReconnect instance when you don't even use it and instead get the instance from the module manager anyways?

I have never used @Unique in my other projects and didnt't know it very well.

> 3. Why didn't you just change from where the even is being fired instead of removing it and hardcoding the mixin?

I forgot runInMainMenu is set to true.

Wide-Cat:
should be resolved as of 3e3f2964da06addf50f1b654041a78e7182a8110

PR #3831, opened 2023-07-13
Added PacketDelay

### LINKED ISSUE #3689: Add packet queue / delay function to packet cancel module
### Describe the feature

What I mean by queueing packets is that instead of the actions you do on your client not being sent to the server at all (like how packet cancel works now), the selected packet types would be saved to a list instead and all sent at the same time when you disable the module.

This function would be useful for hunting dupe exploits, as things like shop plugins don't generally expect a player to send multiple buy orders at the same time.

Here is a mod that already has this feature, but only for guis: https://github.com/Coderx-Gamer/ui-utils

It might also be handy to make this a seperate module from packet cancel, so you can withhold certain packets entirely while queueing others.

thebest9178:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Added PacketDelay.
Lets you delay sending/receiving set packets until the module is disabled or a set time passes since the packet was supposed to be sent/received.

## Related issues

Closes #3689

# How Has This Been Tested?

Messing around in a test world.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
@@ -560,6 +560,7 @@ public class Modules extends System<Modules> {
         add(new Notebot());
         add(new Notifier());
         add(new PacketCanceller());
+        add(new PacketDelay());
         add(new SoundBlocker());
         add(new Spam());
         add(new ServerSpoof());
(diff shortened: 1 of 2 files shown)


BDARmech:
Is there a reason other than nefarious stuff?

tympanicblock61:
> Is there a reason other than nefarious stuff?

why should there be

zyrafaq:
1. The goal of this project is to be customizable.
2. This can be usefull for some exploits or duping

PR #3869, opened 2023-07-22
Add texture rotations to no render

Wide-Cat:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds an option in no render to change texture rotations and model offsets to use a constant value, instead of one dependant on the block postion. This helps protect against accidently leaking your coordinates.

# How Has This Been Tested?

Disabled:

Enabled:

# Checklist:


RacoonDog:
Looks weird, IMO you should just randomize the implementation so it’s different every join

PR #3879, opened 2023-07-24
Added new sneak parameter to the safe walk module

o0sh4d0w0o:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Added a missing feature to the safe walk module : sneak

- **Sneak parameter** : When enabled this will change the default behavior to if player is close to the edge of a block next to an empty one.
- **Safe Sneak** (_Sneak dependent parameter_) : When enabled is will ensure the player to don't fall if sneaking failed
- **Sneak On Sprint** (_Sneak dependent parameter_) : When disabled pressing sprinting key will cancel the safe walk to being effective, as example to doing a jump
- **Edge Distance** (_Sneak dependent parameter_) : A tweak parameter to manage sensibility of sneaking behavior
- **Render** (_Sneak dependent parameter_) : Will render the box of player after edge distance applied
- **Render Player Box** (_Render dependent parameter_) : Draw the original player box without edge distance applied

## Related issues

None

# How Has This Been Tested?

Locally on a minecraft server on the latest version (1.20.1)

scr

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/SafeWalk.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/SafeWalk.java
@@ -6,17 +6,108 @@
 package meteordevelopment.meteorclient.systems.modules.movement;
 
 import meteordevelopment.meteorclient.events.entity.player.ClipAtLedgeEvent;
+import meteordevelopment.meteorclient.events.render.Render3DEvent;
+import meteordevelopment.meteorclient.renderer.ShapeMode;
+import meteordevelopment.meteorclient.settings.*;
 import meteordevelopment.meteorclient.systems.modules.Categories;
 import meteordevelopment.meteorclient.systems.modules.Module;
+import meteordevelopment.meteorclient.utils.render.color.Color;
 import meteordevelopment.orbit.EventHandler;
+import net.minecraft.util.math.Box;
 
 public class SafeWalk extends Module {
+    private final SettingGroup swGeneral = settings.getDefaultGroup();
+
+    private final Setting<Boolean> sneak = swGeneral.add(new BoolSetting.Builder()
+            .name("sneak")
+            .description("Sneak when approaching edge of block.")
+            .defaultValue(false)
+            .build());
+
+    private final Setting<Boolean> safeSneak = swGeneral.add(new BoolSetting.Builder()
+            .name("safe-sneak")
+            .description("Prevent you to falling if sneak doesn't trigger correctly.")
+            .defaultValue(true)
+            .visible(() -> this.sneak.get())
+            .build());
+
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/movement/SafeWalk.java
```diff
@@ -6,17 +6,108 @@
...
+    }
+
+    private Box getAdjustedPlayerBox(Box playerBox) {
+        return playerBox.stretch(0, -mc.player.getStepHeight(), 0)
+                .expand(-edgeDistance.get(), 0, -edgeDistance.get());
+    }
+
+    @EventHandler
+    private void onRender(Render3DEvent event) {
+        if (this.renderEdgeDistance.get()) {
```
Wide-Cat, line 101: make this check for the sneak setting being active as well as the render one, otherwise if you disable sneak it will still render

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/SafeWalk.java
```diff
@@ -6,17 +6,108 @@
...
+            if (mc.world.isSpaceEmpty(mc.player, adjustedBox) && mc.player.isOnGround())
+                closeToEdge = true;
+
+            if (!isSprinting) {
+                if (closeToEdge == true) {
+                    mc.player.input.sneaking = true;
+                } else if (this.safeSneak.get()) {
+                    event.setClip(true);
+                }
+            }
```
Wide-Cat, line 86: Right now, if you disable sneakSprint and sprint to the edge of a block, safeSneak will never come into effect. Is that intentional?

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/SafeWalk.java
```diff
@@ -6,17 +6,108 @@
...
+            .description("Render edge distance helper.")
+            .defaultValue(false)
+            .visible(() -> this.sneak.get())
+            .build());
+
+    private final Setting<Boolean> renderPlayerBox = swGeneral.add(new BoolSetting.Builder()
+            .name("render-player-box")
+            .description("Render player box helper.")
+            .defaultValue(true)
+            .visible(() -> this.renderEdgeDistance.get())
```
Wide-Cat, line 61: make this check for the sneak setting as well as the render setting

Wide-Cat (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/SafeWalk.java
o0sh4d0w0o: Yeah, it's intentionnal behavior, sneakSprint prevent user to remove sneak only while the option is enabled, once is disabled,  sneak will be cancelled, safeSneak included.

PR #3968, opened 2023-08-19
Liquid filler fixes2

### LINKED ISSUE #709: [Bug] Liquid filler ghost blocks
when filling water with liquid filler it often places ghost ghost blocks above you in positions that seem related to the position of the water
**To Reproduce**
1. place a bunch of water
2. toggle liquid filler
3. move in the water

thebest9178:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

Fixed #709.
Made it possible to place more than one block at a time.
Set the default delay to 0.
Added max blocks per tick.
Added Shape and SortMode
Added ListMode for white and black lists

## Related issues

#709

# How Has This Been Tested?

By filling a river and small lava lake in creative with various settings.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/LiquidFiller.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/LiquidFiller.java
@@ -17,7 +17,6 @@ import meteordevelopment.meteorclient.utils.world.BlockUtils;
 import meteordevelopment.orbit.EventHandler;
 import net.minecraft.block.Block;
 import net.minecraft.block.Blocks;
-import net.minecraft.block.ShapeContext;
 import net.minecraft.fluid.Fluid;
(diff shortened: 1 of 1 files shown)


Wide-Cat:
In this image, it's not placing any blocks, although it should be able to. My guess would be it's something to do with waterlogged blocks, because the same issue also appeared with a fence post under the water. 
I also still noticed a bit of weirdness with it not placing properly on sphere mode with the default range, but it's much less of an issue now, and I'm unsure what the root of that problem was.

thebest9178:
Fixed the waterlogged blocks issue. It won't try to replace things it can't like kelp, but will replace stuff like sea grass.

I don't see any issues placing with default settings.
You mention it being less of an issue placing with sphere mode _now_, but before I added shapemode it only used a cube shape so there was no sphere mode before.

PR #3975, opened 2023-08-20
Offhand Rewrite

CrytoPal:
## Type of change

- [x] New feature

## Description

Rewrote Offhand which adds Right Click Gapple, Potion Offhand, and Auto Totem in one module

# How Has This Been Tested?

Tested it in AC testing servers + CC (Crystalpvp.cc)

# Videos or screenshots of the changes if applicable.

You can try out the module yourself (I was too lazy to record)

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/Offhand.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/Offhand.java
@@ -53,7 +53,7 @@ public class Offhand extends Module {
     private final Setting<Boolean> rightgapple = sgCombat.add(new BoolSetting.Builder()
         .name("right-gapple")
         .description("Will switch to a gapple when holding right click.(DO NOT USE WITH POTION ON)")
-        .defaultValue(true)
+        .defaultValue(false)
         .build()
     );
 
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/Offhand.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/Offhand.java
@@ -30,15 +30,15 @@ public class Offhand extends Module {
     //Combat
 
     private final Setting<Integer> delayTicks = sgCombat.add(new IntSetting.Builder()
-        .name("item-switch-delay")
+        .name("Item Switch Delay")
         .description("The delay in ticks between slot movements.")
-        .defaultValue(1)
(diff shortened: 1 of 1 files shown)


CrytoPal:
Made new branches to make commits look cleaner

Wide-Cat:
- sword gapple has a wrong description
- add an option to have 0 ticks of delay
- if you have a crystal in your offhand and right click to eat a gapple, it will place a crystal on obsidian. Maybe add an option to do something about this, like always holding a gapple when holding a sword
- change it back to just 'offhand', not 'Off Hand'

CrytoPal:
@Wide-Cat 

> * sword gapple has a wrong description
> 
>     * add an option to have 0 ticks of delay
> 
>     * if you have a crystal in your offhand and right click to eat a gapple, it will place a crystal on obsidian. Maybe add an option to do something about this, like always holding a gapple when holding a sword
> 
>     * change it back to just 'offhand', not 'Off Hand'

Fixed the issues stated here

Wide-Cat:
- Sword gapple doesn't seem to work anymore. I was holding a sword and right clicking, but it wasn't transferring a gap to my offhand.
- There seems to be a bug with the delay setting. I couldn't find a method to reproduce it, but if you set the delay to a high value, there are definitely some times where it swaps quicker than it should.
- I'm a little conflicted on the potion setting, it seems like a waste of your offhand slot to me. Maybe remove it from the item enum and instead add an option to hold a strength potion in your offhand if you're holding a sword, you're in a hole and don't have strength already, although I feel like even that could have its flaws and you'd probably just be better off using quiver

CrytoPal:
I

> * Sword gapple doesn't seem to work anymore. I was holding a sword and right clicking, but it wasn't transferring a gap to my offhand.
> 
>     * There seems to be a bug with the delay setting. I couldn't find a method to reproduce it, but if you set the delay to a high value, there are definitely some times where it swaps quicker than it should.
> 
>     * I'm a little conflicted on the potion setting, it seems like a waste of your offhand slot to me. Maybe remove it from the item enum and instead add an option to hold a strength potion in your offhand if you're holding a sword, you're in a hole and don't have strength already, although I feel like even that could have its flaws and you'd probably just be better off using quiver

I Fixed the Sword Gap and added more Potion delays. But I couldn't manage to reproduce the bug with the delay.

PR #3978, opened 2023-08-21
Fixes

SpecKeef:
## Just little improvements

- Added some consistent variables

## Description

 Added some consistent variables for consistency

# How Has This Been Tested?

No visual changes.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/EntityOwner.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/EntityOwner.java
@@ -38,26 +38,30 @@ public class EntityOwner extends Module {
 
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
 
-    private final Setting<Double> scale = sgGeneral.add(new DoubleSetting.Builder()
-        .name("scale")
-        .description("The scale of the text.")
-        .defaultValue(1)
-        .min(0)
-        .build()
-    );
+    private final Setting<Double> scale;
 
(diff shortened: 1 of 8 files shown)


RacoonDog:
`MeteorCleint.NAME` already exists

SpecKeef:
> `MeteorCleint.NAME` already exists

yeah lol i just realized.

Big-Iron-Cheems:
I have no idea why you retouch so much code unnecessarily...

- You added `else` to `if` that already end with a return for no reason.
- You then mangled the settings syntax in a way that is inconsistent from the codebase.
- You also proceed to put asserts that are unnecessary, and add parentheses to single line statements.
- And finally you inverted guard statements, reducing readability and adding more indents.

Overall, I have no idea what the point of all this is, and how is it related with the original task of this PR.

src/main/java/meteordevelopment/meteorclient/systems/modules/render/EntityOwner.java
```diff
@@ -38,26 +38,30 @@ public class EntityOwner extends Module {
...
+            .min(0)
+            .build()
+        );
+
+        projectiles = sgGeneral.add(new BoolSetting.Builder()
+            .name("projectiles")
+            .description("Display owner names of projectiles.")
+            .defaultValue(false)
+            .build()
+        );
```
Fiz-Victor, line 64: why init settings in instantiation

Fiz-Victor (changes requested):

PR #3980, opened 2023-08-21
Added Recast mode to Elytra Fly

supakeks:
## Type of change

- [ ] Bug fix
- [✓] New feature

## Description

I added a "Recast" mode to the ElytraFly module. Recast mode modifies the vanilla behaviour where your elytra closes when you hit the ground. This mode makes it so the elytra remains open when hitting the ground, allowing you to keep bouncing on the ground and go at over 40 blocks per second. Because of the high speed it can reach, it is very normal that you will get rubberbanded, which is why I added a setting to it (which is on by default) that automatically relaunches your elytra when you get rubberbanded.

## Related issues

There are no issues regarding this as this mode was made public just a few days ago.

# How Has This Been Tested?

This was tested on 2b2t.org, which uses Grim AC. It rubberbands every 4/5 seconds, but the "restart" setting makes it so it instantly restarts your elytra. This was also tested on a server with the NCP anti-cheat, as well as a single-player world.

# Checklist:

- [✓] My code follows the style guidelines of this project.
- [✓] I have added comments to my code in more complex areas.
- [✓] I have tested the code in both development and production environments.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/modes/Recast.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/modes/Recast.java
@@ -3,6 +3,10 @@
  * Copyright (c) Meteor Development.
  */
 
+/*
+ * Credit to Luna (https://github.com/InLieuOfLuna) for making the original Elytra Recast mod (https://github.com/InLieuOfLuna/elytra-recast)!
+ */
+
 package meteordevelopment.meteorclient.systems.modules.movement.elytrafly.modes;
 
 import meteordevelopment.meteorclient.events.packets.PacketEvent;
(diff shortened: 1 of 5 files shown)


Wide-Cat:
Can you move the methods that are only used in or relevant for Recast to that specific modes file, not in the primary file

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/modes/Recast.java
```diff
@@ -0,0 +1,114 @@
...
+    int tickDelay = elytraFly.restartDelay.get();
+
+    @Override
+    public void onTick() {
+        super.onTick();
+
+        // Make sure all the conditions are met (player has an elytra, isn't in water, etc)
+        if (checkConditions(mc.player)) {
+
+            if (elytraFly.flightMode.get() == ElytraFlightModes.Recast) {
```
Wide-Cat, line 44: in which situation would the elytra mode not be recast?

Wide-Cat:
also can you add about 100 blocks to each of the default height settings to pitch40 mode

PR #3989, opened 2023-08-22
Bloat removal

CrytoPal:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Removes a bunch of useless modules or combines them into similar modules

## Related issues

nothing

# How Has This Been Tested?

Multiplayer and Survival World

# Checklist:


CrytoPal:
Most of them were bloat

MineGame159:
Bruh

Paddyk45:
> Most of them were bloat and don't work

Then fix them instead of removing them??

CrytoPal:
> > Most of them were bloat and don't work
> 
> Then fix them instead of removing them??

1. How do you fix a patched crash exploit
2. Most of those don't serve a useful purpose in Anarchy especially when we're now competing with other clients that are starting to update to newer versions of Minecraft

PR #4002, opened 2023-08-25
AirPlace

### LINKED ISSUE #2419: [Bug] Fast Use Issue
### Describe the bug

Fast use on "Some" mode without blocks selected will still fast place blocks while using the airplace module

### Steps to reproduce

```bash
Set fast use to "some" mode and enable the airplace module, observe fast use negating the break delay without blocks being enabled
```

### Meteor Version

1505

### Minecraft Version

1.19

### Operating System

Windows

- [X] This is a valid bug.

thebest9178:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

Used to try to place every tick. Now the timing is not influenced by AirPlace.
Now works with both hands.

## Related issues

Closes #2419

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AirPlace.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AirPlace.java
@@ -5,7 +5,7 @@
 
 package meteordevelopment.meteorclient.systems.modules.world;
 
-import net.minecraft.item.SpawnEggItem;
+import meteordevelopment.meteorclient.events.entity.player.InteractItemEvent;
 import meteordevelopment.meteorclient.events.render.Render3DEvent;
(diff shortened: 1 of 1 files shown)


thebest9178:
WIP Because when both hands have valid items and the main hand has an egg, the offhand tries to place at the same time as the main hand when only the main hand should place.

With two eggs, both mobs will be spawned.
With a block in the offhand, the mob will be spawned and the block will show up for one tick.
In both cases the offhand usually swings but not always, also only one hand swings at a time.

Done with hitting my head on the keyboard today. If someone else figures it out that would be nice.

moritz-t-w:
> - [x]  Bug fix
- [x] New feature

Read before you tick

thebest9178:
>Now works with both hands.
It wasn't a bug that it didn't before.

PR #4010, opened 2023-08-28
Improved ElytraFly's Recast mode

supakeks:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

I made it so it would only sprint when touching the ground.

## Related issues

None

# How Has This Been Tested?

On 2b2t.org.

# Checklist:

- [x] My code follows the style guidelines of this project.
- [x] I have added comments to my code in more complex areas.
- [x] I have tested the code in both development and production environments.

PR #4027, opened 2023-08-31
Add speed limit to auto-jump in Elytra recast mode

chairclr:
## Type of change

- [x] New feature

## Description

Adds a limit in ElytraFly recast to only auto-jump when below this limit. This allows the user to limit the maximum speed they travel using ElytraFly recast mode. Traveling at too high of speeds can cause the player to stop and lose all momentum, greatly slowing travel.

## Related issues

I do not believe there are any issues relating to this currently.

# How Has This Been Tested?

https://cdn.discordapp.com/attachments/1117615804238798978/1146876163776594021/javaw_1693506900.mp4
(Uploaded to external source due to large video size)

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/ElytraFly.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/ElytraFly.java
@@ -202,6 +202,16 @@ public class ElytraFly extends Module {
         .build()
     );
 
+    public final Setting<Double> autoJumpSpeedLimit = sgGeneral.add(new DoubleSetting.Builder()
+        .name("auto-jump-speed-limit")
+        .description("Only automatically jumps when below a certain speed.")
+        .defaultValue(27)
+        .min(0)
+        .sliderMax(50)
+        .visible(() -> flightMode.get() == ElytraFlightModes.Recast && autoJump.get())
+        .build()
+    );
+
     public final Setting<Double> pitch = sgGeneral.add(new DoubleSetting.Builder()
         .name("pitch")
         .description("The pitch angle to look at when using the recast mode.")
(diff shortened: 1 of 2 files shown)


supakeks:
What server are you trying it on? The rubberbanding happens because it sprints ALL the time, regardless of the speed. I made a PR (#4010) for that where it only jumps when the player is on the ground, and it stopped rubberbanding.

chairclr:
> What server are you trying it on?
I only got rubberbanding/stopped when traveling above 28m/s on 2b2t.org. Adding a speed limit fixed it.

supakeks:
> > What server are you trying it on?
> 
> I only got rubberbanding/stopped when traveling above 28m/s on 2b2t.org. Adding a speed limit fixed it. I could also manually jump at a slower rate, limiting my speed to 28m/s, which also fixed it. I only ever experienced rubberbanding above 28m/s.

Yeah, my PR fixes that and it lets you go 40 bps without rubberbanding. The issue is with the sprinting, not the speed. If you only sprint when the player is on the ground (mc.player.setSprinting(mc.player.isOnGround()), it stops the rubberbanding.

chairclr:
> Yeah, my PR fixes that and it lets you go 40 bps on 2b without rubberbanding. The issue is with the sprinting, not the speed. If you only sprint when the player is on the ground (mc.player.setSprinting(mc.player.isOnGround()), it stops the rubberbanding.

Interesting, I'll give it a try and see if I can reproduce this.

chairclr:
I cloned your fork and used it, but could not go anywhere close to 40m/s. It only ever got up to around 20-22m/s, and then stopped increasing in speed. Although there was no rubberbanding, it was significantly slower. If I enabled your always sprinting setting, it would go back to higher speeds, but eventually rubberband at around 30m/s.

supakeks:
uhhh that's odd, that hasn't happened to me.

chairclr:
After some tweaking, I found that only sprinting when in the air works perfectly.
`mc.player.setSprinting(!mc.player.isOnGround());`

chairclr:
I'll close this PR and go suggest those changes on yours.

supakeks:
I'm trying that out and it doesn't sprint like over half of the time, weird. Do you have discord? (easier to talk there instead of here)

PR #4043, opened 2023-09-05
Added always show to BetterTab

Dellism:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

adds feature to bettertab to always show tablist whether or not tab is pressed

## Related issues

# How Has This Been Tested?

local server, public server

# Checklist:


Wide-Cat:
what would the use of this be? I fail to see a situation where this would be desireable

PR #4057, opened 2023-09-08
Add a CPS variable to Meteor Starscript

panpanpro888:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Added a CPS variable to the Meteor Starscript that counts the player's clicks per second.

## Related issues

None

# How Has This Been Tested?

As just a Starscript variable, the only form of testing that i can do is open up the game and see if it counts correctly, which it does.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/Utils.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/Utils.java
@@ -73,6 +73,10 @@ public class Utils {
     public static Screen screenToOpen;
     public static VertexSorter vertexSorter;
 
+    public static int cps;
+    private static long lastTime;
+    public static ArrayList<Integer> clickList = new ArrayList<>();
+
     public static final Pattern FILE_NAME_INVALID_CHARS_PATTERN = Pattern.compile("[\\s\\\\/:*?\"<>|]");
 
     @PreInit
@@ -85,6 +89,28 @@ public class Utils {
         if (screenToOpen != null && mc.currentScreen == null) {
             mc.setScreen(screenToOpen);
             screenToOpen = null;
+        }
+    }
+
+    public static void cpsChecker() {
+        long currentTime = System.currentTimeMillis();
+        // Run every second
+        if (currentTime - Utils.lastTime >= 1000) {
+            // If the player did not click in that second
+            // empty the cps list
+            if (Utils.cps == 0) {
+                clickList.clear();
+            } else {
+                // Add the previous cps to the list
+                // and reset the CPS and time
+                clickList.add(clickList.isEmpty() ? 0 : clickList.get(clickList.size() - 1));
+                Utils.cps = 0;
+                Utils.lastTime = currentTime;
+            }
+        }
+        if (!clickList.isEmpty()) {
+            // Set the current CPS to the last element of the list
+            clickList.set(clickList.size() - 1, Utils.cps);
         }
     }
 
--- a/src/main/java/meteordevelopment/meteorclient/utils/Utils.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/Utils.java
@@ -72,11 +72,6 @@ public class Utils {
     public static double frameTime;
     public static Screen screenToOpen;
     public static VertexSorter vertexSorter;
-
-    public static int cps;
-    private static long lastTime;
-    public static ArrayList<Integer> clickList = new ArrayList<>();
-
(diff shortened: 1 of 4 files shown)


Fiz-Victor:
This can be done by a cps checker thread?

MineGame159:
> This can be done by a cps checker thread?

Creating a new thread is absolutely not necessary and wasteful.

And to reply to the original PR. It looks like you forgot to commit some changes since I dont see a way for the `cps` to ever change. Also I have some things to change. It would be best to move this to a separate class, use our static initialization system (look at some other static class) and subscribe to the tick event (possible more events) there. Then there should be a method to get the average CPS instead of doing that manually where the CPS is needed and exposing implementation details. Another thing is that using a list of integer objects is very wasteful since Java doesnt have a proper support for generics, it would be better if you just have a normal int array with the size of 20 (since 20 ticks = one second) and just cycle through it.

panpanpro888:
Sorry for the late reply.
I will look into it and do the refactors you said. I will get to working on the commit right now.

panpanpro888:
Hi, I am working on it, and while transferring the events from the Minecraft Client Mixin to its own class with event handlers I am having trouble finding an event with the equivalent of the minecraft client's `doAttack` function to listen for when the player clicks. Is it okay to add a function to increment the cps directly into the mixin? Or if there is actually an event for that can you please tell it to me?

MineGame159:
There is an `AttackEntityEvent` but that will also count attacks made through modules like Kill Aura. If it shouldnt count those (which in my opinion it shouldnt) then its fine to just directly call the static method from the mixin.

panpanpro888:
I can't understand how to make an int array with the size of 20 since the original code had a new array element for every second and then got the average of those elements. Can you please help with this, by the way all the other things are finished.

MineGame159:
Ah well I couldnt really make out how the list was supposed to work since a) the `cps` was never incremented and b) the list was never cleared / removed from so it would just fill up your entire memory after a while.
Basically what I meant is that lets say you want the average of the last 5 seconds. Then you would create an array of size 5 and fill it up in circles (0, 1, 2, 3, 4 and then back to 0, etc). This is to not create a billion Integer objects and stress the GC.

panpanpro888:
I fixed the previous issues and instead of creating an array of integers, I made it store the total clicks and the seconds the players has been clicking. Already committed and pushed the fixes, code is ready.

Fiz-Victor:
now cps will be affected by client tick rate?

panpanpro888:
The only thing done in the `onTick` event is checking if a second has passed. This was also done in the previous commit but directly from the minecraft client mixin.

src/main/java/meteordevelopment/meteorclient/utils/misc/CPSUtils.java
```diff
@@ -0,0 +1,58 @@
...
+        CPSUtils.cps++;
+    }
+
+    public static Value getCpsAverage() {
+        return Value.number((int)(
+            clicks / (secondsClicking == 0 ? 1 : (float)secondsClicking)
+        ));
+    }
+
+    public static void cpsChecker() {
```
MineGame159, line 43: This doesnt need to be a method, and a public at that. You can just move the body to the `onTick` handler.

src/main/java/meteordevelopment/meteorclient/utils/misc/CPSUtils.java
```diff
@@ -0,0 +1,58 @@
...
+        CPSUtils.cpsChecker();
+    }
+
+
+    public static void onAttack() {
+        CPSUtils.clicks++;
+        CPSUtils.cps++;
+    }
+
+    public static Value getCpsAverage() {
```
MineGame159, line 37: This should return an `int` in case something also wants to get the value.

src/main/java/meteordevelopment/meteorclient/utils/Utils.java
```diff
@@ -72,7 +72,6 @@ public class Utils {
     public static double frameTime;
     public static Screen screenToOpen;
     public static VertexSorter vertexSorter;
-
```
MineGame159, line 75: Why. If you want to cleanup then move the 2 constants to the top of the file and with a new line to separate constants and static variables.

MineGame159 (changes requested):
Otherwise looks good.

MineGame159 (approved):

PR #4058, opened 2023-09-10
Added anti-mine module

paredes00:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds a module that lets you prevent mining specific blocks. 
Possible use cases include:
- Prevent accidentally breaking budding amethyst blocks while harvesting clusters
- Prevent accidentally breaking monster spawners
- Prevent mining ores without fortune
- Prevent mining ender chests without silk touch

Can also use on whitelist mode to only allow breaking selected blocks. 

Includes default-enabled option to notify in chat when the module stopped you breaking a block since I figured it may be confusing if you forget it was enabled.

## Related issues

None

# How Has This Been Tested?

Tested both on singleplayer and multiplayer servers, didn't encounter any problems.

image

https://github.com/MeteorDevelopment/meteor-client/assets/120606980/8fea7aa6-bc80-4499-9990-130a7a7060ec

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
@@ -541,6 +541,7 @@ public class Modules extends System<Modules> {
         add(new Timer());
         add(new VeinMiner());
         add(new HighwayBuilder());
+        add(new AntiMine());
     }
 
     private void initMisc() {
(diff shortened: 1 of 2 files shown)


Wide-Cat:
Pretty sure this is already a feature of the no interact module

paredes00:
Ahh, didn't know about that module, feel free to close then

PR #4059, opened 2023-09-10
Shutdown command

OverFloyd:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

A simple command that shuts down the Minecraft client using vanilla's own function.

## Related issues

Works fine, only one tiny thing is that the info message doesn't get displayed on time but whatever.

# Checklist:


Paddyk45:
.disconnect is already .kick in meteor rejects and why can‘t you just press the close button

MineGame159:
There is this X button in the top right corner of the window or depending on your keyboard shortcuts you can also use one of those too.

PR #4083, opened 2023-09-18
full damage utils refactor

### LINKED ISSUE #2455: [Suggestion] Killaura check weapon enchants
### Describe the feature

If killaura is in sword mode and targeting only single entity check what weapon would deal most damage. For an example you have diamond sword with smite 5 and diamond sword with sharpness 5 and killaura is targeting a zombie it should switch to smite 5 diamond sword as it deals more damage to the darget.

### LINKED ISSUE #4056: CrystalAura calculation
### Describe the feature

Now crystal, the faster you run, the more fps drops

I don't know what the fps problem is, the game is 20 ticks, it makes more unnecessary calculations in this time period (I think about this) and as a result, it reduces unnecessary fps And it doesn't happen when you put it slow

And the more players around me, the more calculations they make and the fps decreases as a result

Maybe maxTargets 1-10 can be made for CrystalAura and redundant calculation correction

RacoonDog:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Made damage calculations faster, with better precision

### Performance Improvements
- Simplified explosion raycasts; Less laggy Crystal Aura, Anchor Aura, Bed Aura, Step w/ safestep', AutoTotem smart mode, & Offhand
- Removed duplicate operations in Anchor Aura, Step w/ safestep, AutoTotem smart mode, & Offhand

### Precision Improvements
- All damage calculations now take into account armor toughness
- Attack damage calculation now works with any weapon, modded ones included
- Attack damage calculation now supports attack damage increasing/decreasing modded armors
- Attack damage calculation now properly calculates attack charge & critical hits
- Attack damage calculation now properly calculates mob damage difficulty modifiers
- Bed damage calculation now properly calculates armor damage reductions

### For developers
- `DamageEvent#amount`: Removed
- `DamageUtils#getSwordDamage(PlayerEntity, boolean)`: Removed, use `DamageUtils#getAttackDamage(LivingEntity, LivingEntity)` instead
- `EntityAttributeManager`: New util class to use instead of `LivingEntity#getAttributeInstance(EntityAttribute)`, `LivingEntity#getAttributeValue(EntityAttribute)`, & `LivingEntity#getAttributes()`
- `StatusEffectManager`: New util class to use instead of `LivingEntity#getStatusEffect(StatusEffect)`, `LivingEntity#hasStatusEffect(StatusEffect)`, `LivingEntity#getStatusEffect()`, & `LivingEntity#getActiveStatusEffects()`
- `DamageUtils#fallDamage(LivingEntity entity)`: New util method to calculate fall damage for a given entity

## Related issues

closes #4056, closes #2455

# Checklist:


src/main/java/meteordevelopment/meteorclient/utils/player/DamageUtils.java
```diff
@@ -180,95 +179,107 @@ private static double getDamageForDifficulty(double damage) {
...
-
-            double d = blockHitResult == null ? Double.MAX_VALUE : raycastContext.getStart().squaredDistanceTo(blockHitResult.getPos());
-            double e = blockHitResult2 == null ? Double.MAX_VALUE : raycastContext.getStart().squaredDistanceTo(blockHitResult2.getPos());
-
-            return d <= e ? blockHitResult : blockHitResult2;
-        }, (raycastContext) -> {
-            Vec3d vec3d = raycastContext.getStart().subtract(raycastContext.getEnd());
-            return BlockHitResult.createMissed(raycastContext.getEnd(), Direction.getFacing(vec3d.x, vec3d.y, vec3d.z), BlockPos.ofFloored(raycastContext.getEnd()));
-        });
+            BlockHitResult hitResult = blockState.getOutlineShape(mc.world, blockPos).raycast(start, end, blockPos);
```
tyrannus00, line 281: blockState.getCollisionShape needs be used here instead of getOutlineShape, because otherwise blocks like signs, which have an outline shape but no collision shape, will result in false negatives when ignore terrain is turned off

RacoonDog: good catch!

Funtimes909:
this looks great

src/main/java/meteordevelopment/meteorclient/events/entity/DamageEvent.java
```diff
@@ -13,12 +13,12 @@ public class DamageEvent {
 
     public LivingEntity entity;
     public DamageSource source;
-    public float amount;
+    @Deprecated(forRemoval = true)
+    public final float amount = 0;
```
Fiz-Victor, line 17: just remove this so addons using this variable will crash

Fiz-Victor (changes requested):

src/main/java/meteordevelopment/meteorclient/events/entity/DamageEvent.java
RacoonDog: i was planning on keeping this for a major version change, but 1.20.2 came out so probably

src/main/java/meteordevelopment/meteorclient/utils/player/DamageUtils.java
```diff
@@ -146,31 +137,34 @@ public static double bedDamage(LivingEntity player, Vec3d bed) {
...
 
     // Anchor damage
 
     public static double anchorDamage(LivingEntity player, Vec3d anchor) {
-        mc.world.removeBlock(BlockPos.ofFloored(anchor), false);
+        BlockPos anchorPos = BlockPos.ofFloored(anchor);
+        mc.world.removeBlock(anchorPos, false);
         double damage = bedDamage(player, anchor);
-        mc.world.setBlockState(BlockPos.ofFloored(anchor), Blocks.RESPAWN_ANCHOR.getDefaultState());
+        mc.world.setBlockState(anchorPos, Blocks.RESPAWN_ANCHOR.getDefaultState());
```
tyrannus00, line 158: The block state shouldn't be set to the default state, because in case the damage calcs get ran on an anchor that has a charge and it doesn't get blown up, it will remove the charge from the client. The block state should rather be cached and reused afterwards.

RacoonDog: I am already aware of this, I'm wondering if there's a way to fix this that doesn't require modifying the actual world instead

PR #4097, opened 2023-09-24
Added item type filters to tracers & nametags

paredes00:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description
Added options to filter specific items when the Item entity is included in Nametags/Tracers modules. Useful if you want item nametags without cluttering the screen with useless items (e.g. netherrack), or if you're looking for specific dropped items. Additional option to always highlight named items.

Code is nearly the same between the two modules, not sure if thats an issue

## Related issues

None as far as I know

# How Has This Been Tested?

image
image
image

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
@@ -5,6 +5,8 @@
 
 package meteordevelopment.meteorclient.systems.modules.render;
 
+import java.util.HashSet;
+import java.util.List;
 import meteordevelopment.meteorclient.events.render.Render2DEvent;
 import meteordevelopment.meteorclient.events.render.Render3DEvent;
 import meteordevelopment.meteorclient.renderer.Renderer2D;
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -73,6 +79,33 @@ public enum TracerStyle {
...
+    // items
+
+    private final Setting<ListMode> shownItemFilterType = sgItems.add(new EnumSetting.Builder<ListMode>()
+        .name("Shown mode")
+        .description("Shown item filter mode")
+        .defaultValue(ListMode.All)
+        .build()
+    );
+
+    private final Set<Integer> shownItemIds = new HashSet<>();
```
RacoonDog, line 91: make this a fastutil `IntHashSet`

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Tracers.java
```diff
@@ -372,4 +422,11 @@ private float getAlpha() {
     public String getInfoString() {
         return Integer.toString(count);
     }
+
+    public enum ListMode {
+        Whitelist,
+        Blacklist,
+        None,
```
RacoonDog, line 429: instead of having a `None` mode, make a boolean setting called "filter-items" or something which hides the three new settings when disabled

RacoonDog (changes requested):

paredes00:
Done and done

image
image
image

I kept the None mode but removed All, I think this makes more sense since disabling the filter is effectively "all", selecting none lets you show only those with custom names.

RacoonDog:
> Done and done
> 
> image image image
> 
> I kept the None mode but removed All, I think this makes more sense since disabling the filter is effectively "all", selecting none lets you show only those with custom names.

Oh perfect that’s what i meant anyways

sebseb7:
have it also filter item frames for items, very useful for finding elytra.

Nekiplay:
> have it also filter item frames for items, very useful for finding

Use Meteor+ its has ItemFrame ESP

GoodGuyCarl (approved):

PR #4098, opened 2023-09-25
mark Quilt versions below 0.23.0 as incompatible

ManInMyVan:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

This pull request marks Quilt versions under 0.23.0 as incompatible with Meteor. Meteor crashes on startup when using Quilt 0.22.0 and below, with this change, Quilt users that haven't updated will know what's wrong.
# How Has This Been Tested?

Tested using quilt 0.20.2+ on 1.20.4

# Checklist:

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -54,6 +54,7 @@
     "origins": "*",
     "wurst": "*",
     "sodium": "<0.5.6",
-    "morechathistory": "*"
+    "morechathistory": "*",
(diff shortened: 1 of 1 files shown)


Wide-Cat:
I just tested this, and meteor seems to work fine on 0.21.0 but not anymore on 0.20.2. Could you update this to reflect that?

ManInMyVan:
Interesting, 0.20.2 still works for me on the latest dev build

PR #4133, opened 2023-10-12
Make Infinite Chat Box actually "infinite"

Paddyk45:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Removes the part that checks if the max limit is reached instead of using the max integer limit

## Related issues

None

# How Has This Been Tested?

On a server

# Checklist:


Paddyk45:
> Also I dont see how this overly complicated and error prone solution works more reliably than simply increasing the limit. 
i just took the code from the TextFieldWidget's write() and removed the part that stops you from entering more than the max limit

Paddyk45:
And here is how to reproduce it not working:
0. Enable Better Chat and Infinite Chat Box
1. Join any server
2. Open the chat box (obviously)
3. Paste this:
`aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`
4. Try to type more than that. It will not work even though "Infinite" chat box is enabled

Paddyk45:
Okay looks like it was fixed on 1.20.2, I was testing the non-working one on 1.20.1

Paddyk45:
Okay appearently it was caused by an addon

PR #4148, opened 2023-10-18
Add loop command

Coderx-Gamer:
Added "loop" (or alias "lp") command which allows you to loop a meteor command any amount of times.

Usage:
  - loop\/lp \<amount of times to loop\> \<meteor command without prefix\>

This could potentially be useful for macros or packet spamming.


Fiz-Victor:
Maybe use spam rn

PR #4190, opened 2023-10-31
fix small hud details

RacoonDog:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

- Display crosses on Player Model Hud and Combat Hud when in editor and there is no content displayed, as they may have a transparent background.
- Move tasks from `render()` and `tick()` to setting `onChanged()` listeners

# How Has This Been Tested?

image
(Combat Hud player model is an unrelated issue)
image

# Checklist:


Wide-Cat:
Looks like it works fine for exisiting hud elements, but if you add a new combat hud or player model element the size isn't set properly or something, and they're literally just single pixels. The white dots you can see are the newly added hud elements.

image

PR #4207, opened 2023-11-10
Improve BetterChat

bruh1273:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Code improvement

## Description

The changes I provided allows the BetterChat module's mixins to be more optimized by not injecting into the class(es) directly but by simply modifying the variables that the class checks. By doing so, I am able to optimize the runtime injection, and I can also remove an unused mixin class as a result.

## Related issues

No issues.

# How Has This Been Tested?

https://cdn.discordapp.com/attachments/402585927135789077/1172442595381760041/2023-11-10_00-47-35.mp4
The result of this change is not visual.

# Checklist:


Symmettry:
Why was this closed? It 100% should be added.

Wide-Cat:
It 100% shouldn't. If I remember correctly it broke infinite chat box with ViaFabricPlus

Symmettry:
Doesn't make much sense, it seems to work just fine when I try it.

PR #4214, opened 2023-11-13
Update PlayerListHudMixin.java

P529:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Fixes compatiblity with future

Would love to push this of 0.5.4

# Checklist:


RacoonDog:
it would be better to have it work alongside future instead of disabling the mixin

P529:
> it would be better to have it work alongside future instead of disabling the mixin

I get that but Future is modifying the same constant and my approach was that it wouldn't really matter if one was at least left with the option from future, I can look for another way to solve this.

0xTas:
> it would be better to have it work alongside future instead of disabling the mixin

This is accomplished by [#4454](https://github.com/MeteorDevelopment/meteor-client/pull/4454)

Wide-Cat:
3145c37fa1473fa5877a665057dc6d99eee0395c

PR #4230, opened 2023-11-22
PMspam: /msg spam support

Lagmaker:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Added a checkbox that will parse nicknames of people on server 
And spam them in private messages 

## Related issues

No issues, but i have idea to add "ignore friends" and cache parsed results for more optimization

# How Has This Been Tested?

It was tested by many ways and now it works perfectly.
No problems, whether it's single-player, local, or multiplayer

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
@@ -76,6 +76,15 @@ public class Spam extends Module {
         .build()
     );
 
+    // PMspam button
+    private final Setting<Boolean> PMspam = sgGeneral.add(new BoolSetting.Builder()
+        .name("PMspam")
+        .description("Parses playersList and /msg them")
+        .defaultValue(false)
+        .build()
+    );
+
+
     private int messageI, timer;
 
     public Spam() {
@@ -100,6 +109,15 @@ public class Spam extends Module {
         if (disableOnLeave.get()) toggle();
     }
 
+
+    /**
+     * Iterator through player list<br>
+     * Declared globally to save its value<br><br>
+     * *Will be reset if (it > playerNames.length)
+     */
+    int it = 0;
+
+
     @EventHandler
     private void onTick(TickEvent.Post event) {
         if (messages.get().isEmpty()) return;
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
```diff
@@ -114,7 +132,47 @@ private void onTick(TickEvent.Post event) {
...
+                    // Var to store String[] of names
+                    String[] playerNames = new String[0];
+
+                    // if LOCAL SERVER
+                    if (mc.getServer() != null && mc.isConnectedToLocalServer()) {
+                        playerNames = mc.getServer().getPlayerNames();
+                    }
+                    // if MULTIPLAYER
+                    else if (mc.player.getEntityWorld().getPlayers() != null) {
+                        playerNames = mc.player.networkHandler.getCommandSource().getPlayerNames().toArray(new String[0]);
```
RacoonDog, line 155: Instead of calling `networkHandler.getCommandSource().getPlayerNames()`, you should be using `networkHandler.getPlayerList()`. `getPlayerNames()` allocates a new `List` every method call and internally uses `getPlayerList()`, so you can cut out the middle man and just use it directly. As a bonus, this returns an already cached list so you dont need to worry about it yourself.

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
```diff
@@ -76,6 +76,15 @@ public class Spam extends Module {
         .build()
     );
 
+    // PMspam button
+    private final Setting<Boolean> PMspam = sgGeneral.add(new BoolSetting.Builder()
+        .name("PMspam")
+        .description("Parses playersList and /msg them")
```
RacoonDog, line 82: Avoid using developer syntax in descriptions, most of Meteor Client's userbase aren't used to it. You should just say "the player list" instead of "playersList".

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
```diff
@@ -76,6 +76,15 @@ public class Spam extends Module {
         .build()
     );
 
+    // PMspam button
```
RacoonDog, line 79: Unnecessary comment, the setting is self-explanatory.

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/Spam.java
```diff
@@ -132,46 +123,23 @@ private void onTick(TickEvent.Post event) {
...
-                            text = "/msg " + playerNames[it] + " " + text;
-                            it++;
-                        }
+                if (playerNames.size() > 0) {
+                    if (playerIndex >= playerNames.size()) playerIndex = 0;
+
+                    // Exclude our name     // TODO: Friends name
+                    if (playerNames.get(playerIndex).getProfile().getName().equals(mc.getSession().getUsername()) && playerNames.size() != 1) playerIndex++;
+
+                    if (playerIndex < playerNames.size()) {
```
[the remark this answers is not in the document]
RacoonDog: It's not a big problem, but it wouldn't be working as intended
I think you should implement that fix

RacoonDog (changes requested):

RacoonDog (approved):

RacoonDog:
the warning is due to the fact that it can return null, but only in the main menu
we can safely ignore it

PR #4245, opened 2023-12-02
Make chest swap close inventory because anticheats

maxsupermanhd:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

On some servers anticheat does not allow attacking entities if you moved inventory without closing inventory first.

# How Has This Been Tested?

On latest meteor dev and then with build from unstaged changes, seems like with option enabled anticheat does not block attacks anymore.
Tested on constantiam.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BookBot.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BookBot.java
@@ -275,6 +275,8 @@ public class BookBot extends Module {
                     lineIndex++;
                     // Wrap to next line, unless wrapping to next page
                     if (lineIndex != 14) page.appendCodePoint(c);
+                } else if (lineWidth == 0f && c == ' ') {
+                    continue; // Prevent leading space from text wrapping
                 } else {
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/player/ChestSwap.java
```diff
@@ -120,6 +128,13 @@ private void equipElytra() {
 
     private void equip(int slot) {
         InvUtils.move().from(slot).toArmor(2);
+        if (closeInventory.get()) {
+            // Notchian clients send a Close Window packet with Window ID 0 to close their inventory even though there is never an Open Screen packet for the inventory.
+            var n = mc.getNetworkHandler();
+            if (n != null) {
+                n.sendPacket(new CloseHandledScreenC2SPacket(0));
```
Paddyk45, line 135: Why are you checking if the network handler is null, it never will be null here

Vaeyxx: fried chicken

PR #4283, opened 2023-12-16
FakePlayer Damage

CrytoPal:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Lets the Fake Player take Damage and pop Totems.

## Related issues

none

# How Has This Been Tested?

Trial and Error, Works with multiple Fake Player Entities as well.

# Checklist:

- [] I have added comments to my code in more complex areas.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/commands/FakePlayerCommand.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/commands/FakePlayerCommand.java
@@ -28,13 +28,13 @@ public class FakePlayerCommand extends Command {
         builder.then(literal("add")
             .executes(context -> {
                 FakePlayer fakePlayer = Modules.get().get(FakePlayer.class);
-                FakePlayerManager.add(fakePlayer.name.get(), fakePlayer.health.get(), fakePlayer.copyInv.get());
+                FakePlayerManager.add(fakePlayer.name.get(), fakePlayer.health.get(), fakePlayer.copyInv.get(), Modules.get().get(FakePlayer.class).allowDamage.get());
                 return SINGLE_SUCCESS;
             })
             .then(argument("name", StringArgumentType.word())
                 .executes(context -> {
                     FakePlayer fakePlayer = Modules.get().get(FakePlayer.class);
-                    FakePlayerManager.add(StringArgumentType.getString(context, "name"), fakePlayer.health.get(), fakePlayer.copyInv.get());
+                    FakePlayerManager.add(StringArgumentType.getString(context, "name"), fakePlayer.health.get(), fakePlayer.copyInv.get(), Modules.get().get(FakePlayer.class).allowDamage.get());
                     return SINGLE_SUCCESS;
                 })
             )
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/FakePlayer.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/FakePlayer.java
@@ -42,6 +42,14 @@ public class FakePlayer extends Module {
         .build()
     );
 
+    public final Setting<Boolean> allowDamage = sgGeneral.add(new BoolSetting.Builder()
+        .name("Allow Damage")
+        .description("Allows FakePlayer taking Damage and Popping Totems")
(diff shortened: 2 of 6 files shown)


src/main/java/meteordevelopment/meteorclient/commands/commands/FakePlayerCommand.java
```diff
@@ -28,13 +28,13 @@ public void build(LiteralArgumentBuilder<CommandSource> builder) {
...
                 FakePlayer fakePlayer = Modules.get().get(FakePlayer.class);
-                FakePlayerManager.add(fakePlayer.name.get(), fakePlayer.health.get(), fakePlayer.copyInv.get());
+                FakePlayerManager.add(fakePlayer.name.get(), fakePlayer.health.get(), fakePlayer.copyInv.get(), Modules.get().get(FakePlayer.class).allowDamage.get());
                 return SINGLE_SUCCESS;
             })
             .then(argument("name", StringArgumentType.word())
                 .executes(context -> {
                     FakePlayer fakePlayer = Modules.get().get(FakePlayer.class);
-                    FakePlayerManager.add(StringArgumentType.getString(context, "name"), fakePlayer.health.get(), fakePlayer.copyInv.get());
+                    FakePlayerManager.add(StringArgumentType.getString(context, "name"), fakePlayer.health.get(), fakePlayer.copyInv.get(), Modules.get().get(FakePlayer.class).allowDamage.get());
```
GreemDev, line 37: you do realize fakePlayer is literally a local variable here you dont need to reget it

src/main/java/meteordevelopment/meteorclient/systems/modules/player/FakePlayer.java
```diff
@@ -42,6 +42,14 @@ public class FakePlayer extends Module {
         .build()
     );
 
+    public final Setting<Boolean> allowDamage = sgGeneral.add(new BoolSetting.Builder()
+        .name("Allow Damage")
```
GreemDev, line 46: change to `allow-damage` for consistency with the rest of the entire project's settings

src/main/java/meteordevelopment/meteorclient/systems/modules/player/FakePlayer.java
```diff
@@ -42,6 +42,14 @@ public class FakePlayer extends Module {
...
+        .description("Allows FakePlayer taking Damage and Popping Totems")
```
GreemDev, line 47: `"Allows the fake player spawned to take damage and pop totems."`

src/main/java/meteordevelopment/meteorclient/commands/commands/FakePlayerCommand.java
CrytoPal: Didn't see that, thanks for pointing it out!

src/main/java/meteordevelopment/meteorclient/systems/modules/player/FakePlayer.java
CrytoPal: Forgot to change this when testing, thanks.

Wide-Cat:
I'm not sure about all these changes. When the damage setting is disabled you can still hit the entity and it simulates taking damage, just without a totem, which it wasn't like before. The movement recording thing seems really finicky; it isn't clear when it's recording and when it isn't, and it's pretty easy to completely break it if you start recording and playing at the same time with loop recording enabled.

Wide-Cat:
I think this just needs a lot more polish. When it works it's pretty good but it just feels like this was committed before it's actually done.

CrytoPal:
@Wide-Cat I fixed the bugs you mentioned.

tympanicblock61:
you could add a way to send a message from one of the fake players

tympanicblock61:
you could instead of having the settings global have them for each fakeplayer (this is just a suggestion dont really think you will add it)

gigglingbed:
bro really asked chat gpt for 3 ideas

Gingerbeard5773:
doesnt carpet have a really good fake player bot, maybe we can just port that somehow

PR #4321, opened 2024-01-10
Jesus improvements

crazycat256:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

This pull request brings the following modifications to the Jesus module:

- Disables the behavior where a player surfaces while swimming
- Fixes the issue where the player wasn't being pulled by bubble columns (but still allows upward swimming in magma block bubble columns)
- Resolves the problem where the player couldn't jump out of water while in a current
- Addresses the issue where the player couldn't enter water from the side

# How Has This Been Tested?
In water, lava, snow powder, bubble columns and waterlogged blocks with both solid and bob mode

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/EntityMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/EntityMixin.java
@@ -13,10 +13,7 @@ import meteordevelopment.meteorclient.events.entity.player.JumpVelocityMultiplie
 import meteordevelopment.meteorclient.events.entity.player.PlayerMoveEvent;
 import meteordevelopment.meteorclient.systems.modules.Modules;
 import meteordevelopment.meteorclient.systems.modules.combat.Hitboxes;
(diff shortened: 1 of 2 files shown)


Wide-Cat:
Are you sure implementing the `isTouchingBubbleColumn` method is necessary? I would have assumed there would be a method already present in minecraft to do that. Could try one of the methods in mc.world similar to `getBlockCollisions`, or just find where bubble column movement is handled in the Player entity classes and see how it detects you're in a bubble column there.

crazycat256:
> Are you sure implementing the `isTouchingBubbleColumn` method is necessary? I would have assumed there would be a method already present in minecraft to do that. Could try one of the methods in mc.world similar to `getBlockCollisions`, or just find where bubble column movement is handled in the Player entity classes and see how it detects you're in a bubble column there.

@Wide-Cat after searching, there seems to be no method that lets you know if a player is touching a bubble column (`mc.world.getBlockCollisions` doesn't work because there are no collisions with fluids).
However, I was able to get around this by using a mixin.

PR #4323, opened 2024-01-11
Allow any characters in cracked names

DAMcraft:
## Type of change

- [X] Bug fix
- [X] New feature

## Description
Cracked account names only allow upper-case, lower-case and underscores. 
While real Minecraft accounts follow this pattern, you can actually have any characters in the name and a vanilla server doesn't care. 
The only software that I'm aware of that checks player names would be Paper (and they even allow dots).
There is no reason to block all other names.

## Related issues
[#12733](https://github.com/meteor/meteor/issues/12733) (posted in the wrong repo)


Tristan3030:
Still doesn't work for me sadly

Issue #3543, opened 2023-04-02
[Bug] Scaffold does not work


Afanasy3000:
### Describe the bug

scaffold stops working if x or z is negative
this problem is similar to incorrect placement of blocks on negative coordinates in spawn proofer. As I understand, the distance to the block is incorrectly calculated if any of the coordinates is negative.

### Steps to reproduce

```bash
to repeat this bug, you need to stand on any negative coordinates and turn on the scaffold
```

### Meteor Version

0.5.3

### Minecraft Version

1.19.4

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported (I have searched bug reports on GitHub).

- [X] This is a valid bug (I am able to reproduce this on the latest dev build).

Wide-Cat:
should be resolved by 20d44c17daf2614bbf9c3b5ff3b9be294ec6b378

PR #4326, opened 2024-01-12
Fix and improvement of the Scaffold module

crazycat256:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

- Fixed an issue where the non-air-place mode would cause the player to fall when descending.
- Fixed a bug in the air-place mode.
- Improved the logic for placing blocks in the non-air-place mode.
- Enhanced the fast-tower functionality.
- Added the "ahead-distance" setting, which allows choosing the distance at which blocks are placed in front of the player (prevent falling in certain cases).

## Related issues
[[Bug] Scaffold does not work #3543](https://github.com/MeteorDevelopment/meteor-client/issues/3543)

# How Has This Been Tested?

:)

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
@@ -280,7 +280,8 @@ public class Scaffold extends Module {
 
     public boolean towering() {
         return isActive() && fastTower.get() && mc.options.jumpKey.isPressed() && !mc.options.sneakKey.isPressed() &&
-            (whileMoving.get() || (!mc.options.forwardKey.isPressed() && !mc.options.backKey.isPressed() && !mc.options.leftKey.isPressed() && !mc.options.rightKey.isPressed()));
+            (whileMoving.get() || (!mc.options.forwardKey.isPressed() && !mc.options.backKey.isPressed() && !mc.options.leftKey.isPressed() && !mc.options.rightKey.isPressed())) &&
+            (!onlyOnClick.get() || (onlyOnClick.get() && mc.options.useKey.isPressed()));
     }
 
     private boolean validItem(ItemStack itemStack, BlockPos pos) {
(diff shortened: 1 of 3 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
```diff
@@ -259,9 +253,23 @@ private void onTick(TickEvent.Pre event) {
...
-            if (!mc.world.getBlockState(bp).isAir()) {
-                prevBp.set(bp);
+            place(bp);
+        }
+
+        if (fastTower.get() && mc.options.jumpKey.isPressed() && !mc.options.sneakKey.isPressed() && InvUtils.findInHotbar(itemStack -> validItem(itemStack, bp)).found()) {
+            Vec3d velocity = mc.player.getVelocity();
+            Box playerBox = mc.player.getBoundingBox();
+            if (Streams.stream(mc.world.getBlockCollisions(mc.player, playerBox.offset(0, 1, 0))).toList().isEmpty() /*||
+                !Streams.stream(mc.world.getBlockCollisions(mc.player, playerBox.offset(0, -1, 0))).toList().isEmpty()*/) {
```
Wide-Cat, line 263: 👁️

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
```diff
@@ -159,88 +170,72 @@ public class Scaffold extends Module {
...
-    public void onActivate() {
-        lastWasSneaking = mc.options.sneakKey.isPressed();
-        if (lastWasSneaking) lastSneakingY = mc.player.getY();
-    }
 
     @EventHandler
     private void onTick(TickEvent.Pre event) {
         if (onlyOnClick.get() && !mc.options.useKey.isPressed()) return;
 
+        Vec3d vec = mc.player.getPos().add(mc.player.getVelocity()).add(0, -0.75, 0);
```
Wide-Cat, line 183: is 0.75 for gravity or something?

Wide-Cat (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/movement/Scaffold.java
crazycat256: This value determines how far below the player the blocks are placed, so logically it should be the smallest value greater than -1 (-0.999...).
However, setting it to -0.75 avoids placing blocks under the bridge like this:

https://github.com/MeteorDevelopment/meteor-client/assets/63475640/eb6e9f40-4592-40cb-ba4d-cf62742f1489

Wide-Cat:
It seems to be quite buggy if you have fast tower enabled and no blocks available to place. If you have no blocks at all in your hotbar you seem to be prevented from jumping when you stand still and it tries to tower up, while if you do have blocks in your hotbar but you have autoswitch disabled and you arent holding a stack, you go up but dont place any blocks at all. It would also be nice to have a speed setting for fast tower.

crazycat256:
@Wide-Cat I've fixed the bugs you describe and added a setting to control the speed of the fast-tower.

Wide-Cat (approved):

PR #4334, opened 2024-01-15
singleton command arguments

RacoonDog:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Since the command argument types are stateless, we can re-use singleton instances to lower memory usage a tiny bit.
(760 bytes to 376 bytes, profiled with regular meteor)

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/arguments/CompoundNbtTagArgumentType.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/arguments/CompoundNbtTagArgumentType.java
@@ -18,15 +18,18 @@ import java.util.Collection;
 import static net.minecraft.nbt.StringNbtReader.EXPECTED_VALUE;
 
 public class CompoundNbtTagArgumentType implements ArgumentType<NbtCompound> {
+    private static final CompoundNbtTagArgumentType INSTANCE = new CompoundNbtTagArgumentType();
     private static final Collection<String> EXAMPLES = Arrays.asList("{foo:bar}", "{foo:[aa, bb],bar:15}");
 
     public static CompoundNbtTagArgumentType create() {
-        return new CompoundNbtTagArgumentType();
+        return INSTANCE;
     }
 
     public static NbtCompound get(CommandContext<?> context) {
         return context.getArgument("nbt", NbtCompound.class);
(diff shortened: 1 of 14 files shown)


JAXPLE:
These functions can be problematic when they run as a single instance.
Single tone should always be considered before using, and if you use it, I recommend the pattern I reviewed

RacoonDog:
> These functions can be problematic when they run as a single instance.
> Single tone should always be considered before using, and if you use it, I recommend the pattern I reviewed

in this case, using singletons is safe because the object is stateless and immutable. i dont understand the difference between my proposed changes and your review suggestion though?

JAXPLE:
> > These functions can be problematic when they run as a single instance.
> > Single tone should always be considered before using, and if you use it, I recommend the pattern I reviewed
> 
> in this case, using singletons is safe because the object is stateless and immutable. i dont understand the difference between my proposed changes and your review suggestion though?

My code and your code do the same thing
The difference is that my code will create instances more lazily and run reliably
Considering the side effects, there's no problem with the current code

RacoonDog:
> > > These functions can be problematic when they run as a single instance.
> > > Single tone should always be considered before using, and if you use it, I recommend the pattern I reviewed
> > 
> > 
> > in this case, using singletons is safe because the object is stateless and immutable. i dont understand the difference between my proposed changes and your review suggestion though?
> 
> My code and your code do the same thing The difference is that my code will create instances more lazily and run reliably Considering the side effects, there's no problem with the current code

Ah, since these are all initialized on startup via command registration, there’s no need to have lazy loading

JAXPLE:
> > > > These functions can be problematic when they run as a single instance.
> > > > Single tone should always be considered before using, and if you use it, I recommend the pattern I reviewed
> > > 
> > > 
> > > in this case, using singletons is safe because the object is stateless and immutable. i dont understand the difference between my proposed changes and your review suggestion though?
> > 
> > 
> > My code and your code do the same thing The difference is that my code will create instances more lazily and run reliably Considering the side effects, there's no problem with the current code
> 
> Ah, since these are all initialized on startup via command registration, there’s no need to have lazy loading

This was an init instance.
Then it doesn't matter lol

PR #4355, opened 2024-01-25
Individual Nametags configuration for each EntityType

crazycat256:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Changes:
- Allows configuring nametags differently for each EntityType.
- Adds a "trough-walls" option.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/DefaultSettingsWidgetFactory.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/DefaultSettingsWidgetFactory.java
@@ -64,6 +64,7 @@ public class DefaultSettingsWidgetFactory extends SettingsWidgetFactory {
         factories.put(StorageBlockListSetting.class, (table, setting) -> storageBlockListW(table, (StorageBlockListSetting) setting));
         factories.put(ScreenHandlerListSetting.class, (table, setting) -> screenHandlerListW(table, (ScreenHandlerListSetting) setting));
         factories.put(BlockDataSetting.class, (table, setting) -> blockDataW(table, (BlockDataSetting<?>) setting));
+        factories.put(EntityTypeDataSetting.class, (table, setting) -> entityTypeDataW(table, (EntityTypeDataSetting<?>) setting));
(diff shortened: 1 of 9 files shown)


Theboiboi8:
What's the expected behavior for through-walls when it's disabled? The current approach has rather lackluster and unrefined results, and a different approach might be better.

crazycat256:
@Theboiboi8 The expected behavior when disabled is simply not to display nametags of entities that are not directly visible to the player.

What do you think could be improved?

Theboiboi8:
Nametags don't render even when part of the entity is in view, so you can be looking at most of the entity and still not see any nametags. i.e. 
image
In this example, nametags are enabled, `through-walls` is set to `false`, and I'm using a fake player (`.fake-player`) to have a non-moving entity. As observed, even though part (most) of the fake player is directly visible, the nametag doesn't appear.
image
In contrast, when the player moves ever so slightly to the right, as to view more of the fake player, the nametag appears, as observed in the above screenshot.

I feel as though this isn't the correct behavior, and I feel that when at least part of an entity is showing, nametags should, logically, appear, as otherwise issues like the one illustrated previously occur and hinder the player.

crazycat256:
It would be quite complicated because it would mean raycasting all over the entity, which would rapidly reduce fps

Theboiboi8:
replacing
```java
((IVec3d) vec2).set(entity.getX(), entity.getY() + entity.getStandingEyeHeight(), entity.getZ());
```
with
```java
((IVec3d) vec2).set(entity.getX() + entity.getWidth(), entity.getY() + entity.getStandingEyeHeight(), entity.getZ() + entity.getWidth());
```
in the `isBehindWall(Entity entity)` boolean seems to yield okay results without compromising on FPS (~5 FPS difference vs without this change, within normal FPS fluctuation range for MC on my machine). This solution isn't perfect, particularly when it comes to the Y-axis, but it does offer a simple and acceptable solution for the problem outlined in my initial comment.

crazycat256:
This modification just adds an unwanted offset to the eye position.
Place your fake-player to the left of your blocks to see this

Issue #4378, opened 2024-02-02
I keep crashing using crystal aura


Thesussycheese69420:
### Steps to reproduce

When someone comes to me using speed I keep crashing even before I place crystals to kill him/her and then it says Saving World... before crashing

### Link to crash report/log (upload to https://mclo.gs and paste resulting link here)

https://mclo.gs/7sutioD

### Meteor Version

Meteor 0.5.6

### Minecraft Version

1.20.4

### Operating System

Windows

### Before submitting a crash report

- [X] This crash wasn't already reported (I have searched crash reports on GitHub).

- [X] This is a valid crash (I am able to reproduce this on the latest dev build).

RacoonDog:
Please post crash log and crystal aura settings

Wide-Cat:
hopefully fixed by 4260f7c0af660a9aba81b787ef5e547393c235bd

PR #4383, opened 2024-02-05
Fix multiple key crash in StatusEffectBruteForce

stanleymw:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

When brute-forcing effects, Absorption was always added to the set of possible effects, and then later added again when using the tracked data. This lead to a crash with an error along the lines of "java.lang.IllegalArgumentException: Multiple entries with same key", while using modules such as Crystal Aura. My changes make it so that only the tracked data code adds the Absorption to the list of all possible effects.

## Related issues

I think this is the related [issue](https://github.com/MeteorDevelopment/meteor-client/issues/4378).

# How Has This Been Tested?

Using the Crystal Aura in-game on a few servers and against other players.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/entity/effects/StatusEffectBruteForce.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/entity/effects/StatusEffectBruteForce.java
@@ -88,8 +88,8 @@ public class StatusEffectBruteForce {
         ALL_ENTRIES.add(StatusEffectEntry.of(StatusEffects.POISON, 2));
         ALL_ENTRIES.add(StatusEffectEntry.of(StatusEffects.WITHER, 1));
         ALL_ENTRIES.add(StatusEffectEntry.of(StatusEffects.HEALTH_BOOST, 1));
-        ALL_ENTRIES.add(StatusEffectEntry.of(StatusEffects.ABSORPTION, 1));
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/utils/entity/effects/StatusEffectBruteForce.java
```diff
@@ -151,8 +151,11 @@ private static void update(int particleColor, LivingEntity entity, EntityEffectC
             int absorptionLevel = Math.round(entity.getAbsorptionAmount() / 4f);
             if (absorptionLevel <= 4) {
                 possibleEntries = new ReferenceOpenHashSet<>(ALL_ENTRIES);
-                possibleEntries.add(ABSORPTION_STRONG);
-                if (absorptionLevel <= 1) possibleEntries.add(ABSORPTION);
+                if (absorptionLevel <= 1) {
+                    possibleEntries.add(ABSORPTION);
+                } else {
```
RacoonDog, line 156: the absorption hearts only give us the lower bound for the absorption level, it should still add both if `absorptionLevel <= 1`

stanleymw: Thanks

RacoonDog: yes im terrible at naming variables :)

PR #4399, opened 2024-02-11
Security patches

RacoonDog:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Patches a vulnerability within Meteor Client where when clicking a message with a click event sent by a malicious server, it could establish a swarm connection and execute arbitrary Minecraft & Meteor Client commands from the client.
This connection had the ability to:
- Send any signed chat message (which could lead to a Microsoft account ban).
- Overwrite any specific non-protected file in the same drive as the Meteor Client folder.
- If Meteor Rejects is installed, shut down the computer.

In order to fix this vulnerability and prevent future ones, this pull request:
- Prevents Meteor Client commands from being executed from a click event that does not come from the client
- Prevents `.notebot` from writing to files outside of its directory.
- Prevents swarm from executing commands outside of its command tree.
- Adds a confirmation dialog to `.swarm join`

# How Has This Been Tested?

Extensively tested using a proof-of-concept malicious server.

# Checklist:


DAMcraft:
:3

PR #4415, opened 2024-02-19
Fix BreakDelay

crazycat256:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Fix the `no-insta-break` setting, which doesn't work and remains active even when the module is disabled.


hexadecimal233:
maybe use `isActive()` instead?

Issue #2404, opened 2022-06-16
Server Spoof not work correcly [Bug] 


Nekiplay:
### Describe the bug

![image](https://user-images.githubusercontent.com/35975332/174006538-54a2f5ce-78c6-4af1-8ba7-1732a67c4a27.png)

### Steps to reproduce

```bash
Join sever, anticheat Wraith
```

### Meteor Version

v0.5.0

### Minecraft Version

1.18.2

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported.
(I have checked every bug report on github)

- [X] This is a valid bug.
(I am able to reproduce this on the latest dev build)

Nekiplay:
Need block this channel's 
![image](https://user-images.githubusercontent.com/35975332/174007541-93441a0c-429a-4320-b09c-fb3e8f4da8b1.png)

Wide-Cat:
open a pr with those changes into meteor

PR #4446, opened 2024-03-09
Fixed server spoof not spoofing branding properly and added exposed only distance setting to X-Ray

Hackerokuz:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

Fixes a bug where turning on brand spoofing in server spoof does not spoof the brand.

Adds a feature to x-ray to filter blocks based on how may blocks should be opaque, just like baritone has.

## Related issues

Fixes issue: #2404  

# How Has This Been Tested?

Checked if the exposedOnly option works with paper anti x-ray using [recommended settings from paper website](https://docs.papermc.io/paper/anti-xray)

Tested server brand spoofing using Vulcan anti cheat.

**Warning:** I do not have access to Wraith anti cheat so I could not test the server spoof with that. Vulcan does not seem to detect the fact that branding is being spoofed. This is Vulcan though so who can say how good this test is...

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/ServerSpoof.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/ServerSpoof.java
@@ -82,7 +82,6 @@ public class ServerSpoof extends Module {
     	
     	@EventHandler
     	private void onClientBrandRetrive(ClientBrandRetrieverEvent event) {
-    		System.out.println("Listener: " + event.info.getReturnValue());
     		if (spoofBrand.get()) event.info.setReturnValue(brand.get());
     	}
     	
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/ServerSpoof.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/ServerSpoof.java
@@ -5,10 +5,19 @@
 
 package meteordevelopment.meteorclient.systems.modules.misc;
 
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+
 import io.netty.buffer.Unpooled;
 import meteordevelopment.meteorclient.MeteorClient;
+import meteordevelopment.meteorclient.events.game.ClientBrandRetrieverEvent;
 import meteordevelopment.meteorclient.events.packets.PacketEvent;
-import meteordevelopment.meteorclient.settings.*;
+import meteordevelopment.meteorclient.settings.BoolSetting;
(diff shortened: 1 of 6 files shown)


src/main/java/meteordevelopment/meteorclient/utils/world/BlockUtils.java
```diff
@@ -317,12 +333,31 @@ public enum MobSpawn {
...
+				}
+				BlockState neighbourBlockState = mc.world.getBlockState(blockPos.offset(direction, i));
+				if(!mc.world.getBlockState(blockPos.offset(direction, i)).isOpaque()) {
+					foundAir = true;
+				} else {
+					break;
+				}
+            }
+		}
+
```
19MisterX98, line 358: Wtf even is this.
- The EXPOSED_POS was used to not create blockstate objects inside a hot loop. Your code does that and you didn't even clean up the EXPOSED_POS variable.
- you don't use the local variable neighbourBlockState which you just created
- you can return true early instead of setting isPosExposed = true and breaking the outer loop

- putting that return logic inside an if statement inside the inner loop is messy and you have a bug now. At least I think that [ore] [air] [block] should not be exposed if distance to check is 2.
```
	outerloop:
        for (Direction direction : Direction.values()) {
            for (int i = 1; i <= distanceToCheck; i++) {
                BlockState neighbourBlockState = mc.world.getBlockState(blockPos.offset(direction, i));
                if(neighbourBlockState.isOpaque()) {
                    continue outerloop;
                }
            }
            return true;
        }
```

19MisterX98 (changes requested):
haven't looked at the rest too much

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/ServerSpoof.java
```diff
@@ -73,14 +79,34 @@ public ServerSpoof() {
...
+	                return;
+	            }
+            }
 
-            if (spoofBrand.get() && id.equals(BrandCustomPayload.ID))
-                event.packet.write(new PacketByteBuf(Unpooled.buffer()).writeString(brand.get()));
+            
 
             if (blockChannels.get()) {
                 for (String channel : channels.get()) {
```
hexadecimal233, line 112: Just defaulting `channels` and set it to `fabric:registry/sync` `fabric:container/open` `fabric-screen-handler-api-v1:open_screen`

Wide-Cat:
This would be better served being split into two separate prs

PR #4453, opened 2024-03-12
Add starscript function meteor.get_module_setting(module, setting)

maxsupermanhd:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Allows for something like `.settings elytra-fly horizontal-speed {meteor.get_module_setting('elytra-fly', 'horizontal-speed')+1}`

## Todo

Update wiki on merge

## Related issues

Essentially https://github.com/MeteorDevelopment/meteor-client/pull/3256 but not scuffed

# How Has This Been Tested?

I ran it once to see if it will correctly return numbers with macro described above, more testing probably required for weird edge cases.

# Checklist:

- [x] I have added comments to my code in more complex areas. (there were none)

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/misc/MeteorStarscript.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/misc/MeteorStarscript.java
@@ -60,6 +60,7 @@ import java.time.LocalTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


MrBreakNFix (approved):
i need this

MineGame159:
I think it would make more sense if the module name was the first argument instead. Also IMO lists are common enough to have a special case for them and return a string like "6 items" (I don't mean minecraft items) instead of the default toString().

maxsupermanhd:
> it would make more sense if the module name was the first argument instead
Changed it

> lists are common enough to have a special case
Added, as a number, formatting can be done in starscript itself

maxsupermanhd:
@MineGame159 I can not seem to find a way to edit the wiki (is it even possible for non-member?), this page needs an update <https://github.com/MeteorDevelopment/meteor-client/wiki/Starscript>

MineGame159:
I don't think it is. I will do it a bit later.

Issue #4377, opened 2024-02-01
Chat event (Better Chat prefix/suffix, other addons) doesn't work


FurnyGo:
### Describe the bug

Prefix and suffix in Better Chat and same modules for chat in different addons (like ChatColor module in Lyra addon) don't work

### Steps to reproduce

1. Enable prefix and suffix in Better Chat
2. Send message
3. No prefix and suffix

### Meteor Version

0.5.5 and 0.5.6-2005

### Minecraft Version

1.20.2 and 1.20.4

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported (I have REALLY searched bug reports on GitHub).

- [X] This is a valid bug (I am REALLY able to reproduce this on the latest dev build).

FurnyGo:
fix pls?

PR #4489, opened 2024-03-29
sendChatMessage event fix (Better Chat fix)

FurnyGo:
## Type of change

- [x] Bug fix

## Description

Fix sendChatMessage event baritone check

## Related issues

#4377 

# How Has This Been Tested?

Don't install baritone, run meteor and enable better chat prefix/suffix/fancy text function

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/ClientPlayNetworkHandlerMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/ClientPlayNetworkHandlerMixin.java
@@ -140,7 +140,7 @@ public abstract class ClientPlayNetworkHandlerMixin extends ClientCommonNetworkH
     private void onSendChatMessage(String message, CallbackInfo ci) {
         if (ignoreChatMessage) return;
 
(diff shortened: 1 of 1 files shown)


FurnyGo:
bump?

FurnyGo:
@Wide-Cat can you check? This is literally 4 chars change in code

Wide-Cat:
I'll test it this evening, but the boolean logic still doesn't look right to me. It's supposed to send the event if the message doesnt start with meteors prefix, and baritone is not available or it doesn't start with baritones prefix, right?

Wide-Cat:
this doesnt work btw you cant use any of meteors commands anymore

FurnyGo:
> It's supposed to send the event if the message doesnt start with meteors prefix, and baritone is not available or it doesn't start with baritones prefix, right?
It looked like this before the changes
If you change "or" to "and" and vice versa (as in pull request) everything works fine

Wide-Cat:
Ok now it will only send the event if baritone is installed. Literally just read through the if statement in your head and see if it makes sense or not.

FurnyGo:
I hope everything works fine now

PR #4520, opened 2024-04-13
Improve block placements

Julflips:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

- Add workstation blocks to clickable blocks. If you try to place on interactable workstations the player interacts with it instead
- Prefer neighbors in the direction the player is looking at. On high anti-cheat servers, the packet gets rejected if the player is unable to look at specific sides of a block. We can improve our chances by preferring sides that the player could look at.

# How Has This Been Tested?

In Singleplayer and on Constatiam, using both the Scaffold and a custom module using the BlockUtils.place() method

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/world/BlockUtils.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/world/BlockUtils.java
@@ -10,10 +10,7 @@ import meteordevelopment.meteorclient.events.world.TickEvent;
 import meteordevelopment.meteorclient.systems.modules.Modules;
 import meteordevelopment.meteorclient.systems.modules.player.InstaMine;
 import meteordevelopment.meteorclient.utils.PreInit;
-import meteordevelopment.meteorclient.utils.player.FindItemResult;
-import meteordevelopment.meteorclient.utils.player.InvUtils;
-import meteordevelopment.meteorclient.utils.player.Rotations;
-import meteordevelopment.meteorclient.utils.player.SlotUtils;
+import meteordevelopment.meteorclient.utils.player.*;
 import meteordevelopment.orbit.EventHandler;
 import meteordevelopment.orbit.EventPriority;
 import net.minecraft.block.*;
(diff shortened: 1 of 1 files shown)


Wide-Cat:
Could you send the custom module you used to test it, and any screenshots or video you have of it working as intended would also be useful.

Julflips:
Ok, let's start with the easier one. In the 2 attached videos, I try to use the scaffold module to place blocks on 4 interactable workstations. As you can see the current version interacts with them instead of using the blocks on the side.

Current version:
https://github.com/MeteorDevelopment/meteor-client/assets/104793554/4a460f69-9517-4d4a-9edb-df67359ca2cf

Fixed version:
https://github.com/MeteorDevelopment/meteor-client/assets/104793554/0d66b43e-8c3c-4b53-b304-aef777fc414d

Julflips:
Regarding the fix in the getPlaceSide(BlockPos blockPos) method:

Let's assume we want to place a block in the center in the following scenario:
https://github.com/MeteorDevelopment/meteor-client/assets/104793554/b6b5d0fa-daa9-4a02-af24-d546e9f7bd9a

Meteor will just return the first viable neighbor it finds. Since there is no block above and the one below is interactable it will always return the north block. So the side the client would need to place the block on is the south side of the north block. However, this side is not facing the player and any decent anti-cheat (like Constatiams) will reject the placement.

The fixed version calculates a vector from the player to the target pos and prioritizes the best neighbor based on it. In the example, the vector would be something like (0, -1.5, 1) since the player is north and above the target pos. So the client will first check the block below and then the south neighbor. If a non-clickable block is found its returned.


In the following example video I use a module that just prints me info about the side of the blockinteract packets of the client. (Just a simple onSendPacket-event that prints every PlayerInteractBlockC2SPacket). Remember that the place side is always the opposite of the selected neighbor.

Current version:
https://github.com/MeteorDevelopment/meteor-client/assets/104793554/a03c1175-d6f8-4761-a347-ffb6574d4ce3

Fixed version:
https://github.com/MeteorDevelopment/meteor-client/assets/104793554/127ffb15-1e0d-4286-a0b7-754090acc017


If one really wants to fix this you would need to use raycasts the check if you can actually look at the side of the neighbor, but I think this is already a really good approximation.

Wide-Cat:
Ok the math for choosing the best side looks good, sorry for needing more clarification on that. However, I'm still rather hesitant on your implementation. Instead of performing all those allocations it would be more efficient to add variables to track the best side and how much you are looking in it's direction, and do the calculation within the for loop. Something more like this (this code is untested):

image

PR #4531, opened 2024-04-19
Make Waypoints use Type

abb3v:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Previously, the deletion of Waypoints was simply determined based on a hardcoded condition:
```java
if (wp.name.get().startsWith("Death ") && "skull".equals(wp.icon.get())) {
```
This isn't really the best, as if the default name of Deaths in `addDeath()` were to change, this clause would also have to be modified. Also Waypoint Types could be useful in the future, if perhaps adding waypoint categories ever were to become a thing. Overall this is just something good to have in case.
# Checklist:


abb3v:
Shit this doesnt work

PR #4546, opened 2024-04-24
Allow Trajectories to be used with off-hand item

snarkyerica:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Allows Trajectories to be used with off-hand item stack, by checking both the MainHand and OffHand stacks when determining whether the item in the stack is in the whitelist instead of just the MainHand stack before returning, even if the MainHand stack is not null.

# How Has This Been Tested?

image

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Trajectories.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Trajectories.java
@@ -140,8 +140,6 @@ public class Trajectories extends Module {
 
         // Get item
         ItemStack itemStack = player.getMainHandStack();
-        if (itemStack == null) itemStack = player.getOffHandStack();
-        if (itemStack == null) return;
         if (!items.get().contains(itemStack.getItem())) {
             itemStack = player.getOffHandStack();
             if (!items.get().contains(itemStack.getItem())) return;
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Trajectories.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Trajectories.java
@@ -142,7 +142,10 @@ public class Trajectories extends Module {
         ItemStack itemStack = player.getMainHandStack();
         if (itemStack == null) itemStack = player.getOffHandStack();
         if (itemStack == null) return;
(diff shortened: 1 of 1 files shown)


abb3v:
You're making a redundant function call twice @snarkyerica

RacoonDog:
> You're making a redundant function call twice @snarkyerica; you are checking the off-hand stack twice if the main hand stack does not contain an item in the list items, this can easily be simplified down to;
> 
> ```java
> ItemStack itemStack = player.getMainHandStack();
> if (itemStack == null || !items.get().contains(itemStack.getItem())) {
>     itemStack = player.getOffHandStack();
>     if (itemStack == null || !items.get().contains(itemStack.getItem())) return;
> }
> 
> // Calculate paths
> if (!simulator.set(player, itemStack, 0, accurate.get(), tickDelta)) return;
> ```

I’d have to double-check to be sure, but I’m not entirely certain these methods can return null values in the first place. IIRC they return a constant empty `ItemStack` instance and you can omit null checks entirely.

snarkyerica:
Yeah, I don't think they can return null, looking at the underlying methods, they always seem to return ItemStack.EMPTY if no other item is found, and IDEA doesn't complain about it potentially returning null, so I've removed those checks sorry for the initial commit being hacky, it was done with no sleep lol

abb3v:
> > You're making a redundant function call twice @snarkyerica; you are checking the off-hand stack twice if the main hand stack does not contain an item in the list items, this can easily be simplified down to;
> > ```java
> > ItemStack itemStack = player.getMainHandStack();
> > if (itemStack == null || !items.get().contains(itemStack.getItem())) {
> >     itemStack = player.getOffHandStack();
> >     if (itemStack == null || !items.get().contains(itemStack.getItem())) return;
> > }
> > 
> > // Calculate paths
> > if (!simulator.set(player, itemStack, 0, accurate.get(), tickDelta)) return;
> > ```
> 
> I’d have to double-check to be sure, but I’m not entirely certain these methods can return null values in the first place. IIRC they return a constant empty `ItemStack` instance and you can omit null checks entirely.

Ah right, I didn't bother looking at the functions, I just noticed that, if that were a correct implementation, then it was formatted incorrectly. Though I see no problems with this latest commit.

PR #4574, opened 2024-05-07
port .nbt to 1.20.5+

RacoonDog:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

make .nbt support data components

# How Has This Been Tested?

it was very fun to play around with

# Checklist:

- [ ] My code follows the style guidelines of this project.


Wide-Cat (changes requested):
Suggestions for the remove subcommand seem slightly broken, presumably because it doesn't a system provided by mojang. You need to specify the `minecraft:` prefix for most of them which is inconvenient, but if you put 's', 'r' or 'c' it suggests one component for some reason.

Working as desired:
image

Broken:
image

???:
image
image
image

Wide-Cat (approved):

PR #4594, opened 2024-05-21
Many improvements to AutoSign

RealMuffinTime:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

- Remove hardcoded time placeholder from MeteorStarscript.java
- Added persistent settings to save sign lines to
- Added Starscript support
- Added possibility to also apply back side of sign
- Added possibility to overwrite or keep existent text on sign
  When not applying anything, open edit screen normally
- 

# How Has This Been Tested?
- Placing many signs, in diffrent situations
- On a local vanilla server.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
@@ -8,25 +8,31 @@ package meteordevelopment.meteorclient.systems.modules.world;
 import meteordevelopment.meteorclient.events.game.OpenScreenEvent;
 import meteordevelopment.meteorclient.gui.GuiTheme;
 import meteordevelopment.meteorclient.gui.widgets.containers.WHorizontalList;
+import meteordevelopment.meteorclient.gui.widgets.pressable.WButton;
+import meteordevelopment.meteorclient.gui.widgets.WWidget;
+import meteordevelopment.meteorclient.gui.utils.StarscriptTextBoxRenderer;
 import meteordevelopment.meteorclient.mixin.AbstractSignEditScreenAccessor;
 import meteordevelopment.meteorclient.settings.*;
 import meteordevelopment.meteorclient.systems.modules.Categories;
 import meteordevelopment.meteorclient.systems.modules.Module;
+import meteordevelopment.meteorclient.utils.misc.MeteorStarscript;
 import meteordevelopment.orbit.EventHandler;
+import meteordevelopment.starscript.Script;
 import meteordevelopment.starscript.StandardLib;
 import net.minecraft.block.entity.SignBlockEntity;
 import net.minecraft.client.gui.screen.ingame.AbstractSignEditScreen;
-import net.minecraft.network.packet.c2s.play.*;
-import meteordevelopment.meteorclient.gui.utils.StarscriptTextBoxRenderer;
-import meteordevelopment.meteorclient.utils.misc.MeteorStarscript;
-import meteordevelopment.starscript.Script;
-import meteordevelopment.meteorclient.gui.widgets.WWidget;
-import meteordevelopment.meteorclient.gui.widgets.pressable.WButton;
+import net.minecraft.network.packet.c2s.play.ClientCommandC2SPacket;
+import net.minecraft.network.packet.c2s.play.PlayerInteractBlockC2SPacket;
+import net.minecraft.network.packet.c2s.play.PlayerActionC2SPacket;
+import net.minecraft.network.packet.c2s.play.UpdateSignC2SPacket;
 import net.minecraft.text.Text;
 import net.minecraft.util.Hand;
 import net.minecraft.util.Util;
 import net.minecraft.util.hit.BlockHitResult;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.TimeZone;
 
 public class AutoSign extends Module {
 
@@ -182,7 +188,7 @@ public class AutoSign extends Module {
 
     @EventHandler
     private void onOpenScreen(OpenScreenEvent event) throws InterruptedException {
-        if (!(event.screen instanceof AbstractSignEditScreen)) return;
+        if (!(event.screen instanceof AbstractSignEditScreen)) {return;}
         if (applyingBack) {
             applyingBack = false;
             event.cancel();
@@ -207,8 +213,9 @@ public class AutoSign extends Module {
         if (frontEnabled.get()) {
             boolean written = false;
             for (Text text : sign.getFrontText().getMessages(false))
-                if (!Objects.equals(text.getString(), "") && !written)
+                if (!Objects.equals(text.getString(), "") && !written) {
                     written = true;
+                }
             if (overwrite.get() || !written) {
                 mc.player.networkHandler.sendPacket(new UpdateSignC2SPacket(sign.getPos(), true,
                     frontScripts.get(0) != null ? MeteorStarscript.run(frontScripts.get(0)) : "error",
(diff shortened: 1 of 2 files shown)


RacoonDog:
what was wrong with the time placeholder? it looks correct from looking at the code

RealMuffinTime:
It was not wrong, but it did always overwrite the time placeholder from the Starscript StandardLib.timeFormat object.

RacoonDog:
> It was not wrong, but it did always overwrite the time placeholder from the Starscript StandardLib.timeFormat object.

ahh i see

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+import java.util.Objects;
+import java.util.TimeZone;
 
 public class AutoSign extends Module {
-    private String[] text;
 
-    public AutoSign() {
-        super(Categories.World, "auto-sign", "Automatically writes signs. The first sign's text will be used.");
-    }
+    private boolean applyingBack = false;
```
RacoonDog, line 39: place variables after settings/setting groups

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+
+    private final SettingGroup front = settings.createGroup("Front");
+    private final SettingGroup back = settings.createGroup("Back");
+    private final SettingGroup other = settings.createGroup("Other");
+
+    private final List<Script> frontScripts = new ArrayList<>();
+    private final List<Script> backScripts = new ArrayList<>();
```
RacoonDog, line 46: place variables after settings/setting groups

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+        .description("Enable auto writing front of sign.")
+        .defaultValue(true)
+        .build()
+    );
+
+    private final Setting<String> line1front = front.add(new StringSetting.Builder()
+        .name("front-line-1")
+        .description("Text in first line.")
+        .defaultValue("{player}")
+        .onChanged(strings -> recompile(front, frontScripts, 1))
```
RacoonDog, line 60: instead of passing the `SettingGroup` and getting the string from the index, you can just pass in `strings`

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
-    public AutoSign() {
-        super(Categories.World, "auto-sign", "Automatically writes signs. The first sign's text will be used.");
-    }
+    private boolean applyingBack = false;
+
+    private final SettingGroup front = settings.createGroup("Front");
+    private final SettingGroup back = settings.createGroup("Back");
+    private final SettingGroup other = settings.createGroup("Other");
+
+    private final List<Script> frontScripts = new ArrayList<>();
```
RacoonDog, line 45: instead of using `ArrayList`, make fixed size lists using `Arrays.asList(new String[4])`

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+        }
+    }
+
+    private void recompile(SettingGroup side, List<Script> sideScripts, int line) {
+        String compileLine = (String) side.getByIndex(line).get();
+
+        Script script = MeteorStarscript.compile(compileLine);
 
-        event.cancel();
+        if (sideScripts.size() <= line - 1) {
```
RacoonDog, line 292: you can remove this condition with the fixed size lists

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+        if (!(event.screen instanceof AbstractSignEditScreen)) {return;}
+        if (applyingBack) {
+            applyingBack = false;
+            event.cancel();
+            return;
+        }
 
+        // apply custom format and timezone, not the nicest (temporary overwriting global date and time format)
+        try {
+            StandardLib.dateFormat.applyPattern(date.get());
```
RacoonDog, line 200: i would rather the module not modify global state like that, it would be better to modify `MeteorStarscript` to add specialized time & date functions that accept format strings & timezone identifiers as arguments

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+        }
+
+        boolean updated = false;
         SignBlockEntity sign = ((AbstractSignEditScreenAccessor) event.screen).getSign();
 
-        mc.player.networkHandler.sendPacket(new UpdateSignC2SPacket(sign.getPos(), true, text[0], text[1], text[2], text[3]));
+        if (frontEnabled.get()) {
+            boolean written = false;
+            for (Text text : sign.getFrontText().getMessages(false))
+                if (!Objects.equals(text.getString(), "") && !written) {
```
RacoonDog, line 216: `!Objects.equals(text.getString(), "")` can be rewritten as `!text.getString().isEmpty()`

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+            }
+
+            if (offhand_swapped) {
+                mc.player.networkHandler.sendPacket(new PlayerActionC2SPacket(PlayerActionC2SPacket.Action.SWAP_ITEM_WITH_OFFHAND, mc.player.getBlockPos(), mc.player.getFacing()));
+            }
+        }
+
+        if (backEnabled.get()) {
+            boolean written = false;
+            for (Text text : sign.getBackText().getMessages(false))
```
RacoonDog, line 259: please add braces

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
```diff
@@ -6,42 +6,293 @@
...
+            StandardLib.timeFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+        }
+
+        boolean updated = false;
         SignBlockEntity sign = ((AbstractSignEditScreenAccessor) event.screen).getSign();
 
-        mc.player.networkHandler.sendPacket(new UpdateSignC2SPacket(sign.getPos(), true, text[0], text[1], text[2], text[3]));
+        if (frontEnabled.get()) {
+            boolean written = false;
+            for (Text text : sign.getFrontText().getMessages(false))
```
RacoonDog, line 215: please add braces

RacoonDog (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/world/AutoSign.java
RealMuffinTime: True.

I saw that there is a [pull requests](https://github.com/MeteorDevelopment/starscript/pull/21) which could resolve this issue.

RealMuffinTime:
While testing I found some bigger issues, need to look into how to solve these.

RacoonDog:
> While testing I found some bigger issues, need to look into how to solve these.

what are the issues?

RealMuffinTime:
First something unrelated: sign.getEditor() is always null for me?

So the module works fine on a server which accepts sign editing (vanilla).
But I also tested it on servers where editing is not possible, this turned out to have negative side effects, while having back sign writing enabled. Double placed signs on 2b2t (2b2t doesn't allow editing), every second sign empty on constantiam (older mc version).

RealMuffinTime:
I tried to use sign.getEditor() to verify if the BlockInteraction was successfull and if the sign could be edited.

RealMuffinTime:
Is checked and tested. Added a warning, if server doesn't support back sign editing.

RealMuffinTime:
Hi, I would really like, to see this merged. :)
Also if https://github.com/MeteorDevelopment/starscript/pull/21 is not yet being merged.

Or is there something else holding up this PR?

PR #4534, opened 2024-04-20
1.20.5/.6 update

Big-Iron-Cheems:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Update Meteor for 1.20.5/.6
The 1.20.5 compatibility will be dropped if necessary later on.

## Related issues

None.

# How Has This Been Tested?

This is still WIP, it builds and runs, but could have some bugs.

# Checklist:

- [X] My code follows the style guidelines of this project.
- [X] I have added comments to my code in more complex areas.
- [ ] I have tested the code in both development and production environments.

PR #4597, opened 2024-05-22
Fixed Criticals not working

tyrannus00:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Fixes Criticals packet/bypass mode.
The bug was introduced in the update to 1.20.5/6 (https://github.com/MeteorDevelopment/meteor-client/pull/4534/commits/653eadc8964cf16aff50d6246a854cae9efa7960)

The issue is that when sending a packet from within the PacketEvent.Send stack, that due to the singleton nature of the event, the packet variable would be changed to the newly sent packet, so that this packet effectively gets sent twice while the original packet, which triggered the event, doesn't get sent.
This caused Criticals in packet or bypass mode to cancel the attack packet and send an additional movement packet instead, and potentially caused issues in other places as well.

# How Has This Been Tested?

I tested it myself, and since I am basically reverting to the state before the update this has been tested for years, if you want.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/events/packets/PacketEvent.java
+++ b/src/main/java/meteordevelopment/meteorclient/events/packets/PacketEvent.java
@@ -27,7 +27,6 @@ public class PacketEvent {
     public static class Send extends Cancellable {
         private static final Send INSTANCE = new Send();
 
-        // overwriting this will overwrite the packet being sent
         public Packet<?> packet;
 
         public static Send get(Packet<?> packet) {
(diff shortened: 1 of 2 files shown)


19MisterX98:
dammit. I thought replacing the packet would solve the serverspoof dilemma.
anyways, you should probably revert the rest of the mentioned commit too. Mixin extras is no longer necessary and the comment is simply wrong now.

the brand channel can stay removed. Ill fix serverspoof in another way.

PR #4637, opened 2024-06-12
fix chest swapping after getting an error with pitch40

jtris:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

When chestswap is set to 'always' and you get an error while enabling Pitch40 elytra fly (`"Player must be above uppper bounds!"`), the module itself remains disabled while your chestplate still gets swapped as if the module was active.

image

Now the module has a chance to call its `toggle()` in peace without the chestplate getting swapped again by ElytraFly's `onActivate()`. 

# How Has This Been Tested?
By seeing if it still swaps the chestplate for an elytra.

image

# Checklist:

- [ ] I have added comments to my code in more complex areas. (there are no complex areas)

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/ElytraFly.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/elytrafly/ElytraFly.java
@@ -341,11 +341,11 @@ public class ElytraFly extends Module {
 
     @Override
     public void onActivate() {
+        currentMode.onActivate();
         if ((chestSwap.get() == ChestSwapMode.Always || chestSwap.get() == ChestSwapMode.WaitForGround)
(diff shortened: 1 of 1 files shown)


Wide-Cat:
It seems that with this change the swap happens twice in one tick, which does solve the issue but it would be better to just not swap in the `onActivate` method if the module is disabled. I would change it to something like this instead:

```java
@Override
public void onActivate() {
    currentMode.onActivate();
    if ((chestSwap.get() == ChestSwapMode.Always || chestSwap.get() == ChestSwapMode.WaitForGround)
        && mc.player.getEquippedStack(EquipmentSlot.CHEST).getItem() != Items.ELYTRA && isActive()) {
        Modules.get().get(ChestSwap.class).swap();
    }
}
```

jtris:
That seems to do it, thanks for the swift feedback.

PR #4655, opened 2024-06-18
Implement a 3D sphere mode for the Markers module

moxvallix:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds a 3D sphere as a possible marker.

Such a feature is useful for visualizing the spawn radius of various farms. Also looks cool.

Most of the code already existed in Sphere2DMarker, just modified it to render all layers at once, and optimized it a bit, as it has to do a lot more work to render the full sphere.

Slider caps the radius at 64 (making a diameter of 128, which is useful for most farms), but users can use the text input to make it larger if they really need to. Anything larger than 64 might cause frame drops (512 ran at 7fps for me).

# How Has This Been Tested?
image

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/marker/Sphere3dMarker.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/marker/Sphere3dMarker.java
@@ -1,0 +1,203 @@
+/*
+ * This file is part of the Meteor Client distribution (https://github.com/MeteorDevelopment/meteor-client).
+ * Copyright (c) Meteor Development.
+ */
+
+package meteordevelopment.meteorclient.systems.modules.render.marker;
+
+import meteordevelopment.meteorclient.events.render.Render3DEvent;
+import meteordevelopment.meteorclient.renderer.ShapeMode;
+import meteordevelopment.meteorclient.settings.*;
+import meteordevelopment.meteorclient.utils.misc.Keybind;
+import meteordevelopment.meteorclient.utils.network.MeteorExecutor;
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/render/marker/Sphere3dMarker.java
```diff
@@ -0,0 +1,203 @@
...
+            this.z = z;
+        }
+    }
+
+    public static final String type = "Sphere-3D";
+
+    private final SettingGroup sgGeneral = settings.getDefaultGroup();
+    private final SettingGroup sgRender = settings.createGroup("Render");
+    private final SettingGroup sgKeybinding = settings.createGroup("Keybinding");
+
```
Wide-Cat, line 38: remove this

src/main/java/meteordevelopment/meteorclient/systems/modules/render/marker/Sphere3dMarker.java
```diff
@@ -0,0 +1,203 @@
...
+    );
+
+    private final Setting<SettingColor> lineColor = sgRender.add(new ColorSetting.Builder()
+        .name("line-color")
+        .description("The color of the lines of the blocks being rendered.")
+        .defaultValue(new SettingColor(0, 100, 255, 255))
+        .build()
+    );
+
+    private final HashMap<String, Block> blocks = new HashMap<>();
```
Wide-Cat, line 96: probably better as a fastutil map

Wide-Cat (changes requested):

PR #4675, opened 2024-06-24
Update StorageBlockListSettingScreen.java to use reflection

tympanicblock61:
# How Has This Been Tested?

me

# Checklist:

- [ maybe] My code follows the style guidelines of this project.
- [ nope] I have added comments to my code in more complex areas.
- [ yes] I have tested the code in both development and production environments.

accept this if you want i dont really care


MineGame159:
Opened the code to check it out but I got an idea that could be much better than reflection. The `BlockEntityType` class contains a set of `Block`s. So we could just get the block from that set, get the item for the block and create the name from it's id. The only problem are shulker boxes, since those are the only block entities that have more blocks associated with the type because of colors. So there would have to be special code for that to make sure it picks the one without a color. But imo it would be a nicer approach than reflection which depends on the development mappings.

PR #4572, opened 2024-05-06
Update Chest Swap (MeteorDevelopment#4525)

crazymoose77756:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Chest swap will now recognize all and work with chestplate types, best mode will swap with the best available chestplate in inventory. 
idk when someone would want to swap to a leather chestplate, but now you can. 

Note:
Not the cleanest implementation 

## Related issues

Fixes #4525 by including other chestplate types 

# Checklist:

- [x] My code follows the style guidelines of this project.
- [x] I have added comments to my code in more complex areas.
- [x] I have tested the code in both development and production environments.

PR #4690, opened 2024-06-27
better chestswap then whatever the pr #4572 is

tympanicblock61:
## Type of change

- [x ] Bug fix
- [ ] New feature

## Description
the exact same stuff as #4572 just better code

# How Has This Been Tested?

me

# Checklist:

- [x ] My code follows the style guidelines of this project.
- [x ] I have added comments to my code in more complex areas.
- [x ] I have tested the code in both development and production environments.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/ChestSwap.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/ChestSwap.java
@@ -5,6 +5,8 @@
 
 package meteordevelopment.meteorclient.systems.modules.player;
 
+import it.unimi.dsi.fastutil.objects.Object2IntMap;
+import it.unimi.dsi.fastutil.objects.Object2IntOpenHashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/player/ChestSwap.java
```diff
@@ -72,49 +86,45 @@ public void swap() {
...
         if (bestSlot != -1) equip(bestSlot);
         return bestSlot != -1;
     }
 
+    // modified AutoArmor getScore
+    private int getScore(ItemStack itemStack) {
+        if (itemStack.isEmpty()) return 0;
+
+        int score = 0;
+        Object2IntMap<RegistryEntry<Enchantment>> enchantments = new Object2IntOpenHashMap<>();
```
MineGame159, line 112: Can you make this map a `private static final` field instead to avoid an allocation. The `Utils.getEnchantments()` method already clears the map.

MineGame159 (changes requested):

Issue #4372, opened 2024-01-31
[Suggestion] Will other languages ​​be supported in the future?


andy3559167:
### Describe the feature

Will other languages ​​be supported in the future?
For example Traditional Chinese
ESP will not be displayed when MC switches to Traditional Chinese

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

LLKawi:
when font rendering is fixed

Blalala6868:
Temporarily patch it with [Caxton](https://gitlab.com/Kyarei/caxton/) first

andy3559167:
> 先用[Caxton](https://gitlab.com/Kyarei/caxton/)臨時修補一下

not working

hexadecimal233:
> > 先用[Caxton](https://gitlab.com/Kyarei/caxton/)臨時修補一下
> 
[...]

Issue #1854, opened 2021-12-30
[Bug] Font Rendering


RickyTheRacc:
### Describe the bug

Using a resource pack that alters the font will screw with meteor's rendering when custom font is turned off. Below I'm using the font from the compliance 64x pack.

### Steps to reproduce

```bash
Create a resource pack that alters the Minecraft font and disable "Custom Font" in the config tab.
```

### Meteor Version

1373

### Minecraft Version

1.18.1

### Operating System

Windows

### Before submitting a bug report

- [X] This bug wasn't already reported.
(I have checked every bug report on github)

- [X] This is a valid bug.
(I am able to reproduce this on the latest dev build)

RickyTheRacc:
https://medal.tv/games/minecraft/clips/6rXKJzt9paatM/d1337yNHiWXH?invite=cr-MSxlMDQsMzQwMjE5MDEs

LilyRose2798:
The video doesn't load for me, but I'm having what I think is the same issue. When I use the font from faithful/compliance 32x, custom font disabled (so it uses the resource pack font) and have the text alignment centered, then the text rendering looks really bad because it isn't aligned/rounded to the nearest pixel before being drawn.
I have the scaling set to 1.1494 since that results in an integer scaling amount once the [internal vanilla font scaling of 1.74](https://github.com/MeteorDevelopment/meteor-client/blob/45dd8c7c84252b144ec776bfe20696b44c548145/src/main/java/meteordevelopment/meteorclient/renderer/text/VanillaTextRenderer.java#L25) is applied (1.74 * 1.1494 = ~2), but even with that the font rendering only looks ok with the text alignment set to left since that always results in the text being drawn at an x coordinate of ___.0 whereas [center aligned text divides the width by 2](https://github.com/MeteorDevelopment/meteor-client/blob/45dd8c7c84252b144ec776bfe20696b44c548145/src/main/java/meteordevelopment/meteorclient/gui/themes/meteor/widgets/WMeteorModule.java#L84) so can result in an x coord of ___.5 causing the font to look bad.
I've tried a build where I modified the above line to `x += Math.round(w / 2 - titleWidth / 2);` and that fixed the rendering with centered text, but only with a scaling amount that results in an integer, any other scaling amount still has rendering issues.

arlomcwalter:
[...]

Issue #333, opened 2021-03-24
[Bug]


pindavka:
**Describe the bug**
I do not have a font in the cheat

**Screenshots**
![2021-03-24_17 58 48](https://user-images.githubusercontent.com/81314740/112332407-c4532d80-8cca-11eb-8ebc-3c1e15a05378.png)

**Additional context**
meteor 0.4.2, minecraft 1.16.5

seasnail8169:
Thats because your language is not english, meteor custom font cannot render russian characters.

RedCarlos26:
Is this a bug or it is just not supported?

RedCarlos26:
> Is this a bug or it is just not supported?

@seasnail8169 ?

seasnail8169:
Its not really a bug, its known that non english characters don't work but it is something that would be nice to fix.

arlomcwalter:
the font file isn't broken for anyone else so no it shouldn't be, it was clearly caused by something in your system or something you did

ThebestkillerTBK:
Custom font(non-english) works fine with caxton, but that mod is in dev
Just turn off custom fonts and load the resource pack.

arlomcwalter:
what

arlomcwalter:
@MineGame159 we should probably look into supporting the currently loaded language

PR #4713, opened 2024-07-02
Font rendering fix

Nippaku-Zanmu:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Modified Font.java to render characters in other languages (required to be included in the font file)
This helps globalize this client

## Related issues

# How Has This Been Tested?

Chinese
Japanese

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
+++ b/src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
@@ -1,10 +1,13 @@
+/*
+ * This file is part of the Meteor Client distribution (https://github.com/MeteorDevelopment/meteor-client).
+ * Copyright (c) Meteor Development.
+ */
 package meteordevelopment.meteorclient.renderer.text;
 
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
 import meteordevelopment.meteorclient.renderer.Mesh;
 import meteordevelopment.meteorclient.utils.render.ByteTexture;
 import meteordevelopment.meteorclient.utils.render.color.Color;
-import net.minecraft.client.texture.AbstractTexture;
 import org.lwjgl.BufferUtils;
 import org.lwjgl.stb.*;
 import org.lwjgl.system.MemoryStack;
@@ -26,7 +29,7 @@ public class Font {
     private final STBTTFontinfo fontInfo;
     private final ByteBuffer bitmap;
     private final STBTTPackContext packContext;
-    private final Map<Integer, STBTTPackedchar> packedChars = new HashMap<>();
+    private final Int2ObjectOpenHashMap<STBTTPackedchar> packedChars = new Int2ObjectOpenHashMap<>();
 
     public Font(ByteBuffer buffer, int height) {
         this.buffer = buffer;
--- a/src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
+++ b/src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
@@ -34,7 +34,7 @@ public class Font {
 
     private  long loadTimer = 0;
     private int loadCount=0;
-    private final int loadSpeedLimit = 9;
+    private final int loadSpeedLimit = 7;
     //The number of string that can be loaded per 100ms
 
     public Font(ByteBuffer buffer, int height) {
@@ -104,7 +104,7 @@ public class Font {
     }
 
     private void loadCharacter(int codePoint) {
-        if (packedChars.containsKey(codePoint)) return;
+        if (charMap.containsKey(codePoint)) return;
 
         STBTTPackedchar.Buffer cdata = STBTTPackedchar.create(1);
 
@@ -143,20 +143,16 @@ public class Font {
 
     public double getWidth(String string, int length) {
         double width = 0;
-        List<Integer> charPoints = null;
-
+        if (tryLoadString(string)) {
+            return width;
+        }
         for (int i = 0; i < length; i++) {
             int cp = string.charAt(i);
             CharData c = charMap.get(cp);
             if (c == null) {
-                if (charPoints == null) charPoints = new ArrayList<>();
-                charPoints.add(cp);
                 continue;
             }
             width += c.xAdvance;
-        }
-        if (charPoints != null) {
-            loadCharacter(charPoints);
         }
         return width;
     }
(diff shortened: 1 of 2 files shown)


src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
```diff
@@ -1,8 +1,3 @@
-/*
```
RacoonDog, line 1: dont delete the header

src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
```diff
@@ -16,50 +11,37 @@
...
     private final float scale;
     private final float ascent;
     private final Int2ObjectOpenHashMap<CharData> charMap = new Int2ObjectOpenHashMap<>();
     private final static int size = 2048;
 
+    private final ByteBuffer buffer;
+    private final STBTTFontinfo fontInfo;
+    private final ByteBuffer bitmap;
+    private final STBTTPackContext packContext;
+    private final Map<Integer, STBTTPackedchar> packedChars = new HashMap<>();
```
RacoonDog, line 29: use an `Int2ObjectOpenHashMap` like above

src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
```diff
@@ -72,29 +54,83 @@ public Font(ByteBuffer buffer, int height) {
...
+            packedChar.y0() * iph,
+            packedChar.x1() * ipw,
+            packedChar.y1() * iph,
+            packedChar.xadvance()
+        ));
+
+        packedChars.put(codePoint, packedChar);
+
+        // Re-create texture
+        createTexture();
```
RacoonDog, line 129: recreating the texture on every new character is incredibly wasteful if multiple new characters are present in a single string

src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
```diff
@@ -72,29 +54,83 @@ public Font(ByteBuffer buffer, int height) {
...
+    }
+
+    private void createTexture() {
+        texture = new ByteTexture(size, size, bitmap, ByteTexture.Format.A, ByteTexture.Filter.Linear, ByteTexture.Filter.Linear);
```
RacoonDog, line 133: `size` and `bitmap` are constants, there's not really any reason to create a new object rather than modifying `ByteTexture` to make the `upload` method accessible

src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
```diff
@@ -72,29 +54,83 @@ public Font(ByteBuffer buffer, int height) {
...
+        STBTTPackRange.Buffer packRange = STBTTPackRange.create(1);
+        packRange.put(STBTTPackRange.create().set(height, codePoint, null, 1, cdata, (byte) 2, (byte) 2));
+        packRange.flip();
+
+        // pack font ranges
+        STBTruetype.stbtt_PackFontRanges(packContext, buffer, 0, packRange);
+
+        STBTTPackedchar packedChar = cdata.get(0);
+
+        float ipw = 1f / size; // pixel width and height
```
RacoonDog, line 111: the whole `STBTTPackedchar` to `CharData` logic is duplicated from above and could instead be made into its own method and used in both locations

RacoonDog (changes requested):

Nippaku-Zanmu:
I limited the frequency of calls to the upload() function to avoid stalling when rendering a large number of unloaded characters at once
Maybe it would be better to use multithreading but it is beyond my ability and I have tried to use multithreading and it crashes my client every time

Nippaku-Zanmu:
Comparison of loading speed after restriction and before restriction

https://github.com/MeteorDevelopment/meteor-client/assets/120476813/79a988fe-fb62-4420-a2c5-6569557bcb32


https://github.com/MeteorDevelopment/meteor-client/assets/120476813/32aacb91-ed6a-45d5-a3ed-8130651d7d4b

RacoonDog:
> I limited the frequency of calls to the upload() function to avoid stalling when rendering a large number of unloaded characters at once

seems to work great performance-wise, but im not a fan of how it means invisible characters could be rendered for a few frames while its throttling, it'd be better to have a boolean that gets set to true once a new character is found, and that triggers a reupload of the texture right before rendering

a cool thing about that implementation is that we could bind the texture only once for uploading & rendering, but that'd be for later

> Maybe it would be better to use multithreading but it is beyond my ability and I have tried to use multithreading and it crashes my client every time

yes, multithreading is not doable in this situation as the uploading is a GL call, which has to be done on the render thread

src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
```diff
@@ -72,57 +63,120 @@ public Font(ByteBuffer buffer, int height) {
...
+        for (Integer codePoint : codePoints) {
+            loadCharacter(codePoint);
+        }
+        // Re-create texture
+        createTexture();
+        loadCount++;
+    }
+
+    private void loadCharacter(int codePoint) {
+        if (packedChars.containsKey(codePoint)) return;
```
RacoonDog, line 107: the `packedChars` map is redundant, you can replace this check here with `charMap.containsKey(codePoint)`

RacoonDog (changes requested):

Nippaku-Zanmu:
Modified the loading speed limit to make the loading characteristics more obvious (the upper limit is 1)
https://github.com/MeteorDevelopment/meteor-client/assets/120476813/b3b72394-1281-4231-9c0e-eb620e3bf5a9

Nippaku-Zanmu:
Modified the loading speed limit to make the loading characteristics more obvious (the upper limit is 1)

https://github.com/MeteorDevelopment/meteor-client/assets/120476813/8609d7f3-a566-4a4d-b952-7bc170428e3c

hexadecimal233:
\* Closes #4372 #333 #1854

src/main/java/meteordevelopment/meteorclient/renderer/text/Font.java
Nippaku-Zanmu: Is it possible to merge this project ? It's been 4 days without news

Nippaku-Zanmu:
@Wide-Cat 
Sorry to bother you, my PR request hasn't been processed for 6 days, please merge it if you can. Font rendering fixes are necessary for non-native English speakers, and no player likes to use a selection box without text hints

PR #4747, opened 2024-07-15
Added display of custom item sources in InventoryHud

Kirkezz:
## Type of change

- [x] New feature

## Description

Added ability to display items not only from inventory or container in hand, but also from custom sources in InventoryHud. Added incomplete support for inventories of arbitrary size (Texture and Outline backgrounds cannot be used for inventories with size other than 9x3).

To add a new source, you have to add code to enum SupportedSources to handle it.

As an example, the functionality was implemented for the Nametags module. When selecting the source Nametags, items in the rendering distance are displayed as a table, the size of which can be customized. The `getItems()` function was added, and a filter was also added to filter out items such as mushrooms, which are often a false indication of player activity.

## Related issues

# How Has This Been Tested?

meteor1
meteor2
meteor3

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/ESP.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/ESP.java
@@ -24,6 +24,7 @@ import meteordevelopment.orbit.EventHandler;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityType;
 import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.util.hit.EntityHitResult;
 import net.minecraft.util.math.Box;
 import net.minecraft.util.math.MathHelper;
 import org.joml.Vector3d;
@@ -40,6 +41,21 @@ public class ESP extends Module {
         .name("mode")
         .description("Rendering mode.")
         .defaultValue(Mode.Shader)
+        .build()
+    );
+
+    public final Setting<Boolean> highlightTarget = sgGeneral.add(new BoolSetting.Builder()
+        .name("highlight-target")
+        .description("highlights the currently targeted entity differently")
+        .defaultValue(false)
+        .build()
+    );
+
+    public final Setting<Boolean> targetHitbox = sgGeneral.add(new BoolSetting.Builder()
+        .name("target-hitbox")
+        .description("draw the hitbox of the target entity")
+        .defaultValue(true)
+        .visible(highlightTarget::get)
         .build()
     );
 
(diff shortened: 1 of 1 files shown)


TooFastBud:
ass

src/main/java/meteordevelopment/meteorclient/systems/hud/elements/InventoryHud.java
```diff
@@ -35,21 +39,76 @@ public class InventoryHud extends HudElement {
         .build()
     );
 
+    interface SupportedSourcesInterface { ArrayList<ItemStack> getItems(); }
```
RacoonDog, line 42: Interface is unnecessary, define the method in the enum instead. (Or modify the system to allow adding sources from third party addons, but that's more tricky)

src/main/java/meteordevelopment/meteorclient/systems/hud/elements/InventoryHud.java
```diff
@@ -35,21 +39,76 @@ public class InventoryHud extends HudElement {
...
+    private enum SupportedSources implements SupportedSourcesInterface
+    {
+        None {
+            @Override
+            public ArrayList<ItemStack> getItems() { return new ArrayList<>(); }
```
RacoonDog, line 47: `return List.of()`

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Nametags.java
```diff
@@ -708,4 +717,23 @@ public boolean excludeBots() {
     public boolean playerNametags() {
         return isActive() && entities.get().contains(EntityType.PLAYER);
     }
+
+    public ArrayList<ItemStack> getItems() {
+
+
```
RacoonDog, line 723: Remove this whitespace

RacoonDog (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/hud/elements/InventoryHud.java
Kirkezz: Added `public abstract ArrayList<ItemStack> getItems();` method. Removed an interface.

src/main/java/meteordevelopment/meteorclient/systems/hud/elements/InventoryHud.java
Kirkezz: Only works as `return new ArrayList<>(List.of())`.
```
error: incompatible types: no instance(s) of type variable(s) E exist so that List<E> conforms to ArrayList<ItemStack>
```

src/main/java/meteordevelopment/meteorclient/systems/modules/render/Nametags.java
Kirkezz: Done.

PR #4748, opened 2024-07-15
Rewrite `PacketUtils`

crosby-moe:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Removes `PacketUtilsUtils` in favor of filling the `PacketUtils` maps using the new `PlayStateFactories` vanilla registry.
Instead of being identified by their yarn class names, the new packet names follow the following convention: `namespace:Side/UpperCamelCaseName`. The namespace `minecraft` is implicit, mapping vanilla packets will return a name fitting `Side/UpperCamelCaseName`, however you can still unmap vanilla packets with and without the namespace.

To match vanilla use cases, `PacketUtils` and `PacketListSetting` methods that used to take `Class<? extends Packet<?>>` parameters now take `PacketType<? extends Packet<?>>` instead.

Since we are effectively moving away from packets named with yarn mappings to official identifiers, this means that existing `PacketListSetting` saved configs will be reset, and some packets may appear under completely different names.

# How Has This Been Tested?

Works greate

# Checklist:


Gamekind135:
can you update Tokyo to 1.21,
Please

RacoonDog:
> can you update Tokyo to 1.21, Please

no and this is not the place to ask

PR #4761, opened 2024-07-20
Consistently picks up Powder Snow after fall

AnandBhasme:
## Type of change

- [✔] Bug fix
- [] New feature

## Description

Earlier, player wasn't picking up powder snow if it was wearing leather boots. Now it checks whether it has powder snow below it.

## Related issues

Not aware of any. I checked the issues

# How Has This Been Tested?

Tried in various cases. Consistently successful 

# Checklist:

- [✔] My code follows the style guidelines of this project.
- [✔] I have added comments to my code in more complex areas.
- [✔] I have tested the code in both development and production environments.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoLog.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoLog.java
@@ -22,7 +22,6 @@ import meteordevelopment.orbit.EventHandler;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.decoration.EndCrystalEntity;
 import net.minecraft.entity.player.PlayerEntity;
-import net.minecraft.entity.vehicle.TntMinecartEntity;
 import net.minecraft.network.packet.s2c.common.DisconnectS2CPacket;
 import net.minecraft.text.MutableText;
 import net.minecraft.text.Text;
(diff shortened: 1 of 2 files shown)


MineGame159:
Ok well I tried the PR and it seems to work in good conditions, but I have some concers.
1. Isn't it possible for it to pick up the block before the player actually lands? It might be good to only pick it up after the player is touching the ground.
2. Also it might be good to only check the block below if the target is powder snow. It might be possible to pick up for example a water source block before the players lands in it.
Again, none of those things happened in my singleplayer testing, I am just wondering if they could happen in for example worse netwrok conditions on a server. Since in theory they are possible.

MineGame159:
Revert that Auto Log commit.
1. Don't do multiple features in a single PR.
2. Hardcoding that feature for just TNT minecarts is stupid. It should be an entity type list setting allowing the user to choose which entity to check for.

AnandBhasme:
1. Thats what I asked on Discord and you said to just Push. But alright I will remember from next time.
2. Will the entity type list be holistic or only the harmful entities? Will the setting for number of entities and range be different for each or common to all?

AnandBhasme:
> 1. Don't do multiple features in a single PR.
       Thats what I asked on Discord and you said to just Push. But alright I will remember from next time.
> 2. Hardcoding that feature for just TNT minecarts is stupid. It should be an entity type list setting allowing the user to choose which entity to check for.
       Will the entity type list be holistic or only the harmful entities? Will the setting for number of entities and range be different for each or common to all?

MineGame159:
Oh sorry, I didn't bother looking up the issue number you sent in the discord message and just assumed you meant the changes in response to my first comment on this PR.

As for the entity list. I would not limit it to harmful entities only, if the user wants to log on a pig entering the render distance, let them. And for the range and count, I think it's enough if they are global for all entities combined.

Issue #4764, opened 2024-07-20
[Suggestion] Autolog TNT minecart nearby


satyr2b:
### Describe the feature

basically, autolog when a variable amount of TNT mine carts (or other entities) are within render distance

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

AnandBhasme:
I added the feature. Hopefully it gets approved :)

PR #4766, opened 2024-07-22
Added feature #4764 with universal enitity selection

AnandBhasme:
## Type of change

- [] Bug fix
- [ ✔] New feature

## Description

Added the feature requested in #4764 with universal entity selection. The new feature allows player to select any entity(or entities) from list to monitor. When they come close to player in variable distance and variable number the player automatically gets disconnected.

## Related issues

#4764 

# How Has This Been Tested?

Tested in flat World with restrictive conditions using various mobs and entities including but not limited to pigs, end crystal, TNT minecart etc.

# Checklist:

- [✔] My code follows the style guidelines of this project.
- [✔] I have added comments to my code in more complex areas.
- [✔] I have tested the code in both development and production environments.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoLog.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/AutoLog.java
@@ -5,9 +5,14 @@
 
 package meteordevelopment.meteorclient.systems.modules.misc;
 
+import java.util.HashMap;
+import java.util.Set;
+import java.util.Map;
+
 import meteordevelopment.meteorclient.MeteorClient;
 import meteordevelopment.meteorclient.events.world.TickEvent;
 import meteordevelopment.meteorclient.settings.BoolSetting;
+import meteordevelopment.meteorclient.settings.EntityTypeListSetting;
 import meteordevelopment.meteorclient.settings.IntSetting;
 import meteordevelopment.meteorclient.settings.Setting;
 import meteordevelopment.meteorclient.settings.SettingGroup;
@@ -20,6 +25,7 @@ import meteordevelopment.meteorclient.utils.entity.DamageUtils;
 import meteordevelopment.meteorclient.utils.player.PlayerUtils;
 import meteordevelopment.orbit.EventHandler;
 import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityType;
 import net.minecraft.entity.decoration.EndCrystalEntity;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.network.packet.s2c.common.DisconnectS2CPacket;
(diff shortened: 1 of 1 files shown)


MineGame159:
1. Please follow the formatting used in the rest of the project, for example you wrote the fields for the new settings starting with an uppercase letter.
2. Fix the names of the settings, `Total Count` should be `total-count`. Meteor automatically formats it to the "nicer" format when rendering in the UI.
3. Since there are quite a bit of settings all relate to the new entity detector feature I think it would be nice to put them into a new group. To nicely separate them in the UI.
4. Why is the maximum number of entities limited to 96? And the range to 128?
5. You added 2 count settings depending on if the count is combined or separate. In my opinion it would be better to simply have a single count setting whose meaning changes on the boolean setting. Not sure on this one tho, if you or someone else thinks differently, comment on it.
6. You are calculating the separate entity counts all the time, even if the users wants the count to be combined. This creates quite a lot of overhead. Also you are creating a new hash map every tick for it. It would be better to instead make a `private final Object2IntMap<EntityType<?>> entityCounts = new ...();` field and clear the map every tick.

I think those are my main points I would like addressed. I have a nit-pick tho, but you don't really need to worry that much about it, I just like when things are consistent across the code base. I noticed you put all your comments on the same line. The rest of the project puts the comments on separate lines, mostly above the code. I think it makes it easier to read since it doesn't create ultra long lines that overflow the display.

AnandBhasme:
I implemented most of your suggestions. About the max limit on entities, I removed it but I kept the limit on range. I set that according to 8 chunks which is the render distance most servers have. I can change it if you want. 

I gave the separate integer setting a lot of thought. Although it causes redundancy and more settings in the UI part but its pros outweighs the cons. Not only it reduces the complexity of logic needed in the code but also it provides clearer intent with a bit more context, consequently serving a better User Experience. Still I am open to change if anyone wishes otherwise

I want to thank you as it is nearly my first time coding in a collaborative way among such brilliant minds. I am getting to know and learn so much of coding which impacts real life activities. Thanks <3

MineGame159:
Ok yeah, thinking about it again I guess it does make sense to keep the count settings separated. One issue tho. You have two settings with the name `total-count`, that's going to cause issues. I should probably make it throw an exception so people don't make the mistake.

About the maximum values. You have replaced the maximum entity counts with `Integer.MAX_VALUE`. You can simply replace the whole method call with `.min(...)`. And about the range maximum value, I would also remove that. Servers can very well increase the distance to more than 8 chunks and I don't see a reason to limit it. Generally we only limit settings if going over / below the limit causes issues.

One last thing, I would put the `entityCounts.put(...);` call inside an if statement. No need to fill up the map if it is only concerned with total count.

Otherwise I think it looks good. After the above changes I am going to test it out and then merge it.

AnandBhasme:
I changed the name of one of the settings

MineGame159:
So I tested it, it seems to work fine, but. I promise, the last change I will ask for. We talked it over in Discord (private channel) and came up with better names and descriptions for the settings. At least we think they are better, you can give your opinion.

`entities` - Disconnects when a specified entity is present within a specified range.
`use-total-count` - Toggle between counting the total number of all selected entities or each entity individually.
`combined-entity-threshold` - The minimum total number of selected entities that must be near you before disconnection occurs.
`individual-entity-threshold` - The minimum number of entities individually that must be near you before disconnection occurs.

Of course don't forget to also rename the Java fields.

AnandBhasme:
The new names, while clear and descriptive, are somewhat lengthy.

Issue #4765, opened 2024-07-21
Trajectories hand preview option


lc4ro:
### Describe the feature

add a option for trajectories module for showing and calculating the projectile trajectory of arrows while having a bow in your hand still

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

lc4ro:
i think a fully charged trajectory would be the best, when was last time you shot a partially charged bow?

AnandBhasme:
Is it needed for crossbow too? You know it can be charged but you don't have to hold it

PR #4772, opened 2024-07-24
Pre Renders Bow #4765

AnandBhasme:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds option to Pre renders the projectile of arrow from bow.

## Related issues
#4765 

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Trajectories.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Trajectories.java
@@ -46,6 +46,14 @@ public class Trajectories extends Module {
         .build()
     );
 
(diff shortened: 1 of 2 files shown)


AnandBhasme:
I was wondering whether #4773 can be fixed too while I am working on this module. I tested and I am able to fetch the nbt tag of flight duration of rocket present in the crossbow. 
But as per https://minecraft.fandom.com/wiki/Crossbow
Screenshot 2024-07-25 003653
the distance travelled by the rocket is random in a range. 
I reckon we can make render it such that the render line till minimum (of range) will be same as now (a solid line) and the distance of range will have a dotted line and at end something that represents a explosion.

RacoonDog:
I dont believe meteor client has the capabilities to render dotted lines, though we could render a full line up until the minimum, and then fade out the opacity as it reaches the maximum.

PR #4831, opened 2024-08-16
Kotlin Gradle

RickyTheRacc:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Changes the main project and subproject to use the Kotlin Gradle DSL, slightly faster build times and an overall easier to read file

## Related issues

N/A

# How Has This Been Tested?

Worked in both the runClient fabric task as well as the default Minecraft Client configuration by the mcdev plugin, built jar works when tested on a new Prism 1.21.1 instance

# Checklist:


GreemDev:
based pr tbh

Lncvrt:
> slightly faster build times and an overall easier to read file

this is completely pointless lmfao

RickyTheRacc:
> this is completely pointless lmfao

Small upsides, maybe, but not pointless? The build times are faster, and Kotlin gives more robust errors and is often easier to comprehend for new devs who already started on Java because its syntax is much stricter. If you have any legit concerns though I'm glad to listen ^^

RacoonDog:
> > slightly faster build times and an overall easier to read file
> 
> this is completely pointless lmfao

anything that removes groovy dsl will never be pointless

build.gradle.kts
```diff
@@ -0,0 +1,188 @@
...
+
+        inputs.properties(propertiesMap)
+        filesMatching("fabric.mod.json") {
+            expand(propertiesMap)
+        }
+    }
+
+    jar {
+        from("LICENSE") {
+            rename{
```
RacoonDog, line 113: add a space (`rename {`) and make into single line, like in the `shadowJar` task

RacoonDog (changes requested):

RickyTheRacc:
Updated all versions and made the thingies oneliners as crobie asked

Lncvrt:
Why was this canceled?

Issue #4912, opened 2024-09-24
Better Tooltips causes crash when using the plugin Advanced Kits


jharrison123:
### Steps to reproduce

So after some tinkering i found out that its Better Tooltips causing the crash
disabling the module prevents the crash

### Link to crash report/log (upload to https://mclo.gs and paste resulting link here)

https://mclo.gs/RskY8ZD

### Meteor Version

meteor client v0.5.8 build #2172

### Minecraft Version

1.21.1

### Operating System

Linux

### Before submitting a crash report

- [X] This crash wasn't already reported (I have searched crash reports on GitHub).

- [X] This is a valid crash (I am able to reproduce this on the latest dev build).

RacoonDog:
crash caused by `ItemStack#getTooltip` when an item has `DataComponentTypes.HIDE_TOOLTIP` since it short-circuits and returns `List.of()`, which is immutable

PR #4913, opened 2024-09-24
Fix BetterTooltips Crash

0xTas:
## Type of change

- [x ] Bug fix
- [ ] New feature

## Description

I modified ItemStackMixin#onGetTooltip to return an empty ArrayList instead of an immutable list when there is no tooltip.

This prevents a crash when BetterTooltips attempts to add to that immutable list.

## Related issues

https://github.com/MeteorDevelopment/meteor-client/issues/4912

# How Has This Been Tested?

It hasn't.

# Checklist:

- [x ] My code follows the style guidelines of this project.
- [x ] I have added comments to my code in more complex areas.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/events/game/ItemStackTooltipEvent.java
+++ b/src/main/java/meteordevelopment/meteorclient/events/game/ItemStackTooltipEvent.java
@@ -5,9 +5,55 @@
 
 package meteordevelopment.meteorclient.events.game;
 
+import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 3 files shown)


0xTas:
In the last two commits, I refactored the ItemStackTooltipEvent so that other modules using the event do not need to manually implement the previous changes in their event handler in order to prevent this crash. They can instead elect use the new 'appendStart' and 'appendLast' methods built into the event class to safely insert elements into the tooltip list.

I tested these changes on a 1.21 paper server and could not replicate the crash in the original issue.
These changes also shouldn't break any pre-existing users of the ItemStackToolTipEvent because no existing methods were changed.

Thanks to @RacoonDog for the idea.

PR #4917, opened 2024-09-27
Allow using a different command than /msg for message aura

Siriusmart:
## Type of change

Improvement of existing feature.

## Description

Some server (specifically constantiam) use a different command for private messages than /msg (specifically /w), I added a command option so it can be changed to work on constantiam.

The default option is still /msg, so nothing will break when updated.

# How Has This Been Tested?

I went to spawn to get some players in my render distance, they took my stuff and killed my horse, but the message aura works.

Also would like to bring to attention, when I connect to the server where my location already have players in it, it will try to message all players at once and get me kicked for spamming, adding a delay/per player cooldown would be a good idea for the future.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/macros/Macro.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/macros/Macro.java
@@ -60,26 +60,26 @@ public class Macro implements ISerializable<Macro> {
     }
 
     public boolean onAction() {
-            if (dirty) {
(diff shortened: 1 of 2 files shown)


OrangeMan432:
IMG_0514

Wide-Cat:
Message `crosby.moe` on discord in regards to making it use starscript, he's working on a variable scope system for this.

Issue #4054, opened 2023-09-07
[Suggestion] Allow the account manager to support more than 3 accounts


terrysgg:
### Describe the feature

When you load a 4th, 5th, 6th alt account into the account manager - it deletes every account after the 3rd one when Minecraft is restarted

IDK if this one is a "Feature" or a bug with the account manager

But I saw lots of posts on the discord about this issue

Would love to see this fixed / improved - as account manager is also something Rusherhack currently does not have in their 1.20 version

3 accounts is currently kinda a weird limitation

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

terrysgg:
Anyone know if this one has been looked into?

Would love to add more than 3 accounts to the account manager

spazmilk:
Bump? how has this not been fixed yet.

terrysgg:
I've tested this change in the latest version of Meteor for 1.20.6, and Im guessing its not in the release yet, because its still occurring in 0.5.7

Do you know what version this code will make it into the game?

PR #4949, opened 2024-10-12
Fix the alt manager's limit of 3 Microsoft accounts.

0xTas:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

This pr fixes the alt manager's incidental limit of 3 Microsoft accounts, which was occurring due to an API ratelimit during the process of loading accounts from disk. I accomplished this by moving usage of the fetchInfo() method out of Accounts#fromTag and relocating it to instead run when an account is actually logged into.

This prevents accounts beyond the 3rd from being thanos-snapped off the account list due to ratelimits on startup. You can still be ratelimited if you're switching between accounts too quickly, but if this happens just be patient, wait 30 seconds, try again and it will work.

## Related issues

Closes https://github.com/MeteorDevelopment/meteor-client/issues/4054

# How Has This Been Tested?

This was tested by adding more than 3 Microsoft accounts to the alt manager, closing and reopening the client, and then logging into and joining a server with each account in turn, to ensure the feature was working as expected. As a result of these changes, the alt manager can now support more than 3 accounts, which successfully persist through restarts.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/widgets/WAccount.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/widgets/WAccount.java
@@ -58,7 +58,7 @@ public abstract class WAccount extends WHorizontalList {
             screen.locked = true;
 
             MeteorExecutor.execute(() -> {
-                if (account.login()) {
+                if (account.fetchInfo() && account.login()) {
                     name.set(account.getUsername());
 
                     Accounts.get().save();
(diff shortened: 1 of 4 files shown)


Wide-Cat (changes requested):
Calling `fetchInfo` and then `login` for a microsoft account results in the `auth` method getting called twice in a row, which not only causes the login process to slow down dramatically but you end up getting rate limited if you log into one account and then try to log into another within 30 seconds. It's also probably problematic for altening accounts because code is duplicated between the two methods as well. This needs to be changed to prevent duplicate code execution between the two methods.

0xTas:
Got it, I'll take another crack at it when I get some time.

0xTas:
How does this look? I tested with Microsoft accounts and the slow login speeds and ratelimit issues seem to be fixed.
However, I did not test with any altening accounts.

Wide-Cat (approved):

Issue #1394, opened 2021-08-21
[Suggestion] Make zoom not make waypoints smaller


TheRedHippo:
**Describe the feature**
Make zoom not make waypoints smaller. When you zoom in, waypoints and nametags remain the same size despite everything else zooming in.

tyrannus00:
This happens with any zoom mod btw, so it's not a zoom issue

RedCarlos26:
Duplicate of #1150

RedCarlos26:
@seasnail8169 ^

PR #4979, opened 2024-10-23
Increase nametage scale when using Zoom

chairclr:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

When using the Zoom module before, Nametags stayed at the same size and were unreadable when looking at players in the distance without explicitly increasing Nametage scale. This also includes a renaming of the `Zoom::value` function to `getScaling` and exposes it as a public method.

## Related issues

None (maybe #1394)

# How Has This Been Tested?

Zoomed out:

image

Zoomed in:

image

# Checklist:


ppebb:
lgtm (let's get this merged)!

Wide-Cat:
Might be worth adding a config option in the future to set the text quality now. Maybe something to do in the future.

nickcat325:
Very nice. I'll try this out.

PR #4981, opened 2024-10-25
feat: Baritone Settings - Max values

RuriYS:
## Type of change

- [x] Bug fix (maybe)
- [x] New feature

## Description

Some of Baritone's settings needs to have max values, or it will cause unexpected errors,
So I added `maxValue` for specific setting(s).

### Affected settings:
  - `pathCutoffFactor`
  - might be even more...

## Related issues

# How Has This Been Tested?

When pathing with Baritone, setting value of `pathCutoffFactor` above 1.0 would cause `index out of bounds` exceptions.

# Checklist:


RuriYS:
figured I don't need to touch [`DoubleSetting`](https://github.com/MeteorDevelopment/meteor-client/tree/42fb9083c306fd7ba4fdf8348e5729a7e5243782/src/main/java/meteordevelopment/meteorclient/settings/DoubleSetting.java), mb

Wide-Cat:
Currently this sets `sliderMax`, should we keep it as that or set the `max` value? I'm open to hearing thoughts either way. Also I don't think baritone caps this value either, might be worth opening a bug report on their repo if you haven't already.

RuriYS:
> Currently this sets `sliderMax`, should we keep it as that or set the `max` value?

Setting the `max` value is a great idea

> Also I don't think baritone caps this value either, might be worth opening a bug report on their repo

I might, and unless my PR would be approved there then this is just a fix in meteor itself.

PR #5069, opened 2024-12-28
Use modules instead of moduleInstances.values() to access modules

PoTheMagicDragon:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description
A rework of #5026 due to mangled git history

## Related issues
N/A

# How Has This Been Tested?
Running an instance and making sure that modules load correctly

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
@@ -134,7 +134,7 @@ public class Modules extends System<Modules> {
     }
 
     public Module get(String name) {
-        for (Module module : moduleInstances.values()) {
+        for (Module module : modules) {
             if (module.name.equalsIgnoreCase(name)) return module;
         }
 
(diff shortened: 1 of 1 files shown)


RacoonDog:
rather than calling
```java
if (modules.removeIf(module1 -> {
    if (module1.name.equals(module.name)) ...
})) {
    getGroup(...).remove(...);
}

if (moduleInstances.values().removeIf(module1 -> {
    if (module1.name.equals(module.name)) ...
})) {
    getGroup(...).remove(...);
}
```

it should be safe to assume that, since the `moduleInstances` map *should* always be a subset of `modules`, we can simplify it as so:
```java
AtomicReference<Module> removedModule = new AtomicReference<>();
if (modules.removeIf(module1 -> {
    if (module1.name.equals(module.name)) {
        removedModule.set(module1);
        module1.settings.unregisterColorSettings();

        return true;
    }

    return false;
})) {
    moduleInstances.remove(removedModule.getClass(), removedModule);
    getGroup(removedModule.get().category).remove(removedModule.get());
}
```

PoTheMagicDragon:
Good point. I copied what you suggested, but changed
```java
    moduleInstances.remove(removedModule.getClass(), removedModule);
```
to
```java
    moduleInstances.remove(removedModule.getClass(), removedModule.get());
```

PR #5077, opened 2025-01-01
Show HUD In Editor

machiecodes:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Breaking change moment, added a boolean `showHUDInEditor()` to GuiTheme class for themes to implement, allowing the minecraft hud to always be shown in the meteor hud editor regardless of whether hide hud is on.

## Related issues

N/A

# How Has This Been Tested?

Opened in singleplayer world 👍

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/MeteorClient.java
+++ b/src/main/java/meteordevelopment/meteorclient/MeteorClient.java
@@ -16,6 +16,7 @@ import meteordevelopment.meteorclient.gui.WidgetScreen;
 import meteordevelopment.meteorclient.gui.tabs.Tabs;
 import meteordevelopment.meteorclient.systems.Systems;
 import meteordevelopment.meteorclient.systems.config.Config;
+import meteordevelopment.meteorclient.systems.hud.screens.HudEditorScreen;
 import meteordevelopment.meteorclient.systems.modules.Categories;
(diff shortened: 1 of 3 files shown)


src/main/java/meteordevelopment/meteorclient/MeteorClient.java
```diff
@@ -182,6 +183,10 @@ private void onOpenScreen(OpenScreenEvent event) {
         }
 
         wasWidgetScreen = event.screen instanceof WidgetScreen;
+
+        if (!GuiThemes.get().showHUDInEditor()) return;
+        if (!(event.screen instanceof HudEditorScreen)) return;
+        mc.options.hudHidden = false;
```
RacoonDog, line 189: guard statements here would make it annoying to add onto this event handler, id rather this be a
```java
if (GuiThemes.get().showHUDInEditor() && event.screen instanceof HudEditorScreen) {
    mc.options.hudHidden = false;
}
```

RacoonDog:
i'm also unsure whether this should be part of a theme at all or just a setting in the meteor client Config system, considering it doesn't affect anything in the meteor widget rendering

tyrannus00:
This implementation doesn't toggle the hud off again, after leaving the editor

Wide-Cat:
as tyr said

PR #5080, opened 2025-01-02
Zoom Hide Hud Setting

machiecodes:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds a setting to hide the hud while the module is active. Activating this hides the "show hands" setting since that logic is encapsulated in hud rendering. If the user manually turns the hud on/off while the module is active it will not try to retoggle the hud when deactivated.

## Related issues

N/A

# How Has This Been Tested?

https://github.com/user-attachments/assets/7f2ea8c0-e652-4f74-be5d-6543459d4712

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Zoom.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Zoom.java
@@ -116,12 +116,6 @@ public class Zoom extends Module {
         hudManualToggled = true;
     }
 
-    @EventHandler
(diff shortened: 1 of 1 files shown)


Wide-Cat:
the mouse button event can probably be removed, i dont think its possible for a mouse to call GLFW_KEY_F1 and for it to show up in that event

machiecodes:
Fair enough, I was just thinking of like razer software stuff that lets you rebind keys to diff things but I also guess people wouldn't be rebinding hide hud to anything without a mod lol

Issue #3944, opened 2023-08-12
[Suggestion] 


datcathuh:
### Describe the feature

in antiafk make it so it spins in random yaw not a set 1

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

FlightXamb:
you really want spinbot lol

Issue #2175, opened 2022-03-21
[Suggestion] A whole load of improvements for a few different modules


Funtimes909:
### Describe the feature

Chat Highlight, Highlights certain words (can also be configured to highlight usernames) in chat with a configurable color

Auto Inventory Module, Automatically puts certain items into certain slots of your inventory

Mass TPA module (Thanks Wurst!), Sends and cancels mass TPA requests to other players online, Can also be access via /mass-tpa "username" cancel "true/false" delay in ticks "20"

Twerk Module, Auto Twerks at a configurable speed (Kinda like anti-afk but just twerking!)

Auto disable option for all auras that can be set to disable after being kicked/disconnecting or after a certain time period

You can now set a delay for Auto Respawn

Limit feature for tracers, Similar to the culling feature in the nametags module, allows you to set a strict amount of tracers that can be rendered

Auto disable option to disable spammer module after being kicked/disconnected

Combat info can now display health and armor durability in numbers or percentages

Combat info can also display item names

Blink will now render a player model with your skin where you are standing and will disappear once it is disabled

I am aware this a lot of things to add so sorry if someone of them already exist in some addons or other modules

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client.
(I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested.
(I have checked every suggestion on github)

- [X] This feature is a valid suggestion.
(I have read the FAQs and Guides)

RickyTheRacc:
picrew pfp opinions ignored

C10udburst:
why is this written like a pull request when it is not a oull request

I-Got-You-Dead:
E

SnivyLlama:
> why is this written like a pull request when it is not a pull request

ikr, they even say stuff like "you can now"

Funtimes909:
> why is this written like a pull request when it is not a pull request

I was originally going to add this to my own addon and just copy pasted the ideas list into the suggestion

SIMULATAN:
A few great ideas, I'll look into it on the weekend if nobody else does so

I-Got-You-Dead:
> 

Can’t wait to see some of this 😄

DesiCow:
Instead of mass tpa command, add {players.all} placeholder for Starscript.
In macros we could type `tpa {players.all}` for example and it will automatically parse this placeholder for all online players in server and will execute the command on all those players.

arlomcwalter:
[...]

PR #5082, opened 2025-01-04
AntiAFK Look/Follow/Wander modes and general improvements

0xTas:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

**EDIT - Context:**
AntiAFK currently performs its actions on a random delay that is not configurable, which is unnecessarily restrictive and annoying.<br>
For example, on 2b2t you must only send a single hand swing packet every ~15 minutes to avoid being kicked for AFK. This means that with the ability to configure your swing delay, you could leave AntiAFK on 24/7 without having to notice or think about it, and still never get kicked for being afk. With the current version of Meteor's AntiAFK, however, your only option is to watch your player throw hands every few seconds at all times, or otherwise explicitly enable/disable the module when you "plan" on going AFK for a bit.

I do understand that some people might like/want/need the random delay so I have left it as the default behavior for most actions that already used it. I have also added a fun Look mode with some simple but satisfying-to-use follow/wander modes.

### Additions

- Look mode which rotates randomly or tracks an entity based on the sub-mode settings.
- Added random yaw speed setting to spin mode and strafe duration option to strafe mode.
- Follow & wander sub-settings which follow the targeted entity or wander in random directions.
- Added timer settings to jump/swing/sneak/strafe modes, while preserving random time capabilities.

### Fixes

- Changed module package to reflect its category (movement -> player).
- Made jump and sneak work properly while the their module modes are active (you could not jump when jump was enabled).

## Related issues

Probably #3944
Twerking for #2175 

# How Has This Been Tested?
In singleplayer and on 2b2t.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AntiAFK.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AntiAFK.java
@@ -3,13 +3,14 @@
  * Copyright (c) Meteor Development.
  */
 
-package meteordevelopment.meteorclient.systems.modules.movement;
+package meteordevelopment.meteorclient.systems.modules.player;
 
 import meteordevelopment.meteorclient.events.world.TickEvent;
 import meteordevelopment.meteorclient.settings.*;
 import meteordevelopment.meteorclient.systems.modules.Categories;
 import meteordevelopment.meteorclient.systems.modules.Module;
 import meteordevelopment.meteorclient.utils.Utils;
+import meteordevelopment.meteorclient.utils.misc.input.Input;
 import meteordevelopment.meteorclient.utils.player.ChatUtils;
 import meteordevelopment.meteorclient.utils.player.Rotations;
 import meteordevelopment.orbit.EventHandler;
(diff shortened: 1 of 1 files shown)


RacoonDog (changes requested):
no point in having `jumpRate`, `swingRate`, `sneakRate`, `strafeRate`, and `lookRate` having a minimum of zero when they also have a boolean setting toggling them

setting the minimums to 1 would also remove several `if (xRate.get() > 0) {` conditions everywhere

RacoonDog:
fyi, i'm unsure if this is intentional or not, but `Random#nextInt`'s upper bound is exclusive, so `random.nextInt(99) == 0` is a 1/99 chance, not a 1/100

0xTas:
> fyi, i'm unsure if this is intentional or not, but `Random#nextInt`'s upper bound is exclusive, so `random.nextInt(99) == 0` is a 1/99 chance, not a 1/100

Yes this was intentional, I mostly just kept the bounds used in the previous version of the Module for the random fallbacks.

> no point in having `jumpRate`, `swingRate`, `sneakRate`, `strafeRate`, and `lookRate` having a minimum of zero when they also have a boolean setting toggling them
> 
> setting the minimums to 1 would also remove several `if (xRate.get() > 0) {` conditions everywhere

I should probably have clarified the main motivation of this PR a bit better. AntiAFK currently performs its actions on a random delay that cannot be configured, and is quite fast (`random.nextInt(99) == 0`). However, on some servers such as 2b2t, in order to bypass the AntiAFK plugin you must only send a single hand swing packet every 15 minutes or so.

Using Meteor's AntiAFK is annoying because it has me constantly swinging which is an eyesore imo, so I need to consciously enable/disable AntiAFK when I want to go afk. On the other hand, with something like Rusherhack's AntiAFK I can set the swing delay manually, leave the module on 24/7, not even notice it, and also never get kicked.

So the `xRage.get() > 0` conditions are my way of allowing the module's previous behavior to still exist (and exist by default) while still allowing people who wish to configure their action delays to do so.

The look mode(s) were just something I had fun throwing together on-the-spot while I was tinkering.

PR #5091, opened 2025-01-07
Rewrite build script

RacoonDog:
## Overview
Rewrite build script to make use of Kotlin DSL

- Clear up deprecation warnings
- Update dependencies
- Enable configuration cache
- Selectively enable javadocs & sources jar creation

Javadoc & sources jars are now only created if the `CI` system property evaluates to true. This was done because it used to take up more than half of the build time despite being rarely used.

## Credits
- @machiecodes Initial rewrite to Kotlin DSL
- @FlorianMichael @Kichura [BaseProject](https://github.com/FlorianMichael/BaseProject) from which I took heavy inspiration from
- [The zizmor project](https://github.com/woodruffw/zizmor) GitHub Actions static analysis tool


.github/workflows/build.yml
```diff
@@ -14,14 +14,20 @@ jobs:
     runs-on: ubuntu-latest
```
Kichura, line 14: Set `ubuntu-latest` to 22.04 or 24.04 so that GH Actions doesn't alert repo owner about sudden OS change.

.github/workflows/build.yml
```diff
@@ -14,14 +14,20 @@ jobs:
...
+      - name: Set up Java
+        uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
           java-version: '21'
 
-      - run: chmod +x gradlew
-
       - name: Set up Node JS
         uses: actions/setup-node@v3
```
Kichura, line 32: Must be `v4`.

.github/workflows/issue-moderator.yml
```diff
@@ -11,7 +11,11 @@ jobs:
   issue-moderator:
     runs-on: ubuntu-latest
```
Kichura, line 12: Ubuntu statement from build section applies here.

Kichura (changes requested):
Noticed a few things.

Kichura:
Also one more thing: `DEFAULT_JVM_OPTS` in the gradlew files must include `-Dfile.encoding=UTF-8`.

.github/workflows/build.yml
RacoonDog: will be done after this pull request as minegame will be reworking the publishing process

.github/workflows/build.yml
RacoonDog: will be done after this pull request as minegame will be reworking the publishing process

.github/workflows/issue-moderator.yml
RacoonDog: will be done after this pull request as minegame will be reworking the publishing process

RacoonDog:
> Also one more thing: `DEFAULT_JVM_OPTS` in the gradlew files must include `-Dfile.encoding=UTF-8`.

i looked into it and it seems like i can't change these without the changes being undone whenever we update the wrapper scripts through the gradle `wrapper` task
and the documentation for configurating the task within our `build.gradle.kts` doesn't seem to support changing these either

PR #5114, opened 2025-01-19
Fixed typo at line 63 Flight.java

ltsnumix:
## Type of change

Typo fix

## Description

Fixed typo. Changed "anti kick" to "anti-kick"

## Related issues

More readable description for the module Flight.


MineGame159:
If anything it should be `Anti Kick` since that is how it is displayed in the GUI.

PR #5155, opened 2025-01-25
Add Freecam Baritone click (similar to Kami Blue)

ThrowawayPR:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

When this setting is enabled, clicking on a block while Freecam is on will call IPathManager::moveTo() to that block's face. The reasoning is that #come seems to be incompatible with Meteor's Freecam and I find the click method more convenient anyway.

# How Has This Been Tested?

Intellij dev environment and my 2b2t Prism instance, both with the Baritone API jar.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Freecam.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Freecam.java
@@ -14,6 +14,7 @@ import meteordevelopment.meteorclient.events.meteor.MouseScrollEvent;
 import meteordevelopment.meteorclient.events.packets.PacketEvent;
 import meteordevelopment.meteorclient.events.world.ChunkOcclusionEvent;
 import meteordevelopment.meteorclient.events.world.TickEvent;
+import meteordevelopment.meteorclient.pathing.PathManagers;
 import meteordevelopment.meteorclient.settings.BoolSetting;
 import meteordevelopment.meteorclient.settings.DoubleSetting;
 import meteordevelopment.meteorclient.settings.Setting;
@@ -106,6 +107,13 @@ public class Freecam extends Module {
         .name("static")
         .description("Disables settings that move the view.")
         .defaultValue(true)
+        .build()
+    );
+
+    private final Setting<Boolean> baritoneClick = sgGeneral.add(new BoolSetting.Builder()
+        .name("click-sets-goal")
+        .description("Sets a goal to any block you click.")
+        .defaultValue(false)
         .build()
     );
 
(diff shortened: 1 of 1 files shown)


0xTas:
What if I want to path to a block that isn't in range of my crosshair target? Wouldn't this be nicer to use if you didn't have to Freecam up-close to a block in order to be able to path to it?

What if I click on an entity? Is there any reason you shouldn't want to set the Baritone goal to that entity's position? Right now it ignores everything that isn't a direct block-hit.

Also, could we consider adding an option to require a double-click for the goal to be updated? Clicking once in Freecam can easily happen accidentally, or even completely on purpose, since there are other legitimate reasons to click/hold in Freecam like breaking blocks outside the player's field-of-view. Leaving this setting on would disrupt all of those actions/behaviors that people are likely already used to, but having an option to require a quick double-click instead would prevent that disruption.

I love this idea which is why I implemented it in my addon Stardust, but I do think if we're going to add it to the client proper then it should probably be as user-friendly & polished as possible.

ThrowawayPR:
That's fair, so far I just made it behave like Kami Blue did as that's what I'm used to but what you suggested could be nice additions. I will look into adding them, though wrt double clicks specifically I'm not sure if that's really necessary since it's already toggleable with a setting (though you could be right).

Nekiplay:
Its has in Meteor+ https://github.com/MeteorClientPlus/MeteorPlus

0xTas:
> That's fair, so far I just made it behave like Kami Blue did (except for not setting the goal inside the target block) as that's what I'm used to but what you suggested could be nice additions. I will look into adding them, though wrt double clicks specifically I'm not sure if that's really necessary since it's already toggleable with a setting (though you could be right).

The double-click setting is something I added in [my own implementation](https://github.com/0xTas/stardust/blob/1.21.1/src/main/java/dev/stardust/mixin/meteor/FreecamMixin.java) because I personally don't like having to muck around in my settings GUI or make a macro for something so simple when I could just implement the feature in a way that doesn't disrupt any of the module's previous behaviors. Not strictly necessary by any means, just very nice-to-have, and I made it an option in case some people do want each and every click with Freecam to start pathing them.



> Its has in Meteor+ https://github.com/MeteorClientPlus/MeteorPlus And even better because it works from any distance to the block.

Yes, it's also in Stardust and already includes all of the recommendations I stated above.

ThrowawayPR:
> The double-click setting is something I added in [my own implementation](https://github.com/0xTas/stardust/blob/1.21.1/src/main/java/dev/stardust/mixin/meteor/FreecamMixin.java) because I personally don't like having to muck around in my settings GUI or make a macro for something so simple when I could just implement the feature in a way that doesn't disrupt any of the module's previous behaviors. Not strictly necessary by any means, just very nice-to-have, and I made it an option in case some people do want each and every click with Freecam to start pathing them.

That's a fair point, I'll work on it soon.

ThrowawayPR:
Entity detection, a max distance of 256 blocks and a double click option have been implemented.

PR #5036, opened 2024-12-07
Command api v2

RacoonDog:
## Overview

This pull request contains new API functions for addon developers, bug fixes, performance fixes, QOL fixes, 4 new commands, and rewrites of existing commands to add new functionality.

### .nbt
- Removed all subcommands dealing with itemstacks, they have been moved to the new `.components` command.
- Added selectors for getting NBT data from other entities and block entities.

### .components
- Rewrote all subcommands that used now-deprecated NBT data into using item data components directly.
- Added the `full` flag to switch between get/copying the full `ComponentMap` or just the `ComponentChanges` of an itemstack.
- Readded the `paste` subcommand.
- Added the `rename` subcommand.

### .macro
You can now create new macros through commands with `.macro create`, delete them with `.macro delete`, rebind them with `.macro bind`, and modify their contents with `.macro edit`

### .enderchest
`.enderchest` will now display an error if the ender chest contents are unknown instead of opening an empty inventory.

### .components
Contains the subcommands that deal with itemstacks that were removed from `.nbt`

### .config
Edit your Meteor Client configurations through commands much like how `.setting` works with modules configurations.

### .transmogrify
Lets you change the appearance of an item in your inventory to any other item in the game through item data components. Requires creative mode.

## API Changes

- The command registry access (`Command.REGISTRY_ACCESS`) and command dispatcher (`Commands.DISPATCHER`) are now dynamically rebuilt on every world join to properly support command argument types which make use of dynamic registries.

- `NotebotSongArgumentType` will now throw a command syntax exception when the input does not match an existing file in the notebot folder, meaning manual checking is no longer necessary.

- `FakePlayerArgumentType`, `FriendArgumentType`, `ProfileArgumentType`, and `WaypointArgumentType` are now properly typed instead of using `String` and external parsing.

- `WaypointArgumentType` no longer has an option for using a greedy string, it now uses `StringReader#readString` which either takes in a single word or a quoted string.

- Two new argument types imported from [clientarguments](https://github.com/xpple/clientarguments) under the MIT license: `BlockPosArgumentType` and `EntityArgumentType`.

- Utility classes for dealing with creative inventory packets in commands, `ItemSlotArgumentType` and `CreativeCommandHelper`.

- Utility classes for encoding & decoding itemstack `ComponentMap`s and `ComponentChanges` to and from string, `ComponentMapWriter` and `ComponentMapReader`.

- Utility class for decoding a `Path` in a command, `PathReader`.

## Performance Optimizations

[...]

PR #5178, opened 2025-02-06
Rewrite `.setting` for finer controls and input suggestions

### LINKED ISSUE #418: [SUGGESTION] Command improvements

RacoonDog:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Requires #5036 

## Related issues

Closes #418 

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/Command.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/Command.java
@@ -24,8 +24,8 @@ import java.util.List;
 
 public abstract class Command {
     protected static final CommandRegistryAccess REGISTRY_ACCESS = CommandManager.createRegistryAccess(BuiltinRegistries.createWrapperLookup());
-    protected static final int SINGLE_SUCCESS = com.mojang.brigadier.Command.SINGLE_SUCCESS;
     protected static final MinecraftClient mc = MeteorClient.mc;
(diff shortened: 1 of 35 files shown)


RedCarlos26:
Based

PR #5221, opened 2025-02-25
Update Pitch40 to more efficient angles

Pilot1782:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

When Pitch40 mode is selected in ElytraFly, the player will now pitch down to 32-33 deg then after hitting the lower bound, will pitch at the supplied rate to -49 deg. Finally they will pitch down at roughly 0.5 deg/tick until looking at 32-33 deg. These angles come from the findings of [this video](https://www.youtube.com/watch?v=F69K1-cMaf8&t=187s) where said angles are much more optimal than the standard 40/40 profile. Ideally this will net ~3 block of max altitude gain per cycle until ~90 block above the lower bound.

# How Has This Been Tested?

In testing in a default superflat single-player world, where the player starts on a tower to y=13, pitch rate of 13, lower bound of -40 and upper bound of 0, the maximum height I achieved was roughly y=51.

# Checklist:


Wide-Cat:
Why did you bother making the pitch changes random in a certain range? I don't really see what benefit it has over just using the respective values where appropriate.

Pilot1782:
In terms of the effectiveness of the module, it has no change, but to a server running some level of anticheat looking for repeated movements this level of "error" is within the range of a human.

PR #5225, opened 2025-02-27
Allow for addons to add their systems

maxsupermanhd:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Imagine an addon wanting to store some global info (like hashed seeds of dimensions that got discovered in runtime), adding system to store that kind of info that would be "managed" by meteor would be amazing.

## Related issues

Did not search

# How Has This Been Tested?

It builds

# Checklist:

- [x] I have added comments to my code in more complex areas. (none)


RacoonDog (approved):
although it would be better to have more a sensible way to store persistent data across modules/addons, i can't be asked to write it myself 👍

PR #5249, opened 2025-03-12
Allow hoes and Fortune tools for nether warts (#5243)

### LINKED ISSUE #5243: Nether Wart with Auto Tool
### Describe the feature

Hi there.

Is it possible to collect nether warts with a hoe instead of axe? Or another way, just use any tool with Fortune enchantment. Because i can collect 50% more warts with it.

Was trying to search the same question in the Discord and here but without any luck.
Assume i should be a new with such question.

Thanks.

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

nxtcarson:
Fixes #5243
     
     - Added support for hoes to be used on nether warts
     - Added nether warts to the list of fortunable blocks
     - Added bonus score for hoes when mining nether warts


Wide-Cat:
Don't bother making it switch to hoes for nether wart, it gets instamined with your fist so there's literally no point. Just add nether wart to the list of blocks you can use fortune on and call it a day.

Wide-Cat:
see above comment

PR #5252, opened 2025-03-13
Fix critical hit timing in jump mode (#5231)

### LINKED ISSUE #5231: [Bug] Auto-critical (jump mode) hits too early.
### Describe the bug

When you use auto-crit, it does not actually crit because it hits before you start falling.

### Steps to reproduce

1. Enable "critical."
2. Select "jump" in the options.

### Meteor Version

22

### Minecraft Version

1.21.4

### Operating System

Windows

nxtcarson:
Fixes #5231
   
   Changes:
   - Added peak detection for jump mode critical hits
   - Now attacks at the highest point of jump for maximum damage
   - Fixes issue where attacks would happen too early in jump mode


Gingerbeard5773:
I was unable to reproduce the bug after trying on a few servers

Issue #5161, opened 2025-01-30
momentum with air jump


NielsP13:
### Describe the bug

when you jump normally with air jump on you get launched forwards and air jump does not carry elytra momentum anymore.
air jump used to carry the momentum of an elytra when still sprinting (by sprinting and jumping or flying while sprinting). this still worked in the 1.21 version.

### Steps to reproduce

1. i enabled air jump
2. i got launched forwards when i jump on the ground

1. i enabled air jump
2. i sprint, jump and enable my elytra
3. i jump in the air and all my momentum is gone

### Meteor Version

Meteor-client-1.21.4-9.jar

### Minecraft Version

1.21.4

### Operating System

Windows

### Before submitting a bug report

- [x] This bug wasn't already reported (I have searched bug reports on GitHub).

- [x] This is a valid bug (I am able to reproduce this on the latest dev build).

NielsP13:
when i get launched forwards im also moving btw

Wide-Cat:
I'm unable to reproduce any of these issues. Jumping on the ground with air jump enabled just sends me upwards normally, I dont move forwards, and jumping with an elytra carries my momentum on and propels me forwards. Can you send a list of your active modules?

Wide-Cat:
Reopen this if you find a consistent way to reproduce it

PR #5253, opened 2025-03-13
Fix horizontal momentum preservation in AirJump (#5161)

nxtcarson:
fixes an issue with AirJump where it boosts the player's horizontal momentum, preventing that. Changes include:
- preserving horizontal velocity (x and z) during air jumps
- attempts to fix the issue where elytra momentum was being lost during air jumps


Wide-Cat:
see https://github.com/MeteorDevelopment/meteor-client/issues/5161#issuecomment-2721755575

Wide-Cat:
Will accept this pr if the bug is confirmed to be reproducible

PR #5254, opened 2025-03-13
Highway builder: baritone pathing

hexadec1mal:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

Now, pathing, breaking, and placing don't depend on the player's position. Collecting shulkers and obsidian is now also done using baritone, so blockade states have been removed. I was rewriting some parts of the code, so I might have deleted something important. If you have any questions about my code, please ask.

## Related issues

None.

# How Has This Been Tested?

Will test properly later, after review. And I will probably change something else, so making this draft for now.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/Modules.java
@@ -555,7 +555,6 @@ public class Modules extends System<Modules> {
         add(new EChestFarmer());
         add(new EndermanLook());
         add(new Flamethrower());
-        add(new HighwayBuilder());
         add(new LiquidFiller());
         add(new MountBypass());
         add(new NoGhostBlocks());
@@ -568,6 +567,7 @@ public class Modules extends System<Modules> {
 
         if (BaritoneUtils.IS_AVAILABLE) {
             add(new Excavator());
+            add(new HighwayBuilder());
             add(new InfinityMiner());
         }
     }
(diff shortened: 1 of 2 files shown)


Wide-Cat:
I have some local changes regarding highway builder that I want to clean up and commit before examining something like this. I haven't tested this yet but I skimmed the code, and I don't really like the fact that the module now requires baritone to be added in order to be used. If this were to be added I think it should be optional for it to use baritone.

How does the module perform using baritone compared to without? Is it faster/slower, or more reliable? There are a couple of small issues with the module as it stands, but from my testing it's very stable overall, so I can't say with certainty if using baritone comes with more benefits that not yet. I'll test it on my own after I finish my local changes but I want to hear your perspective.

hexadec1mal:
It's faster now because it can break/place and move simultaneously.

These changes
image
prevent moving away from the starting direction, so it should be more stable now.

In hwu discord, I saw people saying that they lose obsidian when mining ender chests. I have never seen this myself, but since baritone now collects obsidian, it shouldn't happen anymore.

Making baritone optional requires 2 versions of the module, which can't be consistently kept in sync, so I don't think that's an option.

RedCarlos26:
@hexadec1mal check discord

hexadec1mal:
Should not have taken this long.
Task scheduler is now much simpler and should be more stable.
Some old code is still here. I will remove it later.
Double mining is not yet reimplemented.
@Wide-Cat

Gingerbeard5773:
why convert to baritone if the movement is already good? is baritone more reliable? to me it seems unnecessary especially since it would make the module require baritone api which many meteor users don't have.

hexadec1mal:
Baritone is more reliable for both walking and collecting obisdian/shulkers. Reduces code size as well

PR #5255, opened 2025-03-14
feat: improved .modules command for listing and toggling modules

RisDN:
This commit adds a small change to the .modules command.

## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Originally the ```.modules``` command had no arguments and it just displayed all modules. Now that behaviour moved to ```.modules list```. The addition is that now you can toggle, disable and enable modules by this command. 

Examples: 
```.modules list```
```.modules enable nuker```
```.modules disable auto-jump```
```.modules toggle kill-aura``` 

# How Has This Been Tested?

It has been tested on my local machine in a singeplayer world by trying out every usecases possible.

[Showcase video](https://www.youtube.com/watch?v=TZAF_Ndj5SI)

# Checklist:


DrParanoya:
you can use .toggle for that

Wide-Cat:
There isn't really any reason for this to be added considering `.toggle` already does what this proposes. If you feel like that command could be extended in functionality then by all means open a pr for it, but there's not really much point in duplicating it to `.modules` as well.

PR #5285, opened 2025-03-30
Caught nullpointererror caused by getting world name

### LINKED ISSUE #5259: minecraft crashes when saving and exiting survival world
### Steps to reproduce

1. join the world
2. press save and exit button

### Link to crash report/log (upload to https://mclo.gs and paste resulting link here)

https://mclo.gs/C4Y4vva

### Meteor Version

1.21.4-22

### Minecraft Version

1.21.4

### Operating System

Linux

### Before submitting a crash report

- [x] This crash wasn't already reported (I have searched crash reports on GitHub).

- [x] This is a valid crash (I am able to reproduce this on the latest dev build).

kirxw:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description
When saving a singleplayer world after leaving the Utils.getWorldName() function thinks you're still in the world, causing a NullPointerError. I decided to just catch this error and return "FAILED_BECAUSE_LEFT_WORLD" instead of a world name. This is given instead of an empty string so StashFinder doesn't make unnamed directories if it triggers this bug.

## Related issues
Fixes #5259

# How Has This Been Tested?

Here's a youtube video demonstrating the issue and the fix. https://youtu.be/LRlU5--Dwks

# Checklist:

- [ ] My code follows the style guidelines of this project.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/utils/Utils.java
+++ b/src/main/java/meteordevelopment/meteorclient/utils/Utils.java
@@ -378,11 +378,21 @@ public class Utils {
         if (mc.isInSingleplayer()) {
             if (mc.world == null) return "";
 
-            File folder = ((MinecraftServerAccessor) mc.getServer()).getSession().getWorldDirectory(mc.world.getRegistryKey()).toFile();
-            if (folder.toPath().relativize(mc.runDirectory.toPath()).getNameCount() != 2) {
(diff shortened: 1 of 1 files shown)


Wide-Cat:
Wouldn't it be more straightforward to add another null check for `mc.getServer()` under the one for `mc.world`, instead of using a try block?

PR #5289, opened 2025-03-31
Adding parity to glow item frame nametags

ponjomethodos:
Hello friends,

This commit will add parity to both item frames and glow item frames in nametags, ensuring they behave the same.
This commit is very simple.
Kindly review and merge. 😊

-Ponjo


ponjomethodos:
👍

PR #5294, opened 2025-04-01
merge

tw0ten:


Big-Iron-Cheems:
Yeah I'll let you know that manually calling `System.gc()` is both stupid and useless.
Is this blatantly skidded from [TrouserStreak](https://github.com/etianl/Trouser-Streak/blob/main/src/main/java/pwn/noobs/trouserstreak/commands/GarbageCleanerCommand.java)? 👀

machiecodes:
I for one think this should be reopened and merged

Issue #3694, opened 2023-05-31
[Suggestion] Nuker whitelist hotkey


RaceIsIm:
### Describe the feature

Self explanatory. Nuker whitelist / unwhitelist onKey. Huge QoL since it takes long to remove / add stuff to nuker
Edit: I meant you aim at a block and hit a key to whitelist/remove a block and some people didnt get that srry.

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

GoofyBoi420:
or make a profile and make macro to change profile

RaceIsIm:
This doesnt help really, there are too many blocks, you would wanna be able to hit a key to whitelist / remove blocks to nuker

GoofyBoi420:
> This doesnt help really, there are too many blocks, you would wanna be able to hit a key to whitelist / remove blocks to nuker

i dont get whats wrong with the profiles

GoofyBoi420:
So do you have a keyboard with more than 830 keys?

RaceIsIm:
IM sorry what? either im bad at explaining or u are bad at understanding, what i meant was you aim at a block, and you hit a key to whitelist / unwhitelist it.

GoofyBoi420:
you are bad at explaining 😄

RaceIsIm:
😂 Yea no my bad, but you got it now? aim at block, hit key, whitelisted for nuker, hit again, removed

GoofyBoi420:
yes 👍

PR #5298, opened 2025-04-02
Add auto-select mode to nuker

LittleApple-fp16:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Added a new whitelist mode 'AutoSelect'. It can just select the block you are looking at when active.
This is convenient when using keybind to mine specific blocks and also can be switched through activation.

## Related issues

# How Has This Been Tested?

It works on my local test.

# Checklist:


Gingerbeard5773:
It works but is kind of niche in its use. There is also no information in the module in game to let players know how to use it. Some players might get blindsided on how it works and end up confused

LittleApple-fp16:
> It works but is kind of niche in its use

This is actually from Wurst😂

Gingerbeard5773:
Perhaps this PR should do #3694 instead. I think it would be more useful and it would allow for ppl to choose multiple blocks at the same time

PR #5299, opened 2025-04-02
Support for line breaks in module descriptions

CrowTheBest:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Added support for line breaks in module description and fixed an infinity loop caused a game crash if a word is longer than module window size! Hints and usage info look way better now 

# How Has This Been Tested?
**Before:**
image

**After:**
image

These screenshots were taken at different window resolutions, because original code gets stuck in an infinite loop on a small window size
I actually use this fix myself

# Checklist:

- [x] I have added one unnecessary comment to my code!

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/widgets/WMultiLabel.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/widgets/WMultiLabel.java
@@ -36,7 +36,6 @@ public abstract class WMultiLabel extends WLabel {
         int iInLine = 0;
 
         for (String line : textLines) {
-            System.out.println(Arrays.stream(textLines).toList());
             for (String word : line.split(" ")) {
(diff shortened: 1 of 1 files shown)


machiecodes:
Good idea but the line overflow definitely needs to be fixed, I also noticed a println statement still in there assumedly from debugging

CrowTheBest:
I was thinking of very long words. But I believe it hasn't happened to anyone yet, because it causes the game to crash. I think there would be an open issue about this. In most cases, long words like in my example would never be used. Is it really necessary to handle and wrap such words? If it's really needed, I'll implement it

Sorry for the println statement, it will be removed!

CrowTheBest:
I also suggest to always use `WMultiLabel` instead of `WLabel`. Maybe make the background less transparent or add a border line

image
image

PR #5356, opened 2025-04-25
Add 3d rendering with depth-test to api, among other things

RacoonDog:
Changes:
- `Render3DEvent` now has a field called `depthRenderer` which holds a separate `Renderer3D` with depth test enabled (does not render through blocks).
- `LightOverlay` now uses the depth-test renderer.

Improvements:
- `MeshBuilder` no longer accidentally writes data out of bounds into unmanaged memory, causing offheap corruption.
- `MeshBuilder` now has a separate constructor that takes in initial buffer sizes.
- `MeshBuilder` buffers are now lazily allocated.

Extra considerations:
If you were rendering things by manually calling `MeshBuilder` methods, you now have to manually do bounds checking through calling `MeshBuilder#ensureCapacity`. (This is not required when using existing renderer classes like `Renderer2D` and `Renderer3D`)

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/events/render/Render3DEvent.java
+++ b/src/main/java/meteordevelopment/meteorclient/events/render/Render3DEvent.java
@@ -14,13 +14,15 @@ public class Render3DEvent {
 
     public MatrixStack matrices;
     public Renderer3D renderer;
+    public Renderer3D depthRenderer;
     public double frameTime;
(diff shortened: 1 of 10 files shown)


MineGame159:
I don't know how to feel about MeshBuilder not automatically growing the buffers. Maybe we could add bound checks into every method that writes into the buffers and only enable them in a development environment?

PR #5368, opened 2025-05-01
Fix game crash caused by NullPointerException in KillAura

RagingEnby:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Added a check to ensure `mc.player` has value before checking the `.isAlive()` function. This fixes a NullPointerException caused by leaving a server right as KillAura attacks an enemy.

## Related issues

# How Has This Been Tested?

I used it and it worked

# Checklist:


tyrannus00:
You obviously didn't test this, and it doesn't seem like you know how basic java syntax works. Unlike in js, `!mc.player` is not a valid Boolean expression, so that's not sufficient for a null check. Instead you are supposed to do this: `mc.player == null`

machiecodes:
W code

I-No-oNe:
> W code

Best code that I saw in a while

PR #5374, opened 2025-05-03
Update game version

ggc00lkid112:
i thought that maybe we can do create mod with meteor client


Big-Iron-Cheems:
My man are you good?

ggc00lkid112:
yes i am

Issue #4836, opened 2024-08-17
AutoEat forces slot until Health Threshold is satisfied


un-ited:
### Describe the bug

I am using the Auto Eat module with the following settings:
```
- General settings do not matter -
- Threshold -
Threshold Mode: Any
Health Threshold: 10
Hunger Threshold 16
```

Once I take damage, the Auto Eat module forces my hotbar to stay on one specific slot, the one it found food in, until my Health is above the defined Threshold again.
The issue with that is that if I do not have enough food to recover to the defined Health Threshold, it still won't let go of the slot. This also happens when all the food in your inventory has been used up, it will not unclip from the slot it found food in before the health is not (in my example) at least at 10 again.

I would expect it to temporarily release the slot once hunger is saturated (unless using enchanted/golden apples, which ignore saturation) and start eating again when possible or abort the eating once all food has been consumed.

### Steps to reproduce

1. Enable AutoEat
2. Set Threshold Mode to Health
3. Set Health Threshold to 18 (can be anything)
4. Put only 1 piece of food in your hotbar (must be an insufficient amount to reach the threshold)
5. Take some damage

### Meteor Version

meteor-client-0.5.8-2151

### Minecraft Version

Minecraft 1.21.1 with Fabric Loader 0.15.11

### Operating System

Windows

### Before submitting a bug report

- [x] This bug wasn't already reported (I have searched bug reports on GitHub).

- [X] This is a valid bug (I am able to reproduce this on the latest dev build).

Markrealhd:
The problem is still present, but I don’t think it’s a bug. If he waited until he could eat again, it would be just like the 'Both' mode. The only thing that could probably be changed is that he shouldn’t force the slot until he can actually eat again.

Wide-Cat:
13e10f851219cda295059d21f18262438bccd782

PR #5380, opened 2025-05-05
fix: AutoEat not switching slots when food runs out 

uxmlen:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Fixed condition that checks for food in current slot to allow switching when food is depleted.

When using AutoEat, if the food in the current slot runs out, the module fails to switch to another slot with food. The player gets stuck with the empty slot until another food item appears in that same slot.

- Fixed the condition in `onTick` that checks if the current slot contains food
- refactor method  onTick to remove nesting

bug demo:

## Related issues

#4836

# How Has This Been Tested?

you can see in the gif that the slots can be switched after applying this patch

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoEat.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoEat.java
@@ -124,18 +124,18 @@ public class AutoEat extends Module {
             // If we are eating check if we should still be eating
             if (shouldEat()) {
                 // Check if the item in current slot is not food
-                if (mc.player.getInventory().getStack(slot).get(DataComponentTypes.FOOD) != null) {
+                if (mc.player.getInventory().getStack(slot).get(DataComponentTypes.FOOD) == null) {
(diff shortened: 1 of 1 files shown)


Gingerbeard5773:
I was able to reproduce the bug, and then I tried the PR fix and it works! But I noticed another bug while I was testing. It exists prior to the PR, maybe you should fix this one since its kind of similar.

https://github.com/user-attachments/assets/41a7d4a9-0591-4181-b583-d7984f246583

AutoEat forces the player to switch to the food slot despite being unable to eat the food. The food cannot be eaten when the player is at full hunger. This wouldn't happen with foods like Gapples or Egaps because they can be eaten regardless of hunger which is probably why this bug hasn't been fixed.

@uxmlen

uxmlen:
@Gingerbeard5773 done

AutoEat no longer switches to food that can't be eaten at full hunger unless it's always edible (like egaps, chorus etc).

Gingerbeard5773 (approved):
Tested, all bugs were resolved

PR #5385, opened 2025-05-07
added new get method in Settings api

cattyngmd:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

When i was porting some meteor addon to 1.21.5 i saw the ugly casting like this `Setting<Boolean> setting = ((Setting<Boolean>) module.settings.get(...));` so i decided to add new method to the api that prevents that.

## Related issues

# How Has This Been Tested?

it works

# Checklist:


RacoonDog:
better to just make the setting fields public and access the setting directly from there rather than going through the settings object

cattyngmd:
yeah but still not all setting fields are public in meteor modules itself for example

PR #5409, opened 2025-05-17
Avoid mojang API ratelimit exception spamming on startup when refresh…

orsondmc:
…ing friends

## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Meteor does not handle rate limits when refreshing Friends list. This results in the friends list never getting updated and a lot of spam in logs at startup.

1 second delay should do the trick.

## Related issues

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/commands/ReloadCommand.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/commands/ReloadCommand.java
@@ -28,7 +28,7 @@ public class ReloadCommand extends Command {
             Systems.load();
             Capes.init();
             Fonts.refresh();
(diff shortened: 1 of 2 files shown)


RacoonDog (changes requested):
`Friends.get().iterator()` returns an `ArrayList` iterator, which would throw a `ConcurrentModificationException` if any friends are added/removed during the refresh

also opening the friends tab during the initial refresh would cause all of the remaining friends to be refreshed twice (and faster), so there's still rate-limit issues there

edit: considering the player head texture api endpoint is used in a few different areas in meteor, i think it'd just be better to make a dedicated rate limited queue for them

Wide-Cat:
fixed by ff7373a8afb818cdb9670bff802352a09eef7365

Issue #5439, opened 2025-05-30
Throttle Auto Walk on chunk loading issues


magicus:
### Describe the feature

I am using Auto Walk to travel with boat along the blue ice highways on 2b2t. This reaches speeds up to 72 m/s, which puts a strain on the server's ability to load the chunks ahead. If you continue with this high speed when chunks are not loaded properly, you will end up outside loaded chunk and need to relog.

I have been manually throttling the Auto Walk feature, turning it off when the amount of loaded chunks ahead of me reaches a critical minimum, and turning it back on once they have loaded. This works but makes the process require manual supervision, which is not as convenient.

I think it could be possible to implement this kind of functionality directly in Meteor. My thinking is along this lines:

1) Given the player's current speed and direction, calculate which chunks will be entered within a certain time. (For a rough number, let's say 500 ms).
2) Check if all these chunks are loaded.
3) If not, put Auto Walk in pause mode, and start watch for chunk loading.
4) Once all these chunks have been loaded, move Auto Walk out of pause mode.

Most of this sounds like it should be fairly simple to implement, though I am not really sure how easy it is to detect chunk loading. Depending on what kind of chunk loading instrumentation already present in Meteor, this might prove more of a challenge.

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

DrParanoya:
ELytraFly has a dontGoIntoUnloadedChunks setting, so I'm pretty sure it shouldn't be too difficult to add it to AutoWalk

magicus:
> ELytraFly has a dontGoIntoUnloadedChunks setting, so I'm pretty sure it shouldn't be too difficult to add it to AutoWalk

I had a look at the code. 

ElytraFly just checks the very next chunk you are about to enter, and sets your velocity to 0 if it is not loaded. If you're traveling on an ice road with boat, it takes like forever to slow down, so that will (likely) not be near enough.

There is an event in Meteor that is sent when chunks are loaded; I guess the AutoWalk module could subscribe to that, and that way try to keep track of the loaded chunks. But it seems far from trivial; you need to know when to yeet them from the set of remembered chunks, you should only store information about loaded chunks that are ahead of you in your current direction of travel, etc.

31a05b9c:
[...]

Issue #5438, opened 2025-05-30
Option to disable Auto Walk if user press any walk key


magicus:
### Describe the feature

I have been using the [Auto-Run](https://github.com/Emonadeo/autorun) mod prior to discovering Meteor. It has the feature that if you start messing with the ASWD controls, it disabled itself. That makes for a much speedier resolution if you find yourself in a sudden mess. 

I realize there might be use cases for the current Meteor implementation to just continue walking no matter what, but it would be great if it were at least an option to behave as Auto-Run.

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

PR #5440, opened 2025-05-30
Add some QoL options to auto-walk

chri-k:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds options to disable auto-walk when a movement key is pressed and on vertical movement, adds option to disallow movement into unloaded chunks (set by default)

## Related issues

#5439
#5438

# How Has This Been Tested?

works*. interaction of unloaded chunk prevention with anticheats not tested, fix already exists in case it's broken.

\* `GuiMove` causes the emission of a `KeyEvent` with `KeyAction.Press` for held movement keys when the player inventory is opened (no other GUI produces the same result), which causes disable-on-input to trigger on that. (fixed by #5480)
 
# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/AutoWalk.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/AutoWalk.java
@@ -6,8 +6,9 @@
 package meteordevelopment.meteorclient.systems.modules.movement;
 
 import meteordevelopment.meteorclient.events.entity.player.PlayerMoveEvent;
-import meteordevelopment.meteorclient.events.world.TickEvent;
 import meteordevelopment.meteorclient.events.meteor.KeyEvent;
+import meteordevelopment.meteorclient.events.meteor.MouseButtonEvent;
+import meteordevelopment.meteorclient.events.world.TickEvent;
 import meteordevelopment.meteorclient.mixininterface.IVec3d;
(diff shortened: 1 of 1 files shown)


Wide-Cat:
Add a listener for MouseButtonEvent as well, since it's possible people have movement keys bound to mouse buttons.

Wide-Cat:
Can you elaborate on the issue to do with GuiMove emitting key events while you move with the player inventory open, but not with any other screens? Looking at the GuiMove code I don't really see any reason why if you are able to move with a screen open it would not emit key events.

31a05b9c:
you misread that. the issue is not that it doesn't emit key events, but that it emits fake ones.

`GuiMove` emits `KeyEvent` whenever it changes the state of a key. apparently this was added 3 years ago by minegame to fix some issue with `FreeCam`.

`AutoWalk` happens to trigger this when the inventory is opened, causing an unexpected `KeyEvent` to be emitted for the movement key of the direction it is walking in.

at the moment there is no good way to tell whether the key was actually pressed, since many modules alter the key states (and in this case the event is emitted immediately after `GuiMove` itself overrides the key state)

31a05b9c:
making `KeyEvent` keep track of whether it is synthesized or not is the easiest solution, and should be good enough (there are hypothetical situations where it isn't, but realistically it is)

PR #5464, opened 2025-06-05
skip regular lightmap computations with Fullbright & Xray

RacoonDog:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Optimization

## Description

instead of overwriting the work done by `LightmapTextureManager#update`, simply skip the entire function call if applicable

also, replace the `createRenderPass` call with the more lightweight `clearColorTexture` call

## Related issues

none

# How Has This Been Tested?

works on my pc :+1:

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/LightmapTextureManagerMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/LightmapTextureManagerMixin.java
@@ -5,6 +5,7 @@
 
 package meteordevelopment.meteorclient.mixin;
 
+import com.llamalad7.mixinextras.sugar.Local;
 import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.blaze3d.textures.GpuTexture;
 import meteordevelopment.meteorclient.systems.modules.Modules;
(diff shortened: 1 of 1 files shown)


crosby-moe:
Before: https://spark.lucko.me/J8QffK89s1?hl=4094 (~0.20% of render thread)
After: https://spark.lucko.me/uuws554Bjx?hl=1385 (~0.07% of render thread)

MineGame159 (changes requested):
The logic itself looks fine but you are injecting **before** the profiler push() call, calling pop() and then cancelling the injection

crosby-moe:
> The logic itself looks fine but you are injecting **before** the profiler push() call, calling pop() and then cancelling the injection

yeah that sounds right to me, i made the normal `lightTex` profiler location encompass the `clearColorTexture()` call

(i inject before the `push()` call, and i call `pop()` myself because i cancel the rest of the method)

MineGame159:
> > The logic itself looks fine but you are injecting **before** the profiler push() call, calling pop() and then cancelling the injection
> 
> yeah that sounds right to me, i made the normal `lightTex` profiler location encompass the `clearColorTexture` call

What? The injection is called before the push() call, so your pop() call not only happens before push() but it never even gets to push() because the method is cancelled.

crosby-moe:
update: i was the stupid one

PR #5472, opened 2025-06-06
add option to keep sneaking while using freecam

Powie69:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

I added an option to keep sneaking while using freecam cause it was annoying me

## Related issues

# How Has This Been Tested?

it works on my machine 👍

https://github.com/user-attachments/assets/a25aa626-73d3-483a-9f7d-4935c5c1c5c9

# Checklist:


Powie69:
https://github.com/user-attachments/assets/f8835894-5b81-437b-83d6-2602dfd7d0c9

Ynthra:
You don't need an option for it, you can just bind the sneak module to the same key as freecam

Powie69:
The difference between binding sneak the module and this pr is that:
- It will only sneak if I'm pressing the sneak key (right before i toggle freecam)
- If I'm not pressing the sneak key while I toggle freecam, then I will not sneak

PR #5474, opened 2025-06-06
inventory-tweaks anti-drop minor improvements

chri-k:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

- adds option (on by default) that prevents player from accidentally placing items on the anti-drop list into item frames or decorated pots (item frames will no longer randomly steal tools)
- adds an option to temporarily override anti-drop by holding a keybind

## Related issues

none

# How Has This Been Tested?

works locally

# Checklist:


MineGame159:
Sorry for the delay, when you update the PR I am down to merge it. Also, in my opinion you should put the settings into their own group even if it breaks existing configs. Better to do the change now than later when there would be more settings to change.

31a05b9c:
> Also, in my opinion you should put the settings into their own group even if it breaks existing configs. Better to do the change now than later when there would be more settings to change.

saved as precedent

PR #5548, opened 2025-07-03
Add reflection based init for Commands and Modules

Big-Iron-Cheems:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

This implementation provides ways for addons to register their own packages, and add commands/modules conditionally.
Also removes deprecated and long unused `List<Module> modules` field.

## Related issues

None.

# How Has This Been Tested?

Tested in singleplayer world.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/Commands.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/Commands.java
@@ -27,48 +27,16 @@ import static meteordevelopment.meteorclient.MeteorClient.mc;
 public class Commands {
     public static final CommandDispatcher<CommandSource> DISPATCHER = new CommandDispatcher<>();
     public static final List<Command> COMMANDS = new ArrayList<>();
+    private static final Set<String> SCAN_PACKAGES = new HashSet<>();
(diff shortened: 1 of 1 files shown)


Big-Iron-Cheems:
Build died, perhaps a desync in my files, will checkout.

MukjepScarlet:
This is not good because it uses reflection on the user start the client. It would be better to do this in compile-time like Lombok.

Big-Iron-Cheems:
This PR is a very simplified implementation of this branch by @RacoonDog: https://github.com/RacoonDog/meteor-client/tree/module-initialization

The goal is to eventually support proper reflection based init in such a way that we can define flexible conditions and support add-ons as well.

What Lombok does is not enough for what we're aiming to do.

MukjepScarlet:
Yes, If you wanna support external modules, reflection is needed. If only built-in ones, codegen at compile time will be better.

Big-Iron-Cheems:
Closing as this is really bad / unnecessary at the moment (also extremely outdated).
There could be a better impl in the future if we find it useful.

PR #5550, opened 2025-07-05
Ignore AutoTool if not in survival

Powie69:
## Type of change

- [X] Bug fix
- [X] New feature

## Description

yuh uh it wont switch to other tools no more if you aint in survival
Ik like you can still break stuff on adventure mode if the tools has attribute stuff but no one plays adventure anyway and it will be just be bloat

## Related issues

image

# How Has This Been Tested?

👍

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
@@ -27,6 +27,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.item.ShearsItem;
 import net.minecraft.registry.tag.BlockTags;
 import net.minecraft.registry.tag.ItemTags;
+import net.minecraft.world.GameMode;
 
 import java.util.List;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


DrParanoya:
> but no one plays adventure anyway
Pretty sure there are a handful of people out there that use AutoTool in adventure mode.

Powie69:
>  but no one plays adventure anyway
> > Pretty sure there are a handful of people out there that use AutoTool in adventure mode.

yeah true I'll try update the pr to take that into account
But if I don't succeed I'll just double down on the cope and say that no one plays adventure anyway and that would just result in bloat.

src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
```diff
@@ -146,6 +147,7 @@ private void onTick(TickEvent.Post event) {
     @EventHandler(priority = EventPriority.HIGH)
     private void onStartBreakingBlock(StartBreakingBlockEvent event) {
         if (Modules.get().isActive(InfinityMiner.class)) return;
+        if (mc.player.getGameMode() != GameMode.SURVIVAL) return;
```
RacoonDog, line 150: adventure mode

RacoonDog (changes requested):

PR #5576, opened 2025-07-16
fixed sprint jumping not working with HighJump

cattyngmd:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Fixed a HighJump module not allowing player to sprint jumpi with jump-multiplier set to 0

# Checklist:


Wide-Cat:
Maybe I'm misunderstanding something but if you set the HighJump multiplier to zero, essentially making it so you can't jump, wouldn't it also be logical that none of the side effects of jumping happen as well - i.e. the sprint jump speed boost? Would like to hear your thoughts on this.

cattyngmd:
the main reason i made that is to make new 2b2t elytrafly work on meteor without addons.

and yeah you're right about it being unreasonable i just made it in rush.

Issue #5553, opened 2025-07-05
Highway builder Quality of Life changes


dplexlegend:
### Describe the feature

Make it so if you have multiple block placements per tick on, it automatically turns on no ghost blocks when you enable highway builder (since on many servers it won't work otherwise) and off when you turn highway builder off. Or,if you don't have time,  on description of the setting it could maybe say: <what it normally says> (Note: turn on no ghost blocks or else it might not work) because it took me a LONG time to find out what the problem was.

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

Wide-Cat:
We could add some information to highway builder somewhere explaining the benefits of using no ghost blocks, but there are drawbacks to using the module as well, so I don't belive it's appropriate to always use the two together.

PR #5589, opened 2025-07-21
Add auto no ghost block setting to highway builder

mynameisbob1928:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Added a new setting into the highway builder module that will automatically enable the no ghost blocks module when the highway builder is activated.

## Related issues

#5553 

# How Has This Been Tested?

I tested against different cases of if the highway builder was enabled or not and if the no ghost blocks was enabled and the various ways it could happen.

# Checklist:


Wide-Cat:
As said in #5553 there are drawbacks to using both modules at the same time. I don't really think this would be appropriate.

PR #4549, opened 2024-04-27
Improve ClickTP adding paperclip exploit

Not-a-Tyler:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Added the liveoverflow teleportation exploit to click tp and made it work while using freecam.

## Related issues

#2775 

Tested on a papermc server

https://github.com/MeteorDevelopment/meteor-client/assets/57022279/e881687e-4e40-4688-aef9-c5d6646c19b6

# Checklist:

- [x] My code follows the style guidelines of this project.
- [x] I have added comments to my code in more complex areas.
- [x] I have tested the code in both development and production environments.

Issue #2775, opened 2022-09-18
[Suggestion] Click TP should use block aimed at in freecam when active


ryantheleach:
### Describe the feature

At the moment Click TP is limited by having to look at the block in first person.

It'd be far more helpful if you could use Click TP in freecam.

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

I-Got-You-Dead:
Not how it works…

Mathrandom7910:
I don't know what's funnier, these issues, or I-Got-You-Dead's comments

I-Got-You-Dead:
> I don't know what's funnier, these issues, or I-Got-You-Dead's comments

My goofy ahhh comments

ryantheleach:
@I-Got-You-Dead @Mathrandom7910 

This seems like a reasonable request, rather then mocking, care to explain the issues? Mining blocks that is being aimed at already works in freecam.

I-Got-You-Dead:
You will basically Insta rubber band right back to where you were

I-Got-You-Dead:
So there really is no point to adding this

ryantheleach:
That's not true at all, otherwise clicktp wouldn't function.

I-Got-You-Dead:
tp through a block bud

ryantheleach:
I'm not talking about using it to teleport through blocks. Clearly that won't work, the same as if you try to use Click TP currently to teleport too far away.

Click TP currently doesn't respect where you are pointing in freecam, instead using the player entities heading. It would be helpful when building in freecam, if ClickTP wasn't restricted to wherever the player was pointing, and instead teleported to where the camera is pointing.

I-Got-You-Dead:
mid

ryantheleach:
Your so cringe dude.

seasnail8169:
> Your so cringe dude.

You're*

seasnail8169:
:trollface:

Moosbee:
There is now a pull request

PR #5595, opened 2025-07-22
Compatibility between clicktp and freecam

mynameisbob1928:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

This checks for if freecam is enabled and then uses MinecraftClient.crosshairTarget (which is already modified in a mixin) instead of raycast. This does limit it so that you can only tp to blocks that are targeted so you can't tp to a block 30 blocks away (This does not affect use while not in freecam).

While searching issues I did see two other pull requests that did implement a raycast method to get around the aforementioned limitation which may be better to use than this: [4549](https://github.com/MeteorDevelopment/meteor-client/pull/4549) and [4146](https://github.com/MeteorDevelopment/meteor-client/pull/4146)

## Related issues

#2775 

# How Has This Been Tested?

Tested in singleplayer and a public muliplayer server

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/ClickTP.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/movement/ClickTP.java
@@ -11,6 +11,8 @@ import meteordevelopment.meteorclient.settings.Setting;
 import meteordevelopment.meteorclient.settings.SettingGroup;
 import meteordevelopment.meteorclient.systems.modules.Categories;
 import meteordevelopment.meteorclient.systems.modules.Module;
+import meteordevelopment.meteorclient.systems.modules.Modules;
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/movement/ClickTP.java
```diff
@@ -41,7 +43,15 @@ private void onTick(TickEvent.Post event) {
         if (mc.player.isUsingItem()) return;
 
         if (mc.options.useKey.isPressed()) {
-            HitResult hitResult = mc.player.raycast(maxDistance.get(), 1f / 20f, false);
+            HitResult hitResult;
+
+            if (Modules.get().get(Freecam.class).isActive()) {
+                hitResult = mc.crosshairTarget;
```
RacoonDog, line 49: raycast from the freecam position

PR #5636, opened 2025-08-09
[feat] prevent misclicks on delete buttons

chri-k:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

meteor has several destructive buttons positioned a few pixels away from other buttons.
requires such buttons to be pressed twice to avoid destructive misclicks (which have happened at least a few times)

cherry-pick from profile system changes

## Related issues

none

# How Has This Been Tested?

works.

visual indications of the button requiring confirmation:
image image

# Checklist:


31a05b9c:
that is the more logical way to do it

PR #5664, opened 2025-08-27
Enhance `DropCommand` to support item quantity

### LINKED ISSUE #4222: .drop
### Describe the feature

the .drop feature should have a modification. So it can be allowed to drop a certain number of items at once instead of dropping all . For example '.drop dirt 5' 

IExploitableMan:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

The `DropCommand` now supports an optional amount argument, allowing players to drop a specific number of items instead of entire stacks. The dropping logic was refactored into a dedicated `dropItem` method for cleaner structure, and `InvUtils` gained a `dropOne` helper to handle single-item drops.

## Related issues

Closes #4222 

# How Has This Been Tested?

https://github.com/user-attachments/assets/b53904ac-6b09-45a5-8df8-c4dfbdcbff72

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/commands/DropCommand.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/commands/DropCommand.java
@@ -5,7 +5,9 @@
 
 package meteordevelopment.meteorclient.commands.commands;
 
+import com.mojang.brigadier.arguments.IntegerArgumentType;
 import com.mojang.brigadier.builder.LiteralArgumentBuilder;
+import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
 import meteordevelopment.meteorclient.commands.Command;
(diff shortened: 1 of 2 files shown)


IExploitableMan:
I'm not sure how to handle cases where `amount` is greater than items in inventory

I-No-oNe:
> I'm not sure how to handle cases where `amount` is greater than items in inventory

Just send a message using `error()` that says that you can only drop the amount u have on the inv

IExploitableMan:
> > I'm not sure how to handle cases where `amount` is greater than items in inventory
> 
> Just send a message using `error()` that says that you can only drop the amount u have on the inv

I've thought about it, but maybe it's better to warn that fewer items were dropped than were specified

I-No-oNe:
I don't think that will be a good implantation 
You want the user to have as much as possible control, so making the inv-drop happen even when the right amount hasn't been specified isn't a good idea

Wide-Cat:
I'm not sure about the command erroring if you don't have enough items. Even ignoring a bug it created, I feel like it would create more frustration if you want to drop so many items and the command fails because you are slightly underneath it - it would be smoother if it executed anyway and dropped what it could. I am happy to hear arguments against this though.

IExploitableMan:
> I'm not sure about the command erroring if you don't have enough items. Even ignoring a bug it created, I feel like it would create more frustration if you want to drop so many items and the command fails because you are slightly underneath it - it would be smoother if it executed anyway and dropped what it could. I am happy to hear arguments against this though.

I agree with giving warnings, but could you clarify which bug you're talking about?

Wide-Cat:
if you wanted to drop all of an item it would always fail, because the command passed in Integer.MAX_VALUE to the drop method and you always have less than that of an item

Issue #5207, opened 2025-02-17
Recursive macros crash the game


DrParanoya:
### Steps to reproduce

1. Create a macro that calls itself:
![Image](https://github.com/user-attachments/assets/55509270-7db9-4e9f-a61a-6e0b29175019)
2. Activate the macro.
3. Game crash!

### Link to crash report/log (upload to https://mclo.gs and paste resulting link here)

https://mclo.gs/RXmwuNn

### Meteor Version

1.21.4-18

### Minecraft Version

1.21.4

### Operating System

Windows

### Before submitting a crash report

- [x] This crash wasn't already reported (I have searched crash reports on GitHub).

- [x] This is a valid crash (I am able to reproduce this on the latest dev build).

PR #5671, opened 2025-08-30
Improve .macro functionality, allowing recursive macros etc

moxvallix:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description
Previously, if a macro called itself with .macro, the game would crash. To fix this, I have implemented a scheduling system to the .macro command, allowing it to behave similarly to /schedule.

The command syntax is now as follows:
```
.macro <name>                    - Schedules the macro with 1 tick delay
.macro <name> <time>             - Schedules the macro with <time> delay
.macro clear                     - Clear all scheduled macros
.macro clear <name>              - Clear all scheduled macros named <name>
```

As well, if a macro name has spaces in it, the autocomplete will automatically add quotes to it.

## Related issues
This PR fixes https://github.com/MeteorDevelopment/meteor-client/issues/5207

# How Has This Been Tested?
image

Rotate 90 degrees every second:
image

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/arguments/MacroArgumentType.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/arguments/MacroArgumentType.java
@@ -46,7 +46,14 @@ public class MacroArgumentType implements ArgumentType<Macro> {
 
     @Override
     public CompletableFuture<Suggestions> listSuggestions(CommandContext context, SuggestionsBuilder builder) {
-        return CommandSource.suggestMatching(Macros.get().getAll().stream().map(macro -> macro.name.get()), builder);
+        return CommandSource.suggestMatching(Macros.get().getAll().stream().map(macro -> {
+            String name = macro.name.get();
+            if (name.contains(" ")) {
+                name = "\"" + name.replace("\"", "\\\"") + "\"";
+            }
+
+            return name;
+        }), builder);
     }
 
     @Override
--- a/src/main/java/meteordevelopment/meteorclient/commands/commands/MacroCommand.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/commands/MacroCommand.java
@@ -49,9 +49,9 @@ public class MacroCommand extends Command {
                         Macro macro = MacroArgumentType.get(context);
 
                         if (!isScheduled(macro)) {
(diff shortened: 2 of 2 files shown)


31a05b9c:
the default behaviour should possibly be to run it with 0 delay, not 1, since that is what is most commonly wanted

moxvallix:
Fair enough, I have refactored it to allow for a 0 tick delay.

Wide-Cat:
apply this patch and I'll merge it

```
Index: src/main/java/meteordevelopment/meteorclient/commands/commands/MacroCommand.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/meteordevelopment/meteorclient/commands/commands/MacroCommand.java b/src/main/java/meteordevelopment/meteorclient/commands/commands/MacroCommand.java
--- a/src/main/java/meteordevelopment/meteorclient/commands/commands/MacroCommand.java	(revision 1bbcd3424fb6937bb775f44893f4abe563927e69)
+++ b/src/main/java/meteordevelopment/meteorclient/commands/commands/MacroCommand.java	(date 1756577707216)
@@ -49,9 +49,9 @@
                         Macro macro = MacroArgumentType.get(context);
 
                         if (!isScheduled(macro)) {
-                            error("This module is not currently scheduled.");
+                            error("This macro is not currently scheduled.");
                             return SINGLE_SUCCESS;
-                        };
+                        }
 
                         clear(macro);
                         info("Cleared scheduled macro.");
@@ -116,8 +116,7 @@
     public int delay;
     public Macro macro;
 
-
-    public ScheduledMacro (int tickDelay, Macro scheduledMacro) {
+    public ScheduledMacro(int tickDelay, Macro scheduledMacro) {
         delay = tickDelay;
         macro = scheduledMacro;
     }
@@ -127,14 +126,14 @@
     }
 
     public boolean run() {
-        if (delay > 0) { return false; }
+        if (delay > 0) return false;
 
         runMacro();
         return true;
     }
 
     private void runMacro() {
-        if (MeteorClient.mc.player == null) { return; }
+        if (MeteorClient.mc.player == null) return;
 
         macro.onAction();
     }

```

moxvallix:
@Wide-Cat I have applied the patch.

PR #5672, opened 2025-08-30
Allow waypoints to auto hide when a player is close enough

moxvallix:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description
This PR adds functionality to waypoints to allow them to toggle off their visibility once a player is within a particular range of the waypoint. This is useful for creating "temporary" waypoints, that only stay visible whilst the player is travelling to that point, and can automatically turn themselves off once the player has reached the destination.

As well, this PR makes death waypoints automatically hide themselves once the player is within 4 blocks of the waypoint.

# How Has This Been Tested?
image

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/WaypointsModule.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/WaypointsModule.java
@@ -105,6 +105,8 @@ public class WaypointsModule extends Module {
             Vector3d pos = new Vector3d(blockPos.getX() + 0.5, blockPos.getY(), blockPos.getZ() + 0.5);
             double dist = PlayerUtils.distanceToCamera(pos.x, pos.y, pos.z);
 
+            waypoint.hideWhenNear((int) Math.floor(dist));
+
             // Continue if this waypoint should not be rendered
             if (dist > waypoint.maxVisible.get()) continue;
             if (!NametagUtils.to2D(pos, waypoint.scale.get() - 0.2)) continue;
@@ -171,6 +173,10 @@ public class WaypointsModule extends Module {
                 .pos(BlockPos.ofFloored(deathPos).up(2))
                 .dimension(PlayerUtils.getDimension())
                 .build();
+
+            // Configure death waypoints to auto hide when the player is within 4 blocks
+            waypoint.hideWhenNear.set(true);
+            waypoint.hideWhenNearDistance.set(4);
 
             Waypoints.get().add(waypoint);
         }
--- a/src/main/java/meteordevelopment/meteorclient/systems/waypoints/Waypoint.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/waypoints/Waypoint.java
@@ -94,6 +94,21 @@ public class Waypoint implements ISerializable<Waypoint> {
         .build()
     );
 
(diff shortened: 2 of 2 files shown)


Wide-Cat:
apply this patch and ill merge it 

```
Index: src/main/java/meteordevelopment/meteorclient/systems/waypoints/Waypoint.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/meteordevelopment/meteorclient/systems/waypoints/Waypoint.java b/src/main/java/meteordevelopment/meteorclient/systems/waypoints/Waypoint.java
--- a/src/main/java/meteordevelopment/meteorclient/systems/waypoints/Waypoint.java	(revision 3f5c9461589fa2df1f73c140cac2459aed550c11)
+++ b/src/main/java/meteordevelopment/meteorclient/systems/waypoints/Waypoint.java	(date 1756569902171)
@@ -106,6 +106,7 @@
         .description("Hides the waypoint if the player is closer than this distance, and hide when near is enabled.")
         .defaultValue(8)
         .sliderRange(0, 32)
+        .visible(hideWhenNear::get)
         .build()
     );
 
@@ -118,8 +119,7 @@
     public Waypoint(NbtElement tag) {
         NbtCompound nbt = (NbtCompound) tag;
 
-        if (nbt.contains("uuid")) uuid = nbt.get("uuid", Uuids.INT_STREAM_CODEC).get();
-        else uuid = UUID.randomUUID();
+        uuid = nbt.get("uuid", Uuids.INT_STREAM_CODEC).orElse(UUID.randomUUID());
 
         fromTag(nbt);
     }
@@ -129,7 +129,7 @@
         if (texture == null) return;
 
         int preA = color.get().a;
-        color.get().a *= a;
+        color.get().a *= (int) a;
 
         Renderer2D.TEXTURE.begin();
         Renderer2D.TEXTURE.texQuad(x, y, size, size, color.get());
@@ -153,8 +153,8 @@
     }
 
     public void hideWhenNearCheck(int distance) {
-        if (!hideWhenNear.get()) { return; }
-        if (distance > hideWhenNearDistance.get()) { return; }
+        if (!hideWhenNear.get()) return;
+        if (distance > hideWhenNearDistance.get()) return;
 
         visible.set(false);
     }

```

moxvallix:
@Wide-Cat I have applied the patch.

moxvallix:
Actually, hold off on merging this for a bit. There's currently a bug where the death waypoint is getting hidden before the player actually dies and respawns.

I'll push a fix for it later.

moxvallix:
Ok, it should be good to be merged now, unless there are any more changes for me to make.

Wide-Cat:
Looks mostly good but a couple of things. Firstly, apply [this patch](https://github.com/user-attachments/files/22197991/changes.patch) for some small stuff. Secondly, why decide to change the death detection from opening the death screen to death message packets? In both singleplayer and the server I tested it on (`latest-test.ccbluex.net`) it would often duplicate the death waypoint. Is there a tangiable reliability issue it fixes? If not just revert it, I'm pretty sure the previous system was good enough.

java_ap7ITqCWxv
java_gCZ4RhSerK
java_3v51Tpq87Q

moxvallix:
@Wide-Cat I have applied the patch, and reverted the death detection changes.

PR #5690, opened 2025-09-10
Update AutoEat.java

radianeuh:
Also pause Nuker, InfinityMiner and AutoFish

## Type of change

- [X] Bug fix
- [X] New feature

## Description

Pause Nuke, InfinityMiner and AutoFish when eating, i used ai to make this if its okay.

Discarded change for "fixing AutoEat" not pausing baritone correctly, did actually fix but only in very precise scenario. Better to update the Excavator module https://github.com/MeteorDevelopment/meteor-client/pull/5700 than this module, sorry.

## Related issues

i didnt see any issues about that, just needed for myself

# How Has This Been Tested?

Mining for hours on a single player world and 2b2t 

# Checklist:


machiecodes:
First AI pr that isn't completely useless

Wide-Cat:
At this point it would probably be better to just have a setting to choose which modules you want auto eat to pause, and make the defaults the ones in MODULELIST. We already have one like that - `toggleModules` and `modules` in Surround.

radianeuh:
> At this point it would probably be better to just have a setting to choose which modules you want auto eat to pause, and make the defaults the ones in MODULELIST. We already have one like that - `toggleModules` and `modules` in Surround.

okay i will make it like this when i get the time to do it

PR #5700, opened 2025-09-15
feat: Add durability check to Excavator and accept being paused

radianeuh:
Added a new option to stop the module when the main hand tool's durability drops below 100 to prevent breaking.

## Type of change

- [X] Bug fix
- [X] New feature

## Description

Added a new setting, `disable-low-dura`, to Excavator. Will automatically disable itself if the durability of the tool in main hand is below 100.

Also made so Excavator don't force baritone to clear the area on each tick cuz of the onRender3D cuz before that module like AutoEat and AutoGap could not Eat and Gap

## Related issues

N/A

# How Has This Been Tested?

Mining out area with already broken tool in a creative world.

# Checklist:


radianeuh:
im not sure how to make so AutoEat and AutoGap or other module that pause baritone does so Excavator dont force mine everytime and lead you not mining on in loop trying to eat, im just gonna leave it like this.

PR #5704, opened 2025-09-17
per weapon selection for hitboxes module

### LINKED ISSUE #5703: Hitboxes module's only on weapon option doesn't support maces
### Describe the bug

Maces are added to game 1.21 as a new type of weapon, but hitboxes module doesn't recognize mace as a weapon

### Steps to reproduce

1. Turn on hitboxes module
2. Set as "only on weapon"
3. Hold a mace while looking at an entity on debug mode
4. User will notice, hitbox of entity doesn't change at all.

### Meteor Version

Meteor 1.21.8-37

### Minecraft Version

1.21.8

### Operating System

Windows

noramibu:
## Type of change

- [x] Bug fix
- [x] New feature

## Description
I added per weapon selection for only on weapon setting of Hitboxes module, i also added missing mace option as a weapon. 

## Related issues
Fixes #5703 

# How Has This Been Tested?

image

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/Hitboxes.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/Hitboxes.java
@@ -110,7 +110,7 @@ public class Hitboxes extends Module {
 
     private boolean testWeapon() {
         if (!onlyOnWeapon.get()) return true;
-        return InvUtils.testInHands(itemStack -> {
+        return InvUtils.testInMainHand(itemStack -> {
             if (sword.get() && itemStack.isIn(ItemTags.SWORDS)) return true;
             if (axe.get() && itemStack.isIn(ItemTags.AXES)) return true;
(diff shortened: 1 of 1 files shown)


Wide-Cat:
I feel like this isnt really that useful. Is there actually any situation where you would want to have one type of weapon with expanded hitboxes and not another? I think it would just be better to add maces to the only on weapons setting and call it a day.

noramibu:
> I feel like this isnt really that useful. Is there actually any situation where you would want to have one type of weapon with expanded hitboxes and not another? I think it would just be better to add maces to the only on weapons setting and call it a day.

axes. how many times do i need to hit to people while i try to break a log/plank? :(

Wide-Cat:
If you move the only on weapons and the individual weapon settings to a new setting group ill merge this

noramibu:
> If you move the only on weapons and the individual weapon settings to a new setting group ill merge this

done ^^

Issue #5687, opened 2025-09-06
UUID based friends system


Kyuunex:
### Describe the feature

I recently started digging into `friends.nbt` to recovery UUIDs of my friends who changed their names too much while I discovered things that definitely need some improvement.

My `friends.nbt` has about 60 entries, with half of the entries being dupes with different names. Basically, many names share the UUIDs. UUIDs should be unique.

Names with UUIDs of accounts that took the name after the friend dropped the name. Essentially, meteor added UUIDs to my friends list that do not belong to people I friended. 
This should definitely not happen. 

I propose enforcing UUID being unique, and simply update the entry with a new name when it changes. Minecraft accounts cannot change UUIDs, so it is a guaranteed name to track accounts. Not usernames.

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

qnxt:
have a human review this thank you, i dont have perms to reopen

PR #5728, opened 2025-09-30
use uuids for lookup and update usernames

noramibu:
## Type of change

- [x] Bug fix
- [x] New feature

## Description
Uses UUIDs instead of usernames for friends info update/lookup 

Configuration to keep username lookup kept for offline mode environments

## Related issues
#5687 

# How Has This Been Tested?
https://youtu.be/gQRDZIzlGQs

# Checklist:


Wide-Cat:
I don't see the point in making this a config option, just make it default to searching by id and fall back to the name if the id is null

noramibu:
> I don't see the point in making this a config option, just make it default to searching by id and fall back to the name if the id is null

removed config option as requested

Issue #5729, opened 2025-09-30
[Suggestion] Add AutoTool options for blocks without preferred tool


xXBaboBastiXx:
### Describe the feature

AutoTool is unable to swap between tools when mining blocks without preferred tool, such as:
- Glass 
- Sea Lanterns
- Glowstone
- Turtle Eggs
(probably more that are not relevant)

The lack of this feature is **problematic**, because those blocks **don't drop** themself when mined without Silk Touch.
Also, also a **common problem**, for example when you attempt to mine Glass/Sea Lanterns with a Silk Touch Pickaxe already in the hand and Kill Aura changes your hotbar item to the sword to attack Phantoms in the air, the hotbar item remains a sword when you continue to mine Glass, resulting in the Glass breaking without drops.

A **simple fix** for this problem would be adding a selection box for blocks without preferred tool: [Sword/Pickaxe/Axe/Shovel/Sheers]
The player may choose to enchant their preferred tool for those blocks with Silk Touch to mine them and collect them, which is currently not possible.

Another option would be adding a **Silk Touch Whitelist**.
If enabled, AutoTool will always prioritize mining the whitelisted blocks with Silk Touch, if _any tool_ with the Silk Touch enchantment is  in their hotbar currently.
If enabled, AutoTool should attempt to mine blocks that aren't whitelisted (e.g. ores or crops) with Fortune if a Fortune tool is in the hotbar.

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

PR #5730, opened 2025-10-01
implement prefer silk touch to Auto Tool

noramibu:
## Type of change

- [x] Bug fix
- [x] New feature

## Description
removed silk touch for ender chest option and implemented a whitelist for silk touch. If a suitable tool exists and block is in whitelist, it will always prefer silk touch (unless user doesn't have a silk touch tool, in that case it will choose most suitable tool (or fortune tool if fortuneForOresCrops enabled))

also fixed fortuneForOresCrops wasn't working for crops because there was no suitable tool for it (now they simply bypass tool check if fortuneForOresCrops is on)

## Related issues
#5729 #5243 #5983

# How Has This Been Tested?
https://youtu.be/ECl49wCImMs

# Checklist:


src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
```diff
@@ -44,13 +45,21 @@ public class AutoTool extends Module {
         .build()
     );
 
-    private final Setting<Boolean> silkTouchForEnderChest = sgGeneral.add(new BoolSetting.Builder()
-        .name("silk-touch-for-ender-chest")
-        .description("Mines Ender Chests only with the Silk Touch enchantment.")
+    private final Setting<Boolean> silkTouchEnabled = sgSilk.add(new BoolSetting.Builder()
+        .name("silk Touch Whitelist")
```
machiecodes, line 49: Setting names should be in kebab-case

src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
```diff
@@ -44,13 +45,21 @@ public class AutoTool extends Module {
...
+        .description("Require Silk Touch for selected blocks.")
         .defaultValue(true)
         .build()
     );
 
+    private final Setting<List<Block>> silkTouchBlocks = sgSilk.add(new BlockListSetting.Builder()
+        .name("Whitelist Blocks")
+        .description("Blocks that should be mined with Silk Touch.")
```
machiecodes, line 57: Same here

noramibu:
turns out fortune for ores and crops option doesn't work on crops bcs there's simply no suitable tool for it. a simple bypass would do the trick for #5243

noramibu:
this would do the trick

src/main/java/meteordevelopment/meteorclient/systems/modules/player/AutoTool.java
```diff
@@ -260,6 +264,9 @@ && isFortunable(state.getBlock())
         if (itemStack.isIn(ItemTags.SWORDS) && (state.getBlock() instanceof BambooBlock || state.getBlock() instanceof BambooShootBlock))
             score += 9000 + (itemStack.get(DataComponentTypes.TOOL).getSpeed(state) * 1000);
 
+        // Prefer hoes slightly for crops when fortune is enabled and the block is a crop.
+        if (fortuneOre && (state.getBlock() instanceof CropBlock || state.getBlock() instanceof NetherWartBlock) && itemStack.isIn(ItemTags.HOES)) score += 10;
```
noramibu, line 268: there's no point to prefer hoes as default when we don't consider using fortune for crops, thats why its req fortuneore option

PR #5747, opened 2025-10-05
Make onRender2D protected for better module extensibility

S1elDev:
### Summary
This PR changes `onRender2D` in `Nametags.java` from `private` to `protected`.  
No functionality has been changed.

### Why
Making `onRender2D` protected allows addon developers to extend or customize the rendering logic of nametags without copying the entire class.  
Internal helper methods remain private, keeping encapsulation safe and minimizing maintenance risks.


crosby-moe:
that only functions if you intend to add a second nametags module with overlapping features

for modifying existing modules we recommend mixins

crosby-moe:
closing unless there is a use case for this that i have not thought of

PR #5753, opened 2025-10-07
rewrite registry/registry list settings

MukjepScarlet:
## Type of change

- [x] Performance improvement

## Description

We have many registry-based values like Block/Item... But now they are of `List`. This causes issue if we:
- Use `.contains` on them. Typical `List.contains` is of `O(n)` and `Set.contains` is of `O(1)`~`O(log n)`.
- Add same values, `ArrayList` doesn't prevent us to do that.

This PR is in order to:
- Migrate same code of these settings
- Replace List with Set of them

# How Has This Been Tested?

# Checklist:


src/main/java/meteordevelopment/meteorclient/settings/AbstractRegistryItemSetting.java
```diff
@@ -0,0 +1,79 @@
...
+
+        return tag;
+    }
+
+    @Override
+    protected final V load(NbtCompound tag) {
+        value = registry.get(Identifier.of(tag.getString("value", "")));
+
+        if (filter != null && !filter.test(value)) {
+            for (V t : registry) {
```
RacoonDog, line 53: should not fallback to a random valid value from the registry, should fallback to `resetImpl()` instead

src/main/java/meteordevelopment/meteorclient/settings/AbstractRegistryListSetting.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    protected C load(NbtCompound tag) {
+        get().clear();
+
+        NbtList valueTag = tag.getListOrEmpty("value");
+        for (NbtElement tagI : valueTag) {
+            V value = this.registry.get(Identifier.of(tagI.asString().orElse("")));
+
+            if (filter == null || filter.test(value)) get().add(value);
```
RacoonDog, line 83: needs null check on `value`

src/main/java/meteordevelopment/meteorclient/settings/BlockListSetting.java
```diff
@@ -5,93 +5,34 @@
...
+        return new ReferenceOpenHashSet<>(from);
     }
 
-    public static class Builder extends SettingBuilder<Builder, List<Block>, BlockListSetting> {
+    public static class Builder extends SettingBuilder<Builder, Set<Block>, BlockListSetting> {
         private Predicate<Block> filter;
 
         public Builder() {
-            super(new ArrayList<>(0));
+            super(new ReferenceOpenHashSet<>(0));
```
RacoonDog, line 31: `Set#of` or `Collections#emptySet`

src/main/java/meteordevelopment/meteorclient/settings/EnumSetting.java
```diff
@@ -5,43 +5,39 @@
...
     }
 
     @Override
     protected T parseImpl(String str) {
-        for (T possibleValue : values) {
-            if (str.equalsIgnoreCase(possibleValue.toString())) return possibleValue;
-        }
-
-        return null;
+        return nameToValueMap.get(str);
```
RacoonDog, line 29: map is ordered case insensitively, but this operation is case sensitive

src/main/java/meteordevelopment/meteorclient/settings/EnumSetting.java
```diff
@@ -5,43 +5,39 @@
...
     @Override
     protected boolean isValueValid(T value) {
         return true;
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public List<String> getSuggestions() {
-        return suggestions;
+        return (List<String>) (Object) List.of(nameToValueMap.keySet().toArray());
```
RacoonDog, line 40: unnecessary copying, just `new ArrayList<>(nameToValueMap.keySet())`
and for future reference, the generic `List#toArray` methods aren't any slower and dont require unchecked casts

RacoonDog (changes requested):
since switching the setting types is a very breaking change, pr should target the `mc-update` branch rather than `master`

src/main/java/meteordevelopment/meteorclient/settings/AbstractRegistryItemSetting.java
RacoonDog: huh yeah that’s scary why do we do that??
but yeah that’s better, just check if `value` is null

src/main/java/meteordevelopment/meteorclient/settings/EnumSetting.java
RacoonDog: yeah but `Map#get` relies on `String#equals` and `String#hashCode`, which are case sensitive

MukjepScarlet:
I think I messed up the commit history of this branch...

Wide-Cat:
you tried to merge this into the mc-update branch

MukjepScarlet:
I'll try to clean it later

MukjepScarlet:
Okay now cleaned

PR #5330, opened 2025-04-14
Remove Shadow JAR dependency

Big-Iron-Cheems:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Remove Shadow JAR dependency, in favour of JAR-in-JAR packaging.

## Related issues

None.

# How Has This Been Tested?

N/A.

# Checklist:

- [x] My code follows the style guidelines of this project.
- [x] I have added comments to my code in more complex areas.
- [x] I have tested the code in both development and production environments.

PR #5754, opened 2025-10-07
Replace Shadow JAR with proper JiJ implementation

Big-Iron-Cheems:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Replace Shadow JAR with proper JiJ implementation.
This is a rewrite of the now closed #5330 PR, updated to the new 1.21.9 branch.
The implementation relies on [this](https://github.com/FlorianMichael/BaseProject/blob/main/src/main/kotlin/de/florianmichael/baseproject/Fabric.kt) source for its transitive dependency handling.

## Related issues

None.

# How Has This Been Tested?

Tested by loading the produced JAR in a fresh Fabric instance via PrismLauncher.

# Checklist:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,7 +1,6 @@
 plugins {
     id("fabric-loom") version "1.11-SNAPSHOT"
     id("maven-publish")
-    id("com.gradleup.shadow") version "9.2.2"
 }
 
 base {
@@ -50,7 +49,7 @@ repositories {
 }
 
 val modInclude: Configuration by configurations.creating
-val library: Configuration by configurations.creating
+val jij: Configuration by configurations.creating
 
 configurations {
     // include mods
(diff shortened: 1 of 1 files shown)


MukjepScarlet:
image


Game-provided libs should be excluded.

https://github.com/CCBlueX/LiquidBounce/blob/0f4b9ba3c1b254180168aece0be405d7a4abfec1/build.gradle.kts#L68

MukjepScarlet:
And `jsr305` might be compileOnly? I don't know which dependency uses it...

Big-Iron-Cheems:
> 
> image
> 
> 
> Game-provided libs should be excluded.
> 
> https://github.com/CCBlueX/LiquidBounce/blob/0f4b9ba3c1b254180168aece0be405d7a4abfec1/build.gradle.kts#L68

Good catch, I see that the older Shadow impl did exclude this.
Relatively easy fix, will do in a later commit.
Perhaps we can look into excluding other deps if need be.

Issue #5495, opened 2025-06-15
Maintain AutoClicker interactions when moving camera in Freecam


diegogarciarojo:
### Describe the feature

**Describe the feature**  
When using the AutoClicker module to click on an armor stand, the clicks stop registering on the armor stand if the player activates Freecam and moves the camera away, even though the clicking animation continues and the player is still facing the armor stand from their real position. I suggest adding an option in the AutoClicker settings to keep interacting with the targeted armor stand based on the player's actual position and facing direction, regardless of the Freecam camera's position. This would ensure that AutoClicker continues clicking the armor stand when the camera moves in Freecam mode.

**Additional info**  
This feature would improve the consistency of AutoClicker when used with Freecam, especially for tasks like repeatedly interacting with armor stands or similar entities while exploring in Freecam mode. 

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

31a05b9c:
duplicate of #5168
apparently chri-k has some 1-week-old changes to freecam which amongst other things address this. will look at those changes later to figure out why they aren't in a PR

eikhren:
lol can't believe this is still actually an issue

Issue #5168, opened 2025-02-02
Keep the character position as target for AutoClicker in Freecam


Mark7888:
### Describe the feature

When starting an `Auto Clicker` facing an entity, it starts hitting it. But when I go into `Freecam` and look away, it stops. I figured out, that it clicks on the blocks and hits the entities I'm facing with the Freecam, not the ones I have positioned my character at.

Is there any possibility to add a switch in the `Auto Clicker` settings menu to switch between targeting the `Freecam` cursor, or targetting the real character position?

_Meteor-1.21.4-14_

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

diegogarciarojo:
so?

Issue #4663, opened 2024-06-19
Add Option to Disable GUI Vertical Movement in Freecam


IsaacDawg97:
### Describe the feature

Add a checkbox in GUI Move to toggle Shift (and maybe Space) moving the player vertically while in Freecam.

Little annoying being moved down in Freecam when using Shift to move items from Inventory to Hotbar.
I know that I can press Shift and Space to cancel them out, but still would be a nice QoL feature.

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

PR #5767, opened 2025-10-12
questionable freecam improvements

chri-k:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

- compatibility between `AirPlace` and `Freecam`
- add player-relative position mode
- option to not modify the crosshair target when freecam is active
- option to ignore shift key in GUIs (since that is unaffected by `GuiMove.sneak`)
- held keybind to temporarily switch to moving the player instead of the camera
- remove chaotic rotation when window is not focused

## Related issues

#5168 ( duplicated by #5495 )
#4663

p.s.: close all of the duplicate issues about `Freecam` causing crashes in the end dimension

# How Has This Been Tested?

multiple times

# Checklist:


diegogarciarojo:
@chri-k  Regarding the Preserve Crosshair Target, there should be an option so that it is only used in autoclicker mode, while you can still manually open things by right-clicking in freecam on the block you are looking at, whereas in the autoclicker it would be on the player's "cursor", and not the freecam's.

diegogarciarojo:
@chri-k

PR #5777, opened 2025-10-18
fix: criticals

cattyngmd:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

fixed some issues with criticals module

# Checklist:


MineGame159:
I am testing it right now and it fixes mini jump not attacking while in a cobweb at all but both jump and mini jump do not deal critical damage. Or is that not what you were trying to fix?

cattyngmd:
yea, it was trying to jump in cobweb to deal critical damage which is not possible

PR #5783, opened 2025-10-19
Freecam click to path

sofia-nep:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

This is similar to a PR I opened a while ago with a banned account, it allows you to path to a location while in Freecam by pointing at it and clicking (optionally double clicking). This is something I rather enjoyed in Lambda, so I figured it would be worth implementing it here.

## Related issues

N/A as far as I'm aware.

# How Has This Been Tested?

I've been using a local branch with this on 2b for a few months.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Freecam.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/Freecam.java
@@ -50,6 +50,7 @@ import org.lwjgl.glfw.GLFW;
 
 public class Freecam extends Module {
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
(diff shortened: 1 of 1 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/render/Freecam.java
```diff
@@ -131,6 +154,9 @@ public class Freecam extends Module {
 
     private boolean forward, backward, right, left, up, down, isSneaking;
 
+    private boolean clicked = false;
```
crosby-moe: Redundant variable, it is never reset to `false` after being set to `true`

PR #5792, opened 2025-10-20
Ignore creative in KillAura.java

BerryRock0:
On some servers, Killaura recognizes a survival mode player as a creative mode player. This patch allows client user to ignore targets gamemode.


31a05b9c:
servers only ever do that for anticheat purposes for legitimate pvp

zyrafaq:
> servers only ever do that for anticheat purposes for legitimate pvp

Yeah and this is a *hacked* client, so our goal is to hack

crosby-moe:
> Yeah and this is a _hacked_ client, so our goal is to hack

meteor is an anarchy client, cheat with it will get you banned

PR #5795, opened 2025-10-21
Option to disable Middle Click Action in Creative mode

moxvallix:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds an option to Middle Click Extra, that disables it in Creative mode. This is so that it doesn't interrupt pick-blocking when building. The option is set to be on by default.


lachameleon:
Can we please not make this default. I am a touchpad user

lachameleon:
@Wide-Cat

moxvallix:
@lachameleon not really sure how this relates to being a touchpad user tbh.

PR #5808, opened 2025-10-27
Added ImageHUD

Jalvaviel:
## Type of change

New feature

## Description

A new HUD element that can draw still images and animated images from paths/urls in the in-game HUD

## Related issues

Not related

# How Has This Been Tested?

https://private-user-images.githubusercontent.com/51824929/499489242-eafb079f-3120-4a0a-a395-710eba5c37f6.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjE1ODUxOTEsIm5iZiI6MTc2MTU4NDg5MSwicGF0aCI6Ii81MTgyNDkyOS80OTk0ODkyNDItZWFmYjA3OWYtMzEyMC00YTBhLWEzOTUtNzEwZWJhNWMzN2Y2LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTEwMjclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUxMDI3VDE3MDgxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTIwODZjYjVhMzg1ZGZlNTQ1ZjQ3MTJmOTU5MDFiNjM3NzVhMGY1YzA4OGFkYTUzYmM5ZTEwYjI2NDYwOTg4ZDMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.uLKzeuOsSQi2bDrQsm1wq6Ngdl-ffWDvecIbce1UZy0

https://private-user-images.githubusercontent.com/51824929/499490549-0e8c5b15-a82c-4a0e-bce6-15f0de793787.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjE1ODUxOTEsIm5iZiI6MTc2MTU4NDg5MSwicGF0aCI6Ii81MTgyNDkyOS80OTk0OTA1NDktMGU4YzViMTUtYTgyYy00YTBlLWJjZTYtMTVmMGRlNzkzNzg3LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTEwMjclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUxMDI3VDE3MDgxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTUwOGQ3MGRhNDk1ZmE1Zjg0NmM3NDM4ZjMwYjkzN2JhNjI3OTE3OTM5NWFhNmU4ZTIwMzFiOGJjMDBiNDEyOWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.7CU1wvuNc6_i1yUjKb4RZEL6zfKf6aJB9630zqXdOiM

# Checklist:

- ✔ My code follows the style guidelines of this project.
- ✔ I have added comments to my code in more complex areas.
- ✔ I have tested the code in both development and production environments.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/sodium/SodiumBlockRendererMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/sodium/SodiumBlockRendererMixin.java
@@ -7,20 +7,49 @@ package meteordevelopment.meteorclient.mixin.sodium;
 
 import meteordevelopment.meteorclient.systems.modules.render.Xray;
 import net.caffeinemc.mods.sodium.client.render.chunk.compile.pipeline.BlockRenderer;
+import net.caffeinemc.mods.sodium.client.render.chunk.terrain.material.DefaultMaterials;
+import net.caffeinemc.mods.sodium.client.render.chunk.terrain.material.Material;
+import net.caffeinemc.mods.sodium.client.render.model.MutableQuadViewImpl;
 import net.minecraft.block.BlockState;
 import net.minecraft.client.render.model.BlockStateModel;
 import net.minecraft.util.math.BlockPos;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.ModifyArg;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
-@Mixin(value = BlockRenderer.class)
-public abstract class SodiumBlockRendererMixin {
-    @Inject(method = "renderModel", at = @At(value = "INVOKE", target = "Lnet/caffeinemc/mods/sodium/client/model/color/ColorProviderRegistry;getColorProvider(Lnet/minecraft/block/Block;)Lnet/caffeinemc/mods/sodium/client/model/color/ColorProvider;", shift = At.Shift.AFTER), cancellable = true)
-    private void onRenderModel(BlockStateModel model, BlockState state, BlockPos pos, BlockPos origin, CallbackInfo ci) {
-        int alpha = Xray.getAlpha(state, pos);
+@Mixin(value = BlockRenderer.class, remap = false)
+public class SodiumBlockRendererMixin {
+    @Unique
+    private int xrayAlpha;
 
-        if (alpha == 0) ci.cancel();
+    @Inject(method = "renderModel", at = @At("HEAD"), cancellable = true)
+    private void onRenderModel(BlockStateModel model, BlockState state, BlockPos pos, BlockPos origin, CallbackInfo info) {
+        xrayAlpha = Xray.getAlpha(state, pos);
+
+        // Cancel block rendering when alpha is 0, required for Iris support but unnecessary to check for shaders, we already force be disabled when Xray is enabled
+        if (xrayAlpha == 0) {
+            info.cancel();
+        }
+    }
+
+    @Inject(method = "bufferQuad", at = @At("HEAD"))
+    private void onBufferQuad(MutableQuadViewImpl quad, float[] brightnesses, Material material, CallbackInfo info) {
+        if (xrayAlpha != -1) {
+            for (int i = 0; i < 4; i++) {
+                int color = quad.baseColor(i);
+                quad.setColor(i, ((xrayAlpha & 0xFF) << 24) | (color & 0x00FFFFFF));
+            }
+        }
+    }
+
+    @ModifyArg(method = "processQuad", at = @At(value = "INVOKE", target = "Lnet/caffeinemc/mods/sodium/client/render/chunk/compile/pipeline/BlockRenderer;bufferQuad(Lnet/caffeinemc/mods/sodium/client/render/model/MutableQuadViewImpl;[FLnet/caffeinemc/mods/sodium/client/render/chunk/terrain/material/Material;)V"), index = 2)
+    private Material modifyMaterial(Material material) {
+        if (xrayAlpha != -1) {
+            return DefaultMaterials.TRANSLUCENT;
+        }
+        return material;
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/meteordevelopment/meteorclient/systems/hud/elements/ImageHud.java
```diff
@@ -0,0 +1,171 @@
...
+
+public class ImageHud extends HudElement {
+    public static final HudElementInfo<ImageHud> INFO = new HudElementInfo<>(Hud.GROUP, "image", "Cures your ADHD.", ImageHud::new);
+    public static final int MAX_TEX_SIZE = glGetInteger(GL_MAX_TEXTURE_SIZE);
+    private static final Identifier DEFAULT_TEXTURE = Identifier.of(MOD_ID,"textures/icons/gui/default_image.png");
+    private static final Identifier LOADING_TEXTURE = Identifier.of(MOD_ID,"textures/icons/gui/loading_image.png");
+    private static final Color TRANSPARENT = new Color(255, 255, 255, 255);
+    private static final int DEBOUNCE_TIME = 2; // 2 Seconds before rerunning.
+    private Identifier texture;
+    private final NameableExecutor worker = Util.getIoWorkerExecutor();
```
crosby-moe: executor should use daemon threads, `Util#getIoWorkerExecutor` does not daemon

src/main/java/meteordevelopment/meteorclient/utils/misc/texture/TextureUtils.java
```diff
@@ -0,0 +1,57 @@
...
+        mc.getTextureManager().registerTexture(identifier, texture);
+        return identifier;
+    }
+
+    /**
+     * Registers the texture from an imageData object parsing for invalid chars.
+     * @return Identifier of the registered texture.
+     */
+    public static NativeImage bufferedToNative(BufferedImage canvas) {
+        NativeImage tex = new NativeImage(canvas.getWidth(), canvas.getHeight(), true);
```
crosby-moe: 3rd parameter should be `false` since this `NativeImage` does not use stbi

crosby-moe (changes requested):
there are 2 extra major underlying issues with this pull request:

firstly, since the stored `NativeImageBackedTexture` is only destroyed inside `ImageHud`, deleting an `ImageHud` element from the hud editor causes its texture to be lost without closing, leaking both ram and vram

secondly, i highly dislike the usage of `javax.imageio` and `java.awt` for many reasons:
1. `BufferedImage` operations are *very* expensive relative to minecraft's `NativeImage`
2. `java.awt` has been known to cause many issues with minecraft in the past, it would be better to avoid touching it at all
3. `BufferedImage` errors are hard to diagnose :(

minecraft already has the necessary support for parsing jpegs, pngs, bmps, gifs, etc. through stbi, and you can easily modify `NativeImage`s in the ways you need with `NativeImage#fillRect` and `NativeImage#copyRect`

crosby-moe:
~~I'm also unsure about the idea of packing a gif into a single texture, it sounds like that would either quickly hit some upper limit or take up a massive amount of vram~~

it's fine, anything else would be complicated to implement and probably be out of scope for meteor, though the size of the "atlas" should be checked before uploading, with anything too big either rejected or resized to fit

elijah-yum:
LGTM 🚀🚀🚀

src/main/java/meteordevelopment/meteorclient/systems/hud/elements/ImageHud.java
```diff
@@ -0,0 +1,171 @@
...
+    private ImageData cachedImageData;
```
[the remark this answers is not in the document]
Jalvaviel: All of those variables are modified in the main thread, since texture registration doesn't work in other threads that aren't the main one.

Jalvaviel:
> ~I'm also unsure about the idea of packing a gif into a single texture, it sounds like that would either quickly hit some upper limit or take up a massive amount of vram~
> 
> it's fine, anything else would be complicated to implement and probably be out of scope for meteor, though the size of the "atlas" should be checked before uploading, with anything too big either rejected or resized to fit

The atlas is automatically resized by getting the max texture size from OpenGL, and wraps columnwise if it reaches that limit. Maybe a hard limit should be imposed, but I don't know what would be a reasonable one.

src/main/java/meteordevelopment/meteorclient/utils/misc/texture/ImageData.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package meteordevelopment.meteorclient.utils.misc.texture;
+
+import it.unimi.dsi.fastutil.ints.IntList;
+import net.minecraft.client.texture.NativeImage;
+
+import java.util.List;
+
+public class ImageData {
```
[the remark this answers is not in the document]
Jalvaviel: This breaks the factory methods and SRP (The ImageData class shouldn't know how it was made, only what it needs to do). Also splitting ImageData into two records means that the fields are immutable and can't be changed later.
Also what should the methods in the interface be? Only getColumns is actually used in the Animated Image Data case.

src/main/java/meteordevelopment/meteorclient/systems/hud/elements/ImageHud.java
MukjepScarlet: You can use virtual thread here. (if u want daemon)

Issue #5828, opened 2025-11-05
[BUG] AutoLog: fake logout


Kelvinlby:
### Describe the bug

The AutoLog module often get our client logged out, but the server still think we are connected. When connecting again, it shows "You've already joined". This was happening very often on 2b2t.org.

### Steps to reproduce

1. Login to 2b2t.org
2. Switch on AutoLog
3. Switch off "toggle off"
4. Meet a situation to trigger AutoLog
5. You get logged out successfully.
6. Then login again instantly
7. Keep staying in the same situation to trigger AutoLog
8. Then fake logout: you seems logged out on client side, but when you tries to rejoin, server shows you've already joined!
> I've died 2 times because of this fake logout bug.

### Meteor Version

1.21.10-22

### Minecraft Version

1.21.10

### Operating System

Linux

### Before submitting a bug report

- [x] This bug wasn't already reported (I have searched bug reports on GitHub).

- [x] This is a valid bug (I am able to reproduce this on the latest dev build).

Kelvinlby:
After examine the code, i find that original logout logic relies on simulating a server-to-client disconnect packet. I think we should use client side disconnection to guarantee true logout.

0xTas:
Afaik this is a 2b2t-(or perhaps folia)-specific issue. It is easily solved by forcing the server to kick you instead of relying on a traditional disconnect. In most paid clients it is known as an "Illegal Disconnect". My stardust addon is designed for use on 2b2t, and adds this ability as an option to Meteor's built-in AutoLog module, among others.

Issue #5827, opened 2025-11-05
[Suggestion] AutoLog: support logout when certain coordinate reached


Kelvinlby:
### Describe the feature

When AFK flying on 2b2t highways, we always want to stop when reaching our destination. However, currently no module supports auto logout based on coordinate, so i think it's useful to add one. 

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

PR #5829, opened 2025-11-05
Solve fake logout and implement coordinate-based auto logout in AutoLog module

Kelvinlby:
Added position checking settings to AutoLog module.

## Type of change

- [X] Bug fix
- [x] New feature

## Description

Added new function to AutoLog module: now player can auto logout when certain coordinate reached.

## Related issues

[[Suggestion] AutoLog: support logout when certain coordinate reached #5827](https://github.com/MeteorDevelopment/meteor-client/issues/5827)
[[BUG] AutoLog: fake logout #5828
](https://github.com/MeteorDevelopment/meteor-client/issues/5828)

# How Has This Been Tested?

1. Coordinate based logout
Screenshot From 2025-11-05 08-28-04
2. Fake logout seems to be solved because i could not reproduce it again, so I'm not able to take a screenshot.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/AutoLog.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/AutoLog.java
@@ -70,14 +70,14 @@ public class AutoLog extends Module {
     );
 
     private final Setting<Boolean> onlyTrusted = sgGeneral.add(new BoolSetting.Builder()
-        .name("friends-only")
+        .name("only-trusted")
         .description("Disconnects when a player not on your friends list appears in render distance.")
         .defaultValue(false)
         .build()
     );
 
     private final Setting<Boolean> instantDeath = sgGeneral.add(new BoolSetting.Builder()
-        .name("threat-of-instant-death")
+        .name("32K")
         .description("Disconnects when a player near you can instantly kill you.")
         .defaultValue(false)
         .build()


crosby-moe (changes requested):
it is discouraged to change setting names even if they are bad as it breaks existing configs (it is preferred that you open an issue with the proposed names, we will be making a system for changing them safely sometime in the future)

there is no functional difference between calling `ClientPlayNetworkHandler#onDisconnect` and `ClientWorld#disconnect` as they both defer to the same `ClientConnection#disconnect` function

unless there is a specific reason for having them as settings, couldn't the values of `checkXRelation`, `checkYRelation`, and `checkZRelation` be automatically calculated when the module is enabled based on the difference between the current position and the target position?

Kelvinlby:
> it is discouraged to change setting names even if they are bad as it breaks existing configs (it is preferred that you open an issue with the proposed names, we will be making a system for changing them safely sometime in the future)
> there is no functional difference between calling ClientPlayNetworkHandler#onDisconnect and ClientWorld#disconnect as they both defer to the same ClientConnection#disconnect function
> unless there is a specific reason for having them as settings, couldn't the values of checkXRelation, checkYRelation, and checkZRelation be automatically calculated when the module is enabled based on the difference between the current position and the target position?

I've changed the name of setting items back.

I don't know why but the original implementation frequently causes fake logout on 2b2t.org, especially when you tries to rejoin immediately. 

Yes there does have reasons for explicitly define X/Y/Z coord relations: if player have "auto reconnect" switched on, and "toggle off" switched off, and logged out very close to the configured coord, it could make it bouncing back and forth on the configured border, especially when ebouncing with extremely high speed.

crosby-moe:
> I don't know why but the original implementation frequently causes fake logout on 2b2t.org, especially when you tries to rejoin immediately.

in some cases the server can take a bit before processing and finalizing the disconnect, a working alternative would be to cause the server to kick the player instead of logging out, though i'm not an expert on that

also, why was the pull request closed?

Kelvinlby:
I need some time to figure it out. I'll create a new one when i'm ready.

0xTas:
Keep in mind these features do already exist in multiple addons meant for use on 2b2t, like stardust, jeff mod, and probably bephax too. Stardust also has the ability to force the server to kick you, which is especially important on 2b2t because of a fairly common bug where regular disconnects do not actually remove your player from the server world for some time even though your client has already left.

If meteordev wants these features added to the client proper, then there's probably no real reason to reinvent the wheel here.

Issue #5874, opened 2025-11-23
[Suggestion] Add .enemy list like .friend like in "Rusher hack"


towux:
### Describe the feature

And selections in killaura "Only enemy" e.t.c
And enemy color

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

beak2825:
There is now code for this addon.
Check https://github.com/MeteorDevelopment/meteor-client/pull/5876/checks
And go to Upload Artifacts
It is the JAR mod file for 1.21.10

Issue #3767, opened 2023-06-23
[Suggestion]  Enemy Tab


Stayrsxv:
### Describe the feature

Enemy Tab would be the same thing as Friends but it alerts you if these player entire you vincinity and these players are targeted first with killaura

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

Issue #3629, opened 2023-05-04
[Suggestion] Tab idea: Target people


Darknicomars:
### Describe the feature

My Idea is simple, let users to "target" people, so in the tab they appears in red

It can be really useful on big servers, and quick for people to know if there enemies are connected or not

![image](https://user-images.githubusercontent.com/91838105/236206388-d39e4316-e34e-4070-a7cf-6f7ffed49064.png)

### Before submitting a suggestion

- [X] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [X] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

I-Got-You-Dead2:
Even better, make all module targets check for high priority targets

I-Got-You-Dead2:
Pretty sure the vector addon had something similar to this

ImInTheICU:
+rep 
Good idea, you could try getting the dudes over at https://github.com/Nekiplay/MeteorPlus to add it.

I-Got-You-Dead2:
Meteor plus genuinely has the worst code quality possible

I-Got-You-Dead2:
Use meteor rejects

PR #5876, opened 2025-11-23
[feat] Enemies

chri-k:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

adds an enemies list and moderately refactors target selection. 

`TargetUtils` and `Friends` are deprecated and replaced with `systems.targeting.Targeting`.
(it might be better to delete them entirely, since the substitution subtly modifies behaviour)

future modifications to target selection (per-entity-type priorities would be useful) should not disrupt existing code as much as a result.

this is opened with the purpose of collecting feedback on both the implementation and the UX.

a bit of UI is missing. config migration code (for every single module setting which got expanded) is missing and not planned unless crosby's setting migration thing happens. 

**warning:** this was produced near midnight.

## Related issues

#5874 
#3629
#3767

# How Has This Been Tested?

currently completely untested. since a very large amount of modules was affected and this is only testable in a multiplayer context, testing is not trivial.

# Checklist:

(draft)

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/Commands.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/Commands.java
@@ -38,6 +38,7 @@ public class Commands {
         add(new EnchantCommand());
         add(new FakePlayerCommand());
         add(new FriendsCommand());
+        add(new EnemiesCommand());
         add(new CommandsCommand());
         add(new InventoryCommand());
         add(new NbtCommand());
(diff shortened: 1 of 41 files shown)


noramibu:
Isn't it better to implement a relation system instead of having separate enemies and friends system? So people can organize their own relation groups and configure modules for each relation group? 

It can be useful for groups where people have multiple groups or would like to share groups list with each other.

31a05b9c:
that would be a highly complicated system which only differs in effect on `ESP` and `Tracers` (and `BetterTab` ig)

(that is also bad wording, because the word "system" has special meaning in this codebase, and there is only one of those here)

31a05b9c:
and the colour space of the various ESP modules is already cluttered enough for it to be hard to differentiate things, so this is questionably useful

31a05b9c:
however there should be an option to save this to profiles

PR #5888, opened 2025-11-30
No interact anti miss

Marccccccccccccccc:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Added a "missed-attacks" toggle to No Interact to cancel attacks that would have missed and reset attack cooldown.
The Hand will still swing Client-Side
## Related issues

None

# How Has This Been Tested?

https://youtu.be/kpgbJqId6h0

# Checklist:

- [X ] My code follows the style guidelines of this project. (I think so)
- [X ] I have added comments to my code in more complex areas.
- [X ] I have tested the code in both development and production environments.


31a05b9c:
this does not really fit into no-interact, it should probably be a separate module
(no players will discover this feature)

said module could also modify the attack cooldown on successful hits( sometimes useful on specific servers )

i-am-called-glitchy:
lgtm (i have 0 java experience)

src/main/java/meteordevelopment/meteorclient/mixin/MinecraftClientMixin.java
```diff
@@ -116,9 +121,17 @@ private void onTick(CallbackInfo info) {
...
+    @Inject(method = "doAttack", at = @At("HEAD"), cancellable = true)
     private void onAttack(CallbackInfoReturnable<Boolean> cir) {
         CPSUtils.onAttack();
+
+        NoInteract noInteract = Modules.get().get(NoInteract.class);
+
+        if (noInteract.isActive() && noInteract.shouldCancelMissedAttacks() &&
+            crosshairTarget != null && crosshairTarget.getType() == HitResult.Type.MISS) {
+            cir.setReturnValue(false);
+            player.swingHand(Hand.MAIN_HAND, false);
```
chri-k:
```suggestion
            player.swingHand(Hand.MAIN_HAND, false);
            cir.setReturnValue(false);
```
(`setReturnValue` immediately exits the function)

31a05b9c:
don't write useless comments like this, you are sending notifications (and probably also e-mail) to everyone else in the conversation

RuttyboyzCode:
like what 31a said i think its better for it to be a separate module rather than no interact

Marccccccccccccccc:
unnecessary comment again 💩

Marccccccccccccccc:
anyway what should the module be called?

31a05b9c:
> anyway what should the module be called?

Future calls its copy of this `NoHitDelay`, which is the technically correct name, but `NoMissDelay` might be clearer.

(and this is very clearly a combat feature, which is also another reason for it to not fit into no-interact)

PR #5905, opened 2025-12-05
Improve stash hunter

JanikGeist:
## Type of change

- [ ] Bug fix  
- [x] New feature

## Description

The stash finder previously flagged many trial chambers as stashes due to the high density of containers like chests, barrels, hoppers, and dispensers. This update improves detection accuracy and introduces several quality-of-life enhancements, including more flexible visual indicators and direct pathing from chat.

image

### Blacklisted Support Blocks
- Chests placed on newly blacklisted support blocks are excluded from the detection threshold, reducing false positives.
image

### Tracers
- When enabled, a tracer points toward the center of the chunk where a stash was detected.
- Tracer color, draw distance, and auto-removal behavior are configurable.
- Tracers can be toggled individually for each stash in the stash list.
image

### Tracer Columns
- When enabled, a vertical column appears at the stash’s chunk center.
- Columns share the same configuration options as tracers (color, draw distance, auto-removal).
- Columns can also be toggled per stash in the stash list.

### Direct Pathing
- Chat messages now include clickable coordinates.  
- If a `PathManager` is available, clicking will automatically initiate pathing to the stash location.
image

## How Has This Been Tested?

- Manually tested in-game

## Checklist

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/StashFinder.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/StashFinder.java
@@ -53,6 +53,17 @@ public class StashFinder extends Module {
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
     private final SettingGroup sgRender = settings.createGroup("Render");
 
+    private static final List<Block> DEFAULT_SUPPORT_BLOCK_BLACKLIST = List.of(
+        Blocks.OXIDIZED_COPPER,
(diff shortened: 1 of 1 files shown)


Wide-Cat:
Can you add the relevant trial chamber blocks as the default value for the blacklisted support blocks setting? That seems like the primary use case for the setting and it would make more sense to have it by default, rather than enforcing users to add the blocks themselves. If you do that I'll likely merge this.

PR #5928, opened 2025-12-10
implement better locator

noramibu:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds a new module called better locator, displays player heads instead of dots, shows names, distance to player and directions.

## Related issues
None

# How Has This Been Tested?
https://www.youtube.com/watch?v=R-4hEjPHxhc

# Checklist:


crosby-moe:
i dont really like the blanket `@Overwrite`ing of two render functions in the locator class, would it be possible to use many smaller, more compatible injections instead?

PR #5933, opened 2025-12-13
TheElenium: BLockESP: Added support for NBT-Data filtering

TheElenium:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Added a feature to the BlockESP module to allow players to filter searched block by nbt data (e.g. waterlogged, facing=north/west etc.). Mainly implemented it, because i wanted an easier way to find ominous trial vaults, but the normal esp modules had no option to distinguish between a normal and an ominous, since they are the same block (minecraft:vault).

## Related issues

None

# How Has This Been Tested?

image
image

I created some blocks that only differ in the nbt data and filtered the in the blockesp module for those tags in the specific block configs

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/render/blockesp/BlockESP.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/render/blockesp/BlockESP.java
@@ -42,89 +42,60 @@ public class BlockESP extends Module {
 
     // General
 
-    private final Setting<List<Block>> blocks = sgGeneral.add(
-        new BlockListSetting.Builder()
-            .name("blocks")
-            .description("Blocks to search for.")
-            .onChanged(blocks1 -> {
-                if (isActive() && Utils.canUpdate()) onActivate();
-            })
-            .build()
+    private final Setting<List<Block>> blocks = sgGeneral.add(new BlockListSetting.Builder()
+        .name("blocks")
+        .description("Blocks to search for.")
+        .onChanged(blocks1 -> {
+            if (isActive() && Utils.canUpdate()) onActivate();
+        })
+        .build()
     );
-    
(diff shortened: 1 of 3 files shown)


src/main/java/meteordevelopment/meteorclient/systems/modules/render/blockesp/BlockESP.java
```diff
@@ -25,68 +25,106 @@
...
+            .build()
     );
 
     private final BlockPos.Mutable blockPos = new BlockPos.Mutable();
 
-    private final Long2ObjectMap<ESPChunk> chunks = new Long2ObjectOpenHashMap<>();
+    private final Map<
+        Block,
+        Map<Property<?>, Comparable<?>>
+    > activeFilterCache = new HashMap<>();
```
crosby-moe: this map is always empty

src/main/java/meteordevelopment/meteorclient/systems/modules/render/blockesp/ESPBlockDataScreen.java
```diff
@@ -110,6 +113,20 @@ public void initWidgets() {
...
+            .description("Filters with states (e.g. 'waterlogged=false', 'facing=north', 'ominous=true'). Only blocks matching ALL filters will be shown.")
+            .defaultValue(new ArrayList<>())
+            .onModuleActivated(stringSetting -> stringSetting.set(blockData.stateFilters))
+            .onChanged(filters -> {
+                blockData.stateFilters.clear();
+                blockData.stateFilters.addAll(filters);
+                onChanged();
+            })
+            .build()
+        );
```
crosby-moe: imho using a string list for this is super unintuitive, it should be a separate setting type

PR #5934, opened 2025-12-13
added customisation towards the notify setting in MiddleClickExtra

RuttyboyzCode:
some servers uses different prefixes for msging players so i added a customizable prefix that isn't hardcoded 

## Type of change

- [ ] Bug fix
- [ ✔️] New feature

## Description

A summary of the changes along with the reasoning behind the changes.

## Related issues

# How Has This Been Tested?

# Checklist:

- [ ✔️] My code follows the style guidelines of this project.
- [ ✔️] I have added comments to my code in more complex areas.
- [ ✔️] I have tested the code in both development and production environments.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/MiddleClickExtra.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/MiddleClickExtra.java
@@ -41,9 +41,17 @@ public class MiddleClickExtra extends Module {
     );
 
     private final Setting<Boolean> message = sgGeneral.add(new BoolSetting.Builder()
-        .name("message")
-        .description("Sends a message to the player when you add them as a friend.")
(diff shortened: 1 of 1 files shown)


noramibu:
isn't it better to allow editing whole command (including message) instead of only editing command? so people can customize messages too

RuttyboyzCode:
Oh yeah, that's a much better idea

DrParanoya:
~~Typo on the description of the `notifyMessage` setting:
`for the player's name` should be without an apostrophe~~

crosby-moe:
> Typo on the description of the `notifyMessage` setting: `for the player's name` should be without an apostrophe

it looks correct to me

DrParanoya:
> it looks correct to me

hmm I always thought there's no need for an apostrophe when writing stuff like that
guess there's no typo then

PR #5970, opened 2025-12-23
various esp fixes & refactors

crosby-moe:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

Fixes:
- ESP glow no longer renders behind translucent blocks
- ESP alpha fade distance now calculates from the center of the entity instead of the corner
- Remove redundant framebuffer variable in storage outline shader

also refactored the post processing code to enforce `final` usage & remove dead code

## Related issues

somehow never reported before lmao

# How Has This Been Tested?

the J

# Checklist:


Wide-Cat:
java_PBFTi6OLuN

The alpha fade distance setting is still broken for shader mode, while we are here it would be good to fix.

Wide-Cat:
> java_PBFTi6OLuN
> 
> The alpha fade distance setting is still broken for shader mode, while we are here it would be good to fix.

have been told a pr will come for this later

PR #5972, opened 2025-12-23
ensure all blocks have renders & distinct names in gui

crosby-moe:
## Type of change

- [X] Bug fix
- [ ] New feature

## Description

for the renders:
- all blocks with no item models, but usable block models have had an item model autogenerated (under the `meteor-client` namespace to prevent funny business)
- all blocks with no items models or block models have been manually mapped to an associated item model (except air)
- some blocks with better block models than item models have also been manually replaced (see lava cauldron)

for the names, for some reason the 'Wall' part of blocks is always omitted
unfortunately, this replacement can only be done in english languages (because it was manually mapped), so support for other languages will probably be added some time after meteor supports translations

## Related issues

none

# How Has This Been Tested?

the J

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/widgets/WBlock.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/widgets/WBlock.java
@@ -39,7 +39,7 @@ public class WBlock extends WWidget {
 
     protected BlockState state;
     protected boolean initialized;
-    protected boolean chached;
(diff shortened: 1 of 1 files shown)


crosby-moe:
i also looked into dynamically generating item models as needed instead of adding them all to the repo, but that is unfortunately very difficult to get working (something something data dependencies & having to first load all referenced block models)

PR #6006, opened 2025-12-31
Fully working Elytra Target

bbbPriv:
Fully working Elytra Target, works on most servers and all anarchy, if it doesnt then its the Kill Aura itself that is detected in that server


bbbPriv:
feel free to improve it if needed or adding another setting but its working fine

Wide-Cat:
This isn't something that should be added to KillAura. Meteor already has the `Auto Wasp` module, which follows people using elytra control; you could extend that module to use fireworks, and potentially rename it or update the description/aliases as I have seen a lot of people not know what it does.

bbbPriv:
> This isn't something that should be added to KillAura. Meteor already has the `Auto Wasp` module, which follows people using elytra control; you could extend that module to use fireworks, and potentially rename it or update the description/aliases as I have seen a lot of people not know what it does.

got it

bbbPriv:
@Wide-Cat Ok so i really suggest adding the Elytra Target, because i tested that Auto Wasp module and its detected on every server i used it while the Elytra Target works just fine as long as the kill aura works on that server.
and its not quite the same thing, i think it would be a very good addition to the kill aura, its present on some clients.

bbbPriv:
im willing to tweak it a bit and perfect it if you would add it
and recording a vid of how it works

Wide-Cat:
The Auto Wasp module in its current state will definitely be detected on all servers that have anticheats, as it simply sets your velocity to a constant value in the same way that Elytra Fly does. What I meant was to add a setting to Auto Wasp where it would use fireworks instead - essentially copying the code you already wrote here and placing it in that module instead. I don't think it would be totally appropriate to add this feature to killaura for that reason - we already have a module that does something similar, following a player and keeping as close to them as possible using an elytra, so I think it would be better to put this in that module.

I understand the term 'Wasp' and the corresponding module name Auto Wasp is not really intuitive - it's fairly niche and the average minecraft player probably wouldnt get what it means when reading the module name, so if you add this it would probably be better to rename the module to something like `Elytra Target` or `Elytra Follow`.

Also, can you give an example of what other clients have this feature?

bbbPriv:
> The Auto Wasp module in its current state will definitely be detected on all servers that have anticheats, as it simply sets your velocity to a constant value in the same way that Elytra Fly does. What I meant was to add a setting to Auto Wasp where it would use fireworks instead - essentially copying the code you already wrote here and placing it in that module instead. I don't think it would be totally appropriate to add this feature to killaura for that reason - we already have a module that does something similar, following a player and keeping as close to them as possible using an elytra, so I think it would be better to put this in that module.
> 
> I understand the term 'Wasp' and the corresponding module name Auto Wasp is not really intuitive - it's fairly niche and the average minecraft player probably wouldnt get what it means when reading the module name, so if you add this it would probably be better to rename the module to something like `Elytra Target` or `Elytra Follow`.
> 
> Also, can you give an example of what other clients have this feature?

So you are saying that i should modify Auto Wasp and make it follow the players with fireworks like i did with the elytra target? (if auto fire work enabled)

Also clients that have it (that i know of): ThunderHack Recode, CatLean. 
It is very useful, imagine if he tries to escape in elytra, you just follow him and kill him

Wide-Cat:
> So you are saying that i should modify Auto Wasp and make it follow the players with fireworks like i did with the elytra target? (if auto fire work enabled)

Yes, add a setting to switch between flying via velocity or via fireworks.

bbbPriv:
> > So you are saying that i should modify Auto Wasp and make it follow the players with fireworks like i did with the elytra target? (if auto fire work enabled)
> 
> Yes, add a setting to switch between flying via velocity or via fireworks.

mb will open a new PR for the criticals module update, also okay i will do that

ayaankhan1245:
Hey bbbPriv can i get the elytra target hack ?

PR #6011, opened 2026-01-01
Bunch of QOL Improvements

BerryRock0:
Telekinesis - move items or mobs by changing his speed;
Quantity - dupe/waste items by disable decrement/increment function;
Collisions - change block voxel shape(empty/fullblock variants);
Damages - local godmode for entities or items;
Tick Manipulator - manipulate client&server ticks to change performance;

GhostHand: rewrited internal structure - to emptify voxelshape of chosen blocks;
Hitboxes: added box mode, refactored mixin margin code;


DrParanoya:
To me most of this looks like client-side stuff that will cause massive desyncs without being useful (Telekinesis for example).

Wide-Cat:
Well, I can't say I've ever seen a pr like this before. What @DrParanoya said is true but it's not just that, everything in here is so weird.

elijah-yum:
ts pr is so ass 🥀😭

if you want to vibecode better dm me, and I will introduce you to a skool community

PR #6038, opened 2026-01-09
Implement attack-through-shields setting in KillAura

okfv:
Added a setting to allow attacks through shields.

perfect for stun slamming automatically with a legit config during mace PvP or if you just want it to automatically disable an enemy's shield

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java
@@ -99,6 +99,13 @@ public class KillAura extends Module {
         .build()
     );
 
+    private final Setting<Boolean> attackThroughShields = sgGeneral.add(new BoolSetting.Builder()
+        .name("attack-through-shields")
+        .description("Attacks players even when they are blocking with shields.")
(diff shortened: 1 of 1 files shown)


crosby-moe:
does setting shield mode to 'None' not already do this?

okfv:
it only does that if Auto Swap is enabled and that forces you to swap to a weapon but if you have "All" enabled for weapon mode then it causes an issue, one moment

with Kill Aura's Shield Mode you cannot stun slam, Attribute Swap's Shield Breaker mode is much better and doesn't cause issues while using with Kill Aura @crosby-moe

okfv:
Shield Breaker in Attribute Swap doesn't automatically disable a shield which is good for manually doing so, with a legit Kill Aura config you can combine the two and have the swaps look much more real rather than instantly swapping, disabling, and swapping back

i get it's overall an anarchy client, but this is a small change

okfv:
@crosby-moe thoughts?

Wide-Cat:
ba2ccbaa7004aa9c4d1d1af5cb625270ce3a0c0e

okfv:
@Wide-Cat thank you, it works properly when combined with other things now!

PR #6067, opened 2026-01-19
Add warnings for incompatible active modules in Highway Builder

Leonn170709:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Added checks for incompatible active modules in HighwayBuilder and display clear warning messages when they are enabled.

The module now warns the user if the following modules are active, as they can cause errors or misalignment during building:

- **InstantRebreak** – recommends using `instantly-rebreak-echests` instead  
- **SpeedMine** – recommends using the `fast-break` setting instead  
- **Speed** – warns about diagonal misalignment  
- **Timer** – warns about diagonal misalignment  
- **NoGhostBlocks** – warns about potential block desync errors  

These warnings help prevent common issues and make misconfiguration easier to diagnose.

## Related issues

None

## How Has This Been Tested?

- Tested in IntelliJ
- Tested on 6b6t  

## Checklist

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/HighwayBuilder.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/HighwayBuilder.java
@@ -452,7 +452,7 @@ public class HighwayBuilder extends Module {
     public Vec3d start;
     public int blocksBroken, blocksPlaced;
     private final MBlockPos lastBreakingPos = new MBlockPos();
-    private boolean displayInfo;
+    private boolean displayInfo, warned;
     private boolean suspended = true, inventory = true;
     private int placeTimer, breakTimer, count, syncId;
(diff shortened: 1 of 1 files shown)


Wide-Cat:
Can you explain what the issue with No Ghost Blocks is - is it to do with placing? Because last time I checked the module worked fine when breaking blocks, it just has the downside that it slows highway builder down because you have to wait for the server to send you the block breaking info. If it is to do with placing I would make it check for that setting specifically and print a more detailed warning message.

Leonn170709:
When having it enabled and having a high breaks per tick count it kicks you for sending too many packets because it try’s to break the blocks multiple times until they are broken and it also makes the statistics screen that prints when deactivating the module completely false by as I said it’s breaking the blocks multiple times. But I should rewrite the error to say that too and make it more clear

PR #6076, opened 2026-01-21
EnchantmentOptimizer

Big-Iron-Cheems:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adds a command to calculate the optimal enchant order of an item given a list of enchant-level pairs.
Roughly based on https://github.com/iamcal/enchant-order for the calculation logic.
I think it works, but can be improved for sure :)

## Related issues

None.

# How Has This Been Tested?

Very rough testing, need more eyes on this.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/commands/Commands.java
+++ b/src/main/java/meteordevelopment/meteorclient/commands/Commands.java
@@ -36,7 +36,7 @@ public class Commands {
         add(new DamageCommand());
         add(new DropCommand());
         add(new EnchantCommand());
-        add(new EnchantOptimizeCommand());
+        add(new OptimizeEnchantsCommand());
         add(new FakePlayerCommand());
         add(new FriendsCommand());
         add(new CommandsCommand());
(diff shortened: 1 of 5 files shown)


Big-Iron-Cheems:
The command I am using to test this is:
```
.eopt minecraft:netherite_helmet minecraft:aqua_affinity 1 minecraft:mending 1 minecraft:protection 4 minecraft:respiration 3 minecraft:thorns 3 minecraft:unbreaking 3 minecraft:binding_curse 1 minecraft:vanishing_curse 1
```
At the moment there seems to be a discrepancy between my algorithm and the [website](https://iamcal.github.io/enchant-order/) one.

Work is needed to find out why, relevant sources for the site impl are:

- the enchant data they use: https://github.com/iamcal/enchant-order/blob/main/data.js
- the worker that executes the calculation: https://github.com/iamcal/enchant-order/blob/main/work.js
- the main script (though less important perhaps): https://github.com/iamcal/enchant-order/blob/main/script.js

Big-Iron-Cheems:
At this point I believe it works as intended.
Just need to handle TODOs and figure out how we want to proceed with this.

Big-Iron-Cheems:
I'm not sure if the command should accept an item at all, and instead just take pairs of enchant id and level.
I feel like it would be annoying to implement a calculator that takes into account already enchanted items, and the result would be suboptimal anyways since the previously applied anchants may have been done with a less efficient order.
(Plus it would make a lot of the handling simpler, and I'm not really planning to make this more complicated than it already is).

PR #6172, opened 2026-02-25
Colored info string

Marccccccccccccccc:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Add a new method to @Override to add Color to the InfoString

## Related issues

Idk

# How Has This Been Tested?

SinglePlayer

# Checklist:


src/main/java/meteordevelopment/meteorclient/systems/modules/Module.java
```diff
@@ -149,6 +149,10 @@ public String getInfoString() {
         return null;
     }
 
+    public Color getInfoStringColor() {
```
MukjepScarlet, line 152: Adding `@Nullable` to it could be better

PR #6187, opened 2026-02-28
Respect sign translation fallbacks for Meteor key spoofing

MistressOfDNS:
## Type of change

- [x ] Bug fix
- [ ] New feature 

AbstractSignEditScreenMixin was turning Meteor translation keys on signs into the raw key string every time. Which broke vanilla behavior for signs that include a fallback. Vanilla returns the fallback when the key can’t be resolved.

This change keeps the existing key protection. But if the sign text is translatable and includes a fallback, it uses that fallback instead. If there isn’t one, it still uses the raw key. This makes Meteor behave like vanilla for fallback-based sign checks and fixes the detection issue.

# How Has This Been Tested?

Gone on a server with a detection and not detected

# Checklist:

- [ x] My code follows the style guidelines of this project.
- [ Not Needed] I have added comments to my code in more complex areas.
- [ x] I have tested the code in both development and production environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ logs
 .idea/*
 !.idea/copyright/*
 !.idea/scopes/*
+.vscode/*
+.idea/*
--- a/src/main/java/meteordevelopment/meteorclient/mixin/AbstractSignEditScreenMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/AbstractSignEditScreenMixin.java
@@ -34,7 +34,10 @@ public abstract class AbstractSignEditScreenMixin {
         if (message.getContent() instanceof TranslatableTextContent content) {
             String key = content.getKey();
 
-            if (key.contains("meteor-client")) modified = MutableText.of(new PlainTextContent.Literal(key));
+            if (key.contains("meteor-client")) {
+                String fallback = content.getFallback();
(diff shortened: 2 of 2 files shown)


crosby-moe:
this hotfix is still flawed as it does not take server resource packs into account

in my opinion it is better to outright remove it at this time with #6186 to let dedicated mods such as exploit preventer handle it

MistressOfDNS:
sure then pull the changes of yours please

PedroMPagani:
Honestly you should simply allow servers to be able to block this. If they are blocking it's because they don't want anyone to use it. :)

NikOverflow:
This is not only flawed because server resource packs but also for the with thing with variable elements. The best way is removing the faulty protection and use ExploitPreventer.

If people really care about it now: Use my meteor-client fork. There is a release that removes that already and then you need to install a mod that protects against that properly.

crosby-moe:
> Honestly you should simply allow servers to be able to block this. If they are blocking it's because they don't want anyone to use it. :)

hi pedro yea that's the plan mostly we dont intend to push a full fix

though it is encouraged for donutsmp to start detecting meteor more heuristically instead of relying on client exploits as we will eventually (within the coming months) be switching to our own translation format separate from minecraft's

Issue #6056, opened 2026-01-14
Nametags prefixes and suffixes


harmfulladvocado:
### Describe the feature

adding a button to see the prefix & suffix of a user in the nametags module
could be very nice to not need to look in tab to see what rank a user had :)

### Before submitting a suggestion

- [x] This feature doesn't already exist in the client. (I have checked every module and their settings on the **latest dev build**)

- [x] This wasn't already suggested. (I have searched suggestions on GitHub and read the FAQ)

PR #6203, opened 2026-03-06
Show Prefix and Suffix in Nametags

Minchen202:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Adding an option to the module nametags, that allows users to toggle if prefix's and suffix's are supposed to be displayed in there nametags. This is for easy checking which rank they have.

## Related issues

#6056

# How Has This Been Tested?

https://github.com/user-attachments/assets/3f1342f4-6cf1-4505-8979-293cd4cf9526

# Checklist:


MineGame159:
It would probably be better to name the setting more closely to what it does. Since the display name can be much more than just prefix / suffix. `custom-name`, `display-name`, `server-name`, ..., not sure what the best would be.

Issue #6280, opened 2026-04-03
XRAY 1.21.11 ORES SHOWS ONLY EXPOSURE


Sammyjmz12:
### Describe the bug

The Xray module is not working as intended on version 1.21.11. It should render all selected ores through blocks, but instead, it only shows ores that are exposed to air or caves. Fully hidden ores remain invisible.

### Steps to reproduce

1.- Launch the game with Meteor Client and join a world.

2.- Enable the Xray module and ensure ores are selected in the settings.

3.- Look underground.

4.- Observe that only exposed ores are highlighted, while hidden ores are not rendered at all.

### Meteor Version

meteor-client-1.21.11-75

### Minecraft Version

1.21.11

### Operating System

Windows

### Before submitting a bug report

- [x] This bug wasn't already reported (I have searched bug reports on GitHub).

- [x] This is a valid bug (I am able to reproduce this on the latest dev build).

1brane:
hey first make sure you dont have "exposed ores only" checked

Sammyjmz12:
> hey first make sure you dont have "exposed ores only" checked

I already checked that, it was the first thing I tried. I'm pretty sure it's a bug.

choneas:
I am having the same problem in 1.21.1. It worked again after disabling Sodium.

Sammyjmz12:
Yes, your solution worked for me too! But I find it really weird because Xray used to work perfectly fine alongside Sodium in previous versions.

qwertyad1:
its most likely some block culling thing is sodium

PR #6287, opened 2026-04-05
fix xray only shows exposed ores

ghost:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

SodiumBlockOcclusionCacheMixin was applying a sodium class called BlockOcclusionCache which was removed and replaced with AbstractBlockRenderContext causing the mixin to not inject correctly

## Related issues

#6280

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -13,7 +13,7 @@ loom = "1.14-SNAPSHOT"
 # Baritone (https://github.com/MeteorDevelopment/baritone)
 baritone = "1.21.10-SNAPSHOT"
 # Sodium (https://github.com/CaffeineMC/sodium-fabric)
-sodium = "mc1.21.11-0.8.0-fabric"
+sodium = "mc1.21.11-0.8.7-fabric"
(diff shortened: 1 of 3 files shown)


noramibu (approved):

Big-Iron-Cheems:
For this PR to be valid it should:
1. Update the Sodium version we depend on: https://github.com/MeteorDevelopment/meteor-client/blob/master/gradle/libs.versions.toml#L16 (The target class still exists up until `0.8.6`)
2. Update the breaking version os Sodium here: https://github.com/MeteorDevelopment/meteor-client/blob/master/src/main/resources/fabric.mod.json#L60 (Require at least `0.8.7` as older versions would now be invalid targets)
3. Properly import the shadowed fields like:
```java
@Shadow 
protected BlockState state;
```

This would also mean updating any other Sodium mixins that may have broken in the [`0.8.0` to `0.8.7`](https://github.com/CaffeineMC/sodium/compare/mc1.21.11-0.8.0...mc1.21.11-0.8.7) range.

Big-Iron-Cheems (approved):

PR #6330, opened 2026-04-14
"Reach" fix

BerryRock0:
Inability to reach the entity even with the settings turned up to maximum has been fixed


mariethepolarfox:
This is a pretty low effort workaround, maybe try looking into AttackRangeComponent.getEffectiveMaxRange and modifying its return value to reflect the reach setting

PR #6334, opened 2026-04-15
Logout spot imrpovemet

Powie69:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

- Don't remove logout spot when module is deactivated. Instead there's a button for it like in storage esp's "clear rendering cache"
- logout command
  - `remove <name>` removes the player from the logout list
  - `clear` clears the logout list
  
## Related issues

Might break on your machine crosby told me that the ordering of command registry is very complex & hard to understand. But it works on my machine though 👍
image

# How Has This Been Tested?

https://github.com/user-attachments/assets/2158b477-f221-48ad-bed0-543234260e98

# Checklist:


cqb13:
would be nice if there was an option to controls if it clears on deactivate or not.

src/main/java/meteordevelopment/meteorclient/commands/commands/LogoutCommand.java
```diff
@@ -0,0 +1,52 @@
...
+            .executes(_ -> {
+                logoutSpots.clearLogoutSpots();
+                
+                ChatUtils.info("Cleared all logout spots");
+                return SINGLE_SUCCESS;
+            })
+        );
+
+        builder.then(literal("remove")
+            .then(argument("name", StringArgumentType.word())
```
crosby-moe: ideally this should have suggestions implemented

Frko5000:
With placeholder of player names it would be good

Frko5000:
> would be nice if there was an option to controls if it clears on deactivate or not.

Wdym

PR #6346, opened 2026-04-23
Add support HTTP-proxies and encryption with SSL

zuenkoma:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Implemented full support for HTTP/HTTPS proxies and SSL-encrypted connection to proxy servers (HTTPS, SOCKS4S, SOCKS5S). Previously the application worked only with plain SOCKS4/SOCKS5.

# How Has This Been Tested?

- Manually tested with public HTTP, HTTPS and SOCKS5S proxies.
- Verified that proxy check correctly identifies alive proxies of each type.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
+++ b/src/main/java/meteordevelopment/meteorclient/gui/screens/ProxiesImportScreen.java
@@ -49,52 +49,54 @@ public class ProxiesImportScreen extends WindowScreen {
 
                     matcher = Proxies.PROXY_PATTERN.matcher(line);
                     if (matcher.matches()) {
-                        String address = matcher.group(2).replaceAll("\\b0+\\B", "");
-                        int port = Integer.parseInt(matcher.group(3));
+                        String address = matcher.group("address").replaceAll("\\b0+\\B", "");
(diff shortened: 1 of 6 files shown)


Big-Iron-Cheems:
I have some questions regarding the PR if you could address them @zuenkoma.

- In `ConnectionMixin` you do:
```java
pipeline.addFirst(new HttpProxyHandler(...));
pipeline.addFirst(new SslHandler(engine));
```
So the pipeline becomes `SslHandler` -> `HttpProxyHandler` -> `...` 
Is this order correct or should the SSL step happen after the HTTP one?
- `InsecureTrustManagerFactory.INSTANCE` disables certificate validation, is this necessary or can we use the default?
- Current check only verifies response starts with `HTTP/`, meaning failures like 403, 500, or malformed proxy responses still count as alive. Perhaps we should parse the HTTP status line and require a `200` for succesful CONNECT support?
- CONNECT target uses `0.0.0.0:80`, maybe using a real reachable address like `1.1.1.1:80` could be better?
- Perhaps `tls`, `ssl`, or `encryptedTransport` would be a better name than `secure`.

Thanks for the contribution nonetheless.

PR #6348, opened 2026-04-24
Add translation blocking to prevent server-side mod detection

YeaiDoNotKnowAGoodUsername:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Servers like Donut SMP check for translation keys on signs to detect mods. So I added a mixin that strips the `translate` tag from sign NBT packets when they come from the server. It replaces them with the `fallback` text instead. This way translation keys like `key.meteor-client.open-gui` don't get resolved, so servers can't detect which mods you have.

It works on all translation keys on signs (front and back), and it's always-on so you don't need to toggle it.

## How It Works

The mixin intercepts `BlockEntityUpdateS2CPacket` packets when they're received from the server. It looks at the NBT data in the packet and checks both `front_text` and `back_text` for `messages` arrays. For each message in the array, if it contains a `translate` tag, the mixin removes that tag and replaces it with a `text` tag set to the `fallback` value. This prevents the client from trying to resolve the translation key, so the server can't detect if the key exists on your client.

The fallback text is the plain text version that servers send as a backup in case translation fails. By using this instead of the translation key, the sign still displays readable text but without exposing which mods or resource packs you have installed.

## What It Works On

- All translation keys on signs (front and back text)
- All mods and resource packs that use translation keys
- Any server that uses translation key checking for mod detection
- Always-on, So you don't have to look or find it So you don't have to deal with accidental getting banned

Since it works at the NBT packet level, it blocks translation keys from anything - mods, resource packs, or any other source that uses translate tags on signs.

## Related issues

None

# How Has This Been Tested?

Built it and went on a server with translation key checks - didn't get detected.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/ClientPlayNetworkHandlerMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/ClientPlayNetworkHandlerMixin.java
@@ -38,6 +38,9 @@ import net.minecraft.client.world.ClientWorld;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.ItemEntity;
 import net.minecraft.network.ClientConnection;
+import net.minecraft.nbt.NbtCompound;
+import net.minecraft.nbt.NbtList;
+import net.minecraft.network.packet.s2c.common.ResourcePackSendS2CPacket;
 import net.minecraft.network.packet.s2c.play.*;
 import net.minecraft.util.math.Vec3d;
 import net.minecraft.world.chunk.WorldChunk;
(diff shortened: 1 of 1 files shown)


CypherNebula84:
You should add a toggle but its on by default.

mariethepolarfox:
If servers in ANY way check if the normal mc translation keys get translated this will just get detected as well

YeaiDoNotKnowAGoodUsername:
> If servers in ANY way check if the normal mc translation keys get translated this will just get detected as well

well that's for me to fix when they counter with that but for now this should work sense that is a good thing to know i didn't think of but i don't want to make a specific index of translation keys that are banned because that would take lots of time to find all and things change so i will try to work on a future version for if they check it but it should work until servers decide to check if translations work on normal mc

crosby-moe:
1. a partial fix serves nothing if not making it easy to detect meteor client's patch specifically (see [previous "fix"](https://github.com/MeteorDevelopment/meteor-client/commit/ff124ecdedb1d15ac71be94b06f98dba763bb78a) & the donutsmp fiasco)
2. this is not the direction we want to take (see [translations branch](https://github.com/MeteorDevelopment/meteor-client/tree/translations))

Issue #6361, opened 2026-04-28
GUI cannot be opened on the multiplayer screen


gracieancelotti:
### Describe the bug

The GUI can no longer be opened while on the multiplayer screen. Pressing the keybind has no effect on that screen.

### Steps to reproduce

1. Launch Minecraft with Meteor Client
2. Navigate to the Multiplayer screen
3. Press the GUI keybind (by default Right Shift)
4. GUI does not open

### Meteor Version

2

### Minecraft Version

26.1.2

### Operating System

Windows

### Before submitting a bug report

- [x] This bug wasn't already reported (I have searched bug reports on GitHub).

- [x] This is a valid bug (I am able to reproduce this on the latest dev build).

Big-Iron-Cheems:
This is not a bug.
Because of some Minecraft `26.1` changes, we are unable to open a custom screen that could contain block icons anywhere while outside of a world.
Doing so would cause a crash as now all `ItemStack` instances need to load data components provided by a loaded world.

We could try and provide a "restricted" version of the GUI that can be opened everywhere, but it will need to not have any `ItemStack`-based icons, and probably some other stuff that could trigger this.

gracieancelotti:
> This is not a bug. Because of some Minecraft `26.1` changes, we are unable to open a custom screen that could contain block icons anywhere while outside of a world. Doing so would cause a crash as now all `ItemStack` instances need to load data components provided by a loaded world.
> 
> We could try and provide a "restricted" version of the GUI that can be opened everywhere, but it will need to not have any `ItemStack`-based icons, and probably some other stuff that could trigger this.

Thank you for the explanation. I hope a restricted version would be possible at some point, since it's a bit annoying having to join a singleplayer world just to manage my modules before joining a server

Tetogooner67:
hey i found a solution if u use the modmenu mod and click the 3 lines on meteor it can be opened in the mods area while not on a server 

okay i tested it i can open the right click and bind stuff to!

<img width="953" height="549" alt="Image" src="https://github.com/user-attachments/assets/48953fce-409d-46d8-9978-68d61ce66707" />

<img width="1919" height="1031" alt="Image" src="https://github.com/user-attachments/assets/0ec61a64-ed3b-4b27-92f7-a38baadc96e3" />

Big-Iron-Cheems:
We should probably have restricted that too just to be safe.
From what I know if you try to open the settings of a module that would have block icons it would crash the game so be carful.

PR #6386, opened 2026-05-07
Fix opening GUI outside of worlds

machiecodes:
## Type of change

- [x] Bug fix
- [x] New feature

## Description

In 26.1 Itemstacks now require components to be loaded from registry entries that are only provided when we join a world, however since we only need a few specific components that are stored client-side we can simply construct our own stacks for rendering that contain only those.

Implemented utils/DisplayItemUtils which creates these stacks and replaced all getDefaultInstance() calls in rendering contexts with DisplayItemUtils#toStack(Item item)

**Breaking for addons**, any addon that defines category icons will need to do so using the new util as well or NPE

Also added a few placeholder items in inventoryhud that show while outside of a world

## Related issues

#6361 

# How Has This Been Tested?

https://github.com/user-attachments/assets/9da1d635-6cf0-4abe-8bce-a44c4d95d1e8

# Checklist:


machiecodes:
Actually yea that is related I am simply blind :3 thanks

Big-Iron-Cheems (approved):

elijah-yum (approved):

PR #6406, opened 2026-05-16
Nametag module and item list setting screen imporvements

cqb13:
- Added ability to control which items nametags are rendered on instead of just all or no items
- Made item list setting have categories (Blocks, Tools, Weapons, Food, etc...) sorta like how the entity one has categories

## Type of change

- [ ] Bug fix
- [X] New feature

# How Has This Been Tested?

Screenshot 2026-05-15 at 10 34 53 PM
Screenshot 2026-05-15 at 10 35 17 PM

*If a list setting is opened in the main menu before a world is joined, the categories will not be computed correctly. I'm pretty sure this is because the data components have not been loaded yet or something like that. After the player has joined a world, the categories will be fine even in the main menu. (In current meteor, nothing is shown, as seen below, so I think this is a step up, not sure if there is a good solution to this)*

image

# Checklist:


elijah-yum (approved):

Frko5000:
This would be good

PR #6417, opened 2026-05-21
Better Farming module (crop farming improvements)

moxvallix:
## Type of change

- [x] New feature

## Description

This adds a new module, named Better Farming, which adds couple of utilities related to crop farming.

Firstly, the module will attempt to replace a harvested crop automatically, granted the player has its seed in their inventory.

Secondly, using the same method as NoFall, but changing it to work only when above farmland, it suppresses crop trampling, allowing the player to jump on the crops without them trampling.

Then, it will also prevent the player from harvesting crops that are not yet ripe.

Lastly, for sugar cane and bamboo, it will prevent the player from breaking the bottom block.

All of these should help players maintain sizable manual crop farms. All listed behaviours can be toggled in the settings.

image

# How Has This Been Tested?
https://youtu.be/VH0yt58lyZs

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
@@ -1,0 +1,155 @@
+/*
+ * This file is part of the Meteor Client distribution (https://github.com/MeteorDevelopment/meteor-client).
+ * Copyright (c) Meteor Development.
+ */
+
+package meteordevelopment.meteorclient.systems.modules.player;
+
+import meteordevelopment.meteorclient.MeteorClient;
+import meteordevelopment.meteorclient.events.entity.player.StartBreakingBlockEvent;
+import meteordevelopment.meteorclient.events.packets.PacketEvent;
+import meteordevelopment.meteorclient.events.world.TickEvent;
+import meteordevelopment.meteorclient.mixin.ServerboundMovePlayerPacketAccessor;
+import meteordevelopment.meteorclient.mixininterface.IServerboundMovePlayerPacket;
+import meteordevelopment.meteorclient.settings.BoolSetting;
+import meteordevelopment.meteorclient.settings.Setting;
+import meteordevelopment.meteorclient.settings.SettingGroup;
+import meteordevelopment.meteorclient.systems.modules.Categories;
+import meteordevelopment.meteorclient.systems.modules.Module;
+import meteordevelopment.meteorclient.utils.player.FindItemResult;
+import meteordevelopment.meteorclient.utils.player.InvUtils;
+import meteordevelopment.meteorclient.utils.world.BlockUtils;
+import meteordevelopment.orbit.EventHandler;
+import meteordevelopment.orbit.EventPriority;
+import net.minecraft.core.BlockPos;
+import net.minecraft.network.protocol.game.ServerboundMovePlayerPacket;
+import net.minecraft.tags.BlockTags;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.level.block.Blocks;
+import net.minecraft.world.level.block.state.BlockState;
+
+import java.util.ArrayList;
+
+public class BetterFarming extends Module {
+    private final SettingGroup sgGeneral = settings.getDefaultGroup();
+
+    // General
+
+    private final Setting<Boolean> cropReplace = sgGeneral.add(new BoolSetting.Builder()
+        .name("crop-replace")
+        .description("Automatically plant new crop on harvest.")
+        .defaultValue(true)
+        .build()
+    );
+
+    private final Setting<Boolean> suppressTrample = sgGeneral.add(new BoolSetting.Builder()
+        .name("suppress-trample")
+        .description("Attempts to prevent player from trampling crops.")
+        .defaultValue(true)
+        .build()
+    );
+
+    public BetterFarming() {
+        super(Categories.Player, "better-farming", "Improvements for crop farming.");
+    }
+
+    private Item placeItem = null;
+    private ArrayList<BlockPos> placements = new ArrayList<>();
+    private int mineCooldown = 0;
+
+    @EventHandler(priority = EventPriority.HIGH)
+    private void onTick(TickEvent.Pre event) {
+        if (cropReplace.get()) autoPlaceTick();
+    }
+
+    @EventHandler(priority = EventPriority.HIGH)
+    private void onStartBreakingBlockEvent(StartBreakingBlockEvent event) {
+        if (cropReplace.get()) autoPlaceBreakEvent(event);
+    }
+
+    @EventHandler
+    private void onSendPacket(PacketEvent.Send event) {
+        if (suppressTrample.get()) trampleSuppressionSendEvent(event);
+    }
+
+    private void autoPlaceTick() {
+        if (mineCooldown > 0) {
+            mineCooldown--;
+        }
+
+        if (placeItem == null) return;
+        if (placements.isEmpty()) {
+            placeItem = null;
+            return;
+        }
+
+        BlockPos blockPos = placements.getFirst();
+        placements.removeFirst();
+
+        FindItemResult foundItem = InvUtils.find(placeItem);
+
+        // If the found item is not in mainhand yet, its not ready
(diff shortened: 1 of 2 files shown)


crosby-moe:
no break for the bottom block of cactus & the log where cocoa beans is attached could be good too

moxvallix:
> no break for the bottom block of cactus & the log where cocoa beans is attached could be good too

Oh true, good idea. Maybe budding amethyst too (although that can just be done in No Interact)

moxvallix:
I added a new setting: no break supporting blocks. When this is on, blocks that are holding a crop cannot be broken, which also extends to cocoa.

Also cactus is now counted as a cane block.

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
```diff
@@ -0,0 +1,289 @@
...
+            || isCaneBlock(blockState);
+    }
+
+    private boolean isReplaceableCrop(BlockState blockState) {
+        return blockState.is(BlockTags.CROPS)
+            || blockState.is(Blocks.NETHER_WART);
+    }
+
+    private boolean checkForCocoa(BlockPos blockPos) {
+        Direction[] checkDirections = {
```
MukjepScarlet, line 138: Can be replaced with `Direction.BY_2D_DATA`

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
```diff
@@ -0,0 +1,289 @@
...
+
+        if (!foundItem.found()) return;
+
+        placeItem = blockItem;
+        cropPlacements.add(event.blockPos);
+
+        if (foundItem.isMainHand()) {
+            blockBreakCooldown = 3;
+            return;
+        };
```
MukjepScarlet: `;`

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
```diff
@@ -0,0 +1,289 @@
...
+
+        mc.gameMode.handlePickItemFromBlock(event.blockPos, false);
+
+        // Cancel the event to allow pickblock to do its thing
+        event.cancel();
+    }
+
+    private void trampleSuppressionSendEvent(PacketEvent.Send event) {
+        if (mc.player == null) return;
+        if (!(event.packet instanceof ServerboundMovePlayerPacket)
```
MukjepScarlet, line 256: `instance of IServerboundMovePlayerPacket p`

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
```diff
@@ -0,0 +1,289 @@
...
+            || ((IServerboundMovePlayerPacket) event.packet).meteor$getTag() == 1337) return;
+
+        ServerboundMovePlayerPacket packet = (ServerboundMovePlayerPacket) event.packet;
+
+        BlockPos blockPos = new BlockPos(
```
MukjepScarlet: `BlockPos.containing`?

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
```diff
@@ -0,0 +1,289 @@
...
+
+        if (!bsIsCane || bsBelowIsCane) return;
+
+        event.cancel();
+    }
+
+    private void noBreakSupportingBlocksBreakEvent(StartBreakingBlockEvent event) {
+        BlockPos blockPos = event.blockPos;
+        BlockState blockState = mc.level.getBlockState(blockPos);
+        BlockState bsAbove = mc.level.getBlockState(blockPos.offset(0, 1, 0));
```
MukjepScarlet: .up()

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
```diff
@@ -0,0 +1,289 @@
...
+        Direction[] checkDirections = {
+            Direction.NORTH,
+            Direction.SOUTH,
+            Direction.EAST,
+            Direction.WEST,
+        };
+
+        for (Direction direction : checkDirections) {
+            // Check block at blockPos offset by the direction normal vector.
+            BlockState bsCheck = mc.level.getBlockState(blockPos.offset(direction.getUnitVec3i()));
```
MukjepScarlet: `blockPos.relative(direction)`

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
```diff
@@ -0,0 +1,289 @@
...
+        .defaultValue(true)
+        .build()
+    );
+
+    public BetterFarming() {
+        super(Categories.Player, "better-farming", "Improvements for crop farming.");
+    }
+
+    private Item placeItem = null;
+    private final ArrayList<BlockPos> cropPlacements = new ArrayList<>();
```
MukjepScarlet: ArrayDeque/ArrayListDeque will be better because you call `removeFirst` on it

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
moxvallix: That seems to have private access?

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
moxvallix: I don't really understand the recommended change here? Also the way its written there is copied almost directly from NoFall, so we may want to change it there as well if there's a better way.

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
moxvallix: `.up()` doesn't exist, but I'm guessing you meant `.above()`.

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
moxvallix: ArrayList also seems to have `removeFirst`. I assume ArrayListDeque is better for performance?

MukjepScarlet: Yes. ArrayList.removeFirst moves all elements after first one

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
MukjepScarlet: Oh yeah `up` is yarn name

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
MukjepScarlet: Well that's true

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
moxvallix: ah ok makes sense

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
moxvallix: ok cool yeah i will use that then, thanks

src/main/java/meteordevelopment/meteorclient/systems/modules/player/BetterFarming.java
moxvallix: going to resolve this, not sure what to do here otherwise

elijah-yum:
`suppress crop trampling` doesn't work when falling from a great height alongside with great speed

moxvallix:
@elijah-yum yeah unfortunately i dont think much can be done about that. it works fine rn for most use cases.

PR #6435, opened 2026-05-25
fix unclosed IO resources, unify notebot folder creation

MukjepScarlet:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

A summary of the changes along with the reasoning behind the changes.

## Related issues

Nothing I guess

# How Has This Been Tested?

Local env (development)

# Checklist:


src/main/java/meteordevelopment/meteorclient/commands/arguments/NotebotSongArgumentType.java
```diff
@@ -34,12 +34,12 @@ private NotebotSongArgumentType() {
     public Path parse(StringReader reader) throws CommandSyntaxException {
         final String text = reader.getRemaining();
         reader.setCursor(reader.getTotalLength());
-        return MeteorClient.FOLDER.toPath().resolve("notebot/" + text);
+        return Notebot.PATH.toPath().resolve(text);
```
crosby-moe, line 37: would be a good idea to throw a `CommandSyntaxException` on path traversal here

MukjepScarlet: e.g. `text` contains `..`?

crosby-moe: > e.g. `text` contains `..`?

check if the absolute path of `Notebot.PATH.toPath().resolve(text)` startswith `Notebot.PATH.toPath()`

MukjepScarlet: Like this?
```java
Path resolveSafely(Path p1, String p2) {
var r = p1.resolve(p2).normalize()
if (!r.startsWith(p1)) throw new IllegalArgument
return r
}
```

MukjepScarlet: Okay main branch seems to include this logic

PR #6443, opened 2026-05-30
Fix type mismatch in AutoArmor, aura leak in AutoEat, stale packets in Blink

zur1i:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

**AutoArmor — Curse of Binding check never fires + avoided enchantment penalty broken**

`enchantments` is keyed by `Holder<Enchantment>`, but two places were passing `ResourceKey<Enchantment>` directly. `containsKey(Enchantments.BINDING_CURSE)` always returns false since the types never compare equal, meaning AutoArmor would attempt to swap off Binding Curse armor. Same issue in `decreaseScoreByAvoidedEnchantments` — `enchantments.getInt(ResourceKey)` always returns 0, so the score penalty for avoided enchantments was effectively dead code. Fixed both by using `Holder.is()` and `Utils.getEnchantmentLevel`, consistent with how the rest of the class handles enchantment lookups.

**AutoEat — auras get permanently disabled when slot swap fails**

When `searchInventory` is enabled and food is in the main inventory, `changeSlot` can return false if the hotbar is full. In that case `eat()` returns early without setting `eating = true`, but `startEating` still continues and pauses auras and baritone. Since `eating` stays false, `stopEating` is never triggered — including from `onDeactivate`, which guards on `if (eating)`. Added an early return in `startEating` right after `eat()` to avoid pausing auras if the eat didn't actually start.

**Blink — stale packets sent to the next server after disconnect**

`onLeaveGame` called `onDeactivate`, which exits early through `Utils.canUpdate()` when `mc.player` is null. So the `packets` list never gets cleared on disconnect. When the player joins a new server, those old movement packets are still queued and get flushed when Blink is disabled. Replaced the `onDeactivate` call with `dumpPackets(false)` directly since it skips the send path and is safe to call with a null player.

## Related issues

N/A

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 
-<p align="center">1
+<p align="center">
 <img src="https://meteorclient.com/icon.png" alt="meteor-client-logo" width="15%"/>
 </p>
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 
-<p align="center">
+<p align="center">1
 <img src="https://meteorclient.com/icon.png" alt="meteor-client-logo" width="15%"/>
 </p>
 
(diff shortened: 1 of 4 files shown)


ovklol:
Heads up — I accidentally pushed directly to my fork’s master instead of a separate branch. The changes are the same, will use a dedicated branch next time.

PR #6455, opened 2026-06-07
Fix PlayerRadarHud null crash when switching servers

### LINKED ISSUE #6433: [Bug] PlayerRadarHud causes crash
### Describe the bug

PlayerRadarHud causes crash when switching servers in a hub system.

`java.lang.NullPointerException: Cannot invoke "net.minecraft.world.entity.Entity.position()" because "entity" is null
at meteordevelopment.meteorclient.systems.hud.elements.PlayerRadarHud.lambda$getPlayers$0(PlayerRadarHud.java:207)
`

When the server sends a StartConfigurationPacket (which is what triggers a server switch on a hub), Minecraft starts clearing the current level/world. At that exact moment, Meteor's Player Radar HUD element is still rendering and tries to sort nearby players by distance. One of those player entities is already null (being cleaned up), so it crashes trying to call .position() on it.

The second error (this.level is null) is just a cascade from the first crash — the world is already gone by the time the renderer tries to continue.

### Steps to reproduce

It happens every time I try to switch to a different server on a hub network.
I have only tried it on one specific server though.

### Meteor Version

Meteor Client 26.1.2 build #22

### Minecraft Version

26.1.2

### Operating System

Windows

Frko5000:
## Type of change
- [x] Bug fix
- [ ] New feature

## Description
PlayerRadarHud was crashing when switching servers on a hub network. During a server switch, Minecraft starts clearing the current world and entities become null. The HUD was still rendering and trying to call .position() on null entities, causing a NullPointerException. Fixed by adding a null check to filter out null entities and a null camera entity before sorting.

## Related issues
Fixes #6433

## How Has This Been Tested?
Tested by switching servers on a hub network with PlayerRadarHud enabled. The crash no longer occurs.

## Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/hud/elements/PlayerRadarHud.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/hud/elements/PlayerRadarHud.java
@@ -203,6 +203,8 @@ public class PlayerRadarHud extends HudElement {
     private List<AbstractClientPlayer> getPlayers() {
         players.clear();
         players.addAll(mc.level.players());
+        // Filter out null entities and skip if camera entity is null to prevent crashes during server switches
+        players.removeIf(e -> e == null || mc.getCameraEntity() == null);
         if (players.size() > limit.get()) players.subList(limit.get() - 1, players.size() - 1).clear();
         players.sort(Comparator.comparingDouble(e -> e.distanceToSqr(mc.getCameraEntity())));
 


nicdv1337:
W

Frko5000:
@Big-Iron-Cheems Sir i don't see any reason for not merging this :3

src/main/java/meteordevelopment/meteorclient/systems/hud/elements/PlayerRadarHud.java
```diff
@@ -203,6 +203,8 @@ public void render(HudRenderer renderer) {
     private List<AbstractClientPlayer> getPlayers() {
         players.clear();
         players.addAll(mc.level.players());
+        // Filter out null entities and skip if camera entity is null to prevent crashes during server switches
+        players.removeIf(e -> e == null || mc.getCameraEntity() == null);
```
MukjepScarlet, line 207: `mc.getCameraEntity() == null` in `removeIf` is so weird

Frko5000:
hmm thats right

Frko5000:
um updated the fix  moved the `getCameraEntity()` null check to the top so it returns early instead of checking it in removeIf does this look better?

```java
private List<AbstractClientPlayer> getPlayers() {
    players.clear();
    // Prevent NPE when camera entity is null during server switches or world unloading
    if (mc.getCameraEntity() == null) return players;
    players.addAll(mc.level.players());
    if (players.size() > limit.get()) players.subList(limit.get() - 1, players.size() - 1).clear();
    players.sort(Comparator.comparingDouble(e -> e.distanceToSqr(mc.getCameraEntity())));
    return players;
}
```

Frko5000:
ill be closing this PR

PR #6472, opened 2026-06-14
fix(KillAura): use isAggressive() instead of isAngry() for Enderman

### LINKED ISSUE #6463: Enderman is not filtered by the "Ignore Passive" option in Kill Aura
### Describe the bug

Angry Endermen are not detected by the "Ignore Passive" option in Kill Aura, which doesn't happen with other mobs like Piglins.

I don't think it's necessary to share any media about this, since the issue is very easy to notice and reproduce with an angry Enderman. This problem has existed since the first build of 26.1.x.

### Steps to reproduce

1. Turn on Kill Aura
2. Make sure the "Ignore Passive" option is enabled
3. Spawn an Enderman or find one in the world
4. Make it angry
5. Notice that Kill Aura does not detect it

### Meteor Version

Meteor 26.1.2-22

### Minecraft Version

26.1.2

### Operating System

Linux

Frko5000:
## Type of change

- [x] Bug fix

## Description

`isAngry()` is server-side only and always returns false on the client, causing angry Endermen to be incorrectly ignored by the ignore-passive option. Changed to `isAggressive()` which is properly synced to the client.

## Related issues

Fixes #6463

# How Has This Been Tested?

Tested in-game by angering an Enderman with ignore-passive enabled. KillAura now correctly targets the angry Enderman.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/combat/KillAura.java
@@ -418,7 +418,7 @@ public class KillAura extends Module {
             ) return false;
         }
         if (ignorePassive.get()) {
-            if (entity instanceof EnderMan enderman && !enderman.isAngry()) return false;
+            if (entity instanceof EnderMan enderman && !enderman.isAggressive()) return false;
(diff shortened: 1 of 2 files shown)


Wide-Cat:
Funny issue due to a mapping collision

PR #6485, opened 2026-06-20
Fix custom font fallback rendering

gta66666668:
Fixes custom text rendering for glyphs that are not present in the selected Meteor font, such as Chinese characters.

The previous custom font renderer pre-packed a fixed set of Latin, Greek, Cyrillic, and symbol ranges. Characters outside those ranges could not render even when the OS had a suitable font installed.

Changes:
- load an available system fallback font for common CJK font families
- let Font use multiple font buffers
- lazily pack missing glyphs into the texture atlas
- upload pending glyphs before GUI/HUD custom text rendering
- handle Unicode code points instead of only Java char values

Validation:
- ./gradlew.bat build --stacktrace passes

This PR is based on master and does not include the Minecraft 26.2 update branch.

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/renderer/Fonts.java
+++ b/src/main/java/meteordevelopment/meteorclient/renderer/Fonts.java
@@ -17,6 +17,8 @@ import meteordevelopment.meteorclient.utils.PreInit;
 import meteordevelopment.meteorclient.utils.render.FontUtils;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/meteordevelopment/meteorclient/renderer/Fonts.java
```diff
@@ -90,4 +109,57 @@ public static FontFamily getFamily(String name) {
 
         return null;
     }
+
+    public static List<FontFace> getFallbackFonts(FontFace primary) {
+        List<FontFace> fonts = new ArrayList<>();
```
MukjepScarlet: Why you even need a list here? Isn't this just an `Optional<FontFace>`

gta66666668: Changed. You are right: this path selects at most one fallback face, so returning a `List<FontFace>` made the API more general than the implementation actually was. `getFallbackFont(FontFace)` now returns `Optional<FontFace>`, and `readFontBuffers` consumes it directly with `flatMap` / `ifPresent`.

I also tightened the related rendering path while rebasing this PR onto the latest `master`:

- fallback selection skips the entire primary family, prefers the primary style, then Regular, and uses locale-stable name matching;
- primary/fallback font buffers are loaded once and reused by GUI/HUD font sizes instead of rereading large CJK font files for every height;
- width measurement now caches glyph metrics without rasterizing into (and consuming) the texture atlas;
- missing prepacked glyphs correctly fall through to the fallback font, while supplementary Unicode code points remain handled as code points rather than UTF-16 chars;
- partially filled initial atlases are accepted safely, missing/unpackable glyphs are cached, and one packing failure no longer disables all later glyphs;
- renderer replacement and font cleanup are now atomic/idempotent, with HUD font resets deferred until the current frame finishes.

Validated with `./gradlew.bat build --stacktrace --console=plain` (`BUILD SUCCESSFUL`).

PR #6486, opened 2026-06-20
Fix TheAltening authentication

8KV:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

Fixes TheAltening account authentication by replacing the old `WaybackAuthLib` login path with a direct Yggdrasil-compatible `/authenticate` request.

## Related issues

None

# How Has This Been Tested?

Tested in a production client environment with TheAltening login. Authentication succeeds; Mojang profile key fetch may still return `401`.

# Checklist:


1532658585:
Great fix, looks like that is all it needed to be functional again with Meteor. 
Just for clarity, the password doesn't have to be liquidbounce or meteor on crack, any value is accepted.

PR #6500, opened 2026-07-01
fix: path traversal in profile import allows arbitrary file write

seasnub783:
Component: ProfilesTab.java -> importProfile() (lines 151, 157-169)

Vulnerability: Path traversal in profile import allows arbitrary file write.

importProfile() takes entry keys from an imported .nbt file and passes them directly to new File(p.getFile(), filename) with no sanitization. The OS resolves .. segments, so a crafted entry key like ../../../mods/evil.jar writes outside the profile directory. The profile name (line 151) has the same issue -- a name like ../../.. shifts the base path itself, bypassing naive checks anchored to p.getFile().

Steps to reproduce:
1. Craft a .nbt compound where one entry key is ../../../<target> and its value is a ByteArrayTag containing arbitrary bytes
2. Open Profiles tab, click Import, select the file
3. The file is written to the resolved path outside the profile directory

Expected: Entry keys that resolve outside the profile directory are rejected.

Actual: No validation, any path the OS accepts gets written, to any file the user has write access to.

Fix: Resolve canonical path after construction and verify it stays under Profiles.FOLDER before opening FileOutputStream.

File f = new File(p.getFile(), filename).getCanonicalFile();
if (!f.toPath().startsWith(Profiles.FOLDER.getCanonicalFile().toPath())) continue;

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/settings/StringSetting.java
+++ b/src/main/java/meteordevelopment/meteorclient/settings/StringSetting.java
@@ -33,6 +33,10 @@ public class StringSetting extends Setting<String> {
 
     @Override
     protected boolean isValueValid(String value) {
+        if (filter == null) return true;
+        for (int i = 0; i < value.length(); i++) {
+            if (!filter.test(value, value.charAt(i))) return false;
+        }
         return true;
     }
 
--- a/src/main/java/meteordevelopment/meteorclient/settings/StringSetting.java
+++ b/src/main/java/meteordevelopment/meteorclient/settings/StringSetting.java
@@ -35,7 +35,7 @@ public class StringSetting extends Setting<String> {
     protected boolean isValueValid(String value) {
         if (filter == null) return true;
         for (int i = 0; i < value.length(); i++) {
-            if (!filter.test(value, value.charAt(i))) return false;
+            if (!filter.filter(value, value.charAt(i))) return false;
         }
         return true;
     }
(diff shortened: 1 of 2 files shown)


crosby-moe:
good fix for the entry keys, but for line 151 it probably was naively assumed to be safe when it was originally written based on the char filter defined on the `name` setting field
https://github.com/MeteorDevelopment/meteor-client/blob/8a66adcd69cf636c2c0a92c9eb8d50fa4222518b/src/main/java/meteordevelopment/meteorclient/systems/profiles/Profile.java#L31-L36
https://github.com/MeteorDevelopment/meteor-client/blob/8a66adcd69cf636c2c0a92c9eb8d50fa4222518b/src/main/java/meteordevelopment/meteorclient/utils/Utils.java#L671-L673
just checking whether `f.toPath().startsWith(Profiles.FOLDER.getCanonicalFile().toPath())` at the end of `ProfilesTab#importProfile` isn't comprehensive enough because it'd still result in a profile with a poisoned `Profile#getPath`, meaning arbitrary file write would still be possible through `Profile#save` & `Profile#delete`
a good fix for this would be to correct the `StringSetting#isValueValid` implementation to pass the string's characters through the filter, causing line 151's `Setting#set` to fail on `../../..` or any other invalid inputs
https://github.com/MeteorDevelopment/meteor-client/blob/8a66adcd69cf636c2c0a92c9eb8d50fa4222518b/src/main/java/meteordevelopment/meteorclient/settings/StringSetting.java#L34-L37

seasnub783:
i am very sorry crosby i will never let this happen ever

src/main/java/meteordevelopment/meteorclient/settings/StringSetting.java
```diff
@@ -33,6 +33,10 @@ protected String parseImpl(String str) {
 
     @Override
     protected boolean isValueValid(String value) {
+        if (filter == null) return true;
+        for (int i = 0; i < value.length(); i++) {
+            if (!filter.test(value, value.charAt(i))) return false;
```
crosby-moe:
```suggestion
            if (!filter.filter(value, value.charAt(i))) return false;
```

crosby-moe (changes requested):

PR #6520, opened 2026-07-17
Fix NoFall cancelBounce NPE on integrated server thread

Enqvy:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

The `cancelBounce` mixin on `EntityMixin.isSuppressingBounce` called `Modules.get().get(NoFall.class).cancelBounce()` with no null guard. This mixin runs on every entity's bounce calculation, including on the integrated server thread. When a non-player entity (e.g. a villager) stands on a slime block and the server ticks it, `Modules.get(NoFall.class)` returns `null`, throwing an NPE that crashes the entire game.

Fix: guard with `Utils.canUpdate()` (already used elsewhere in this mixin class) and an explicit null check, returning the original value when NoFall isn't active. No behavior change during normal client play.

Changed file: `src/main/java/meteordevelopment/meteorclient/mixin/EntityMixin.java`

## Related issues

None.

# How Has This Been Tested?

Reproduced the crash on a single-player world with a villager standing on a slime block (crash report: `NullPointerException ... Modules.get(NoFall.class)` in `modifyReturnValue$bpi000$meteor-client$cancelBounce`). Built Meteor from this branch, loaded it in-game, and confirmed the same world now loads and runs without crashing.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/mixin/EntityMixin.java
+++ b/src/main/java/meteordevelopment/meteorclient/mixin/EntityMixin.java
@@ -13,6 +13,7 @@ import meteordevelopment.meteorclient.events.entity.player.JumpVelocityMultiplie
 import meteordevelopment.meteorclient.events.entity.player.PlayerMoveEvent;
 import meteordevelopment.meteorclient.mixininterface.ICamera;
 import meteordevelopment.meteorclient.systems.modules.Modules;
+import meteordevelopment.meteorclient.systems.modules.movement.NoFall;
 import meteordevelopment.meteorclient.systems.modules.combat.Hitboxes;
 import meteordevelopment.meteorclient.systems.modules.movement.*;
 import meteordevelopment.meteorclient.systems.modules.movement.elytrafly.ElytraFly;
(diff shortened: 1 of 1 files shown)


Big-Iron-Cheems:
There's multiple issues here:
- you didn't need an entire LLM prompt to splurt this wall of text out
- you did not create a corresponding Issue report, including the crash log from which we may investigate this
- you did not even link said crash report in here, so we cannot analyse it ourselves anyways
- you did not provide any steps on how to reproduce this consistently; I tried in several ways and could not get it to torigger
- and finally, the LLM you let work on this also applied unnecessary and unrelated stylistic changes 

On the grounds on non-reproducibility and what I assume is lack of knowledge of Java, I cannot merge this in good faith.

crosby-moe:
`Modules.get().get(NoFall.class)` should never return `null` when a world is loaded unless something else has gone terribly wrong

4clsd:
> There's multiple issues here:
> 
> * you didn't need an entire LLM prompt to splurt this wall of text out
> * you did not create a corresponding Issue report, including the crash log from which we may investigate this
> * you did not even link said crash report in here, so we cannot analyse it ourselves anyways
> * you did not provide any steps on how to reproduce this consistently; I tried in several ways and could not get it to torigger
> * and finally, the LLM you let work on this also applied unnecessary and unrelated stylistic changes
> 
> On the grounds on non-reproducibility and what I assume is lack of knowledge of Java, I cannot merge this in good faith.

i just was testing some shit and it broke, when  i checked the crash log i found this and told the ai to fix it, its a 2 line fix

4clsd:
> `Modules.get().get(NoFall.class)` should never return `null` when a world is loaded unless something else has gone terribly wrong

i guess it did then idk

PR #6554, opened 2026-08-04
Added Ceasar cypher encryption/decryption to BetterChat module

Adog64:
## Type of change

- [ ] Bug fix
- [x] New feature

## Description

Added a simple Ceasar cipher to encrypt and decrypt chat messages. The feature has been added to the BetterChat module. Along with the toggle to enable/disable encryption, there is also an option to set the encryption offset (see [Ceasar Cipher on Wikipedia](https://en.wikipedia.org/wiki/Caesar_cipher)). To avoid weird unicode issues, the cipher only operates on the ASCII characters in the range [32, 126], which includes basically all the characters on an American QWERTY keyboard.

## Motivation

With the recent crackdowns on chat regulation on Minecraft servers from Microsoft, certain servers have implemented a chat filter. Chat filters are stupid (both in terms of player freedom and algorithmic complexity). Therefore even a simple cipher like a Ceasar cipher can get around them. I have no doubt that this is a permanent solution, but if this gets patched, I have a few more tricks up my sleeve.

# How Has This Been Tested?
Currently, I'm just running `./gradlew runClient` and opening a world to LAN for my vanilla client to join. Messages in chat are rendered correctly on the modded instance and show up as garbage in the vanilla instance. Importantly, echoing the same garbage on the vanilla instance, gets decoded on the modded side to produce the original message.

So far I've tested with all the typable characters on my keyboard and checked that unicode characters are unaffected. I've also done testing at different offset amounts (min, max, and in between).

image
image

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
@@ -48,8 +48,13 @@ import java.util.regex.PatternSyntaxException;
 
 public class BetterChat extends Module {
 	private final int FIRST_USABLE_ASCII = (int) ' '; // 32
-	private final int LAST_USABLE_ASCII = (int) '~'; // 126
+	private final int LAST_USABLE_ASCII = (int) '}'; // 125
 	private final int NUM_USABLE_ASCII = LAST_USABLE_ASCII - FIRST_USABLE_ASCII + 1;
+    private final int FIRST_LOWERCASE = (int) 'a';
+    private final int LAST_LOWERCASE = (int) 'z';
+    private final int FIRST_CAPITAL = (int) 'A';
+    private final int LAST_CAPITAL = (int) 'Z';
+    private final int NUM_LETTERS = 26;
 
     private final SettingGroup sgGeneral = settings.getDefaultGroup();
     private final SettingGroup sgFilter = settings.createGroup("Filter");
@@ -119,16 +124,6 @@ public class BetterChat extends Module {
         .name("decrypt")
         .description("Applies a decryption routine to encrypted chat messages")
         .defaultValue(true)
-        .build()
-    );
-
-    private final Setting<Integer> cypherOffset = sgGeneral.add(new IntSetting.Builder()
-        .name("cypher-offset")
-        .description("Ceasar cypher offset used for encrypting messages")
-        .defaultValue(20)
-        .range(0, NUM_USABLE_ASCII - 1)
-        .sliderRange(0, NUM_USABLE_ASCII - 1)
-        .visible(encrypt::get)
         .build()
     );
 
@@ -339,6 +334,8 @@ public class BetterChat extends Module {
     private void onMessageSend(SendMessageEvent event) {
         String message = event.message;
 
+        if (encrypt.get()) message = applyEncryption(message);
+
         if (annoy.get()) message = applyAnnoy(message);
 
         if (fancy.get()) message = applyFancy(message);
@@ -356,8 +353,6 @@ public class BetterChat extends Module {
             event.cancel();
             return;
         }
-
-        if (encrypt.get()) message = applyEncryption(message);
 
         event.message = message;
     }
@@ -552,28 +547,39 @@ public class BetterChat extends Module {
 
     private String applyEncryption(String message) {
         int msgLen = message.length();
-        int offset = cypherOffset.get();
+        int offset = Utils.random(0, 26); //cypherOffset.get();
         StringBuilder modString = new StringBuilder(msgLen + 2);
 
         // Encode offset into message with char for validation
-        modString.append((char) (FIRST_USABLE_ASCII + offset));
-        modString.append(encryptChar('%', offset));
+        modString.append('[');
+        modString.append((char) (FIRST_CAPITAL + offset));
+        modString.append(encryptChar('E', offset));
 
         for (int i = 0; i < msgLen; ++i) {
             char chr = message.charAt(i);
             chr = encryptChar(chr, offset);
             modString.append(chr);
         }
+        modString.append(']');
         return modString.toString();
     }
 
     private char encryptChar(char chr, int offset) {
         int ascii = (int) chr;
-        if (ascii > FIRST_USABLE_ASCII && ascii <= LAST_USABLE_ASCII) {
-            ascii -= FIRST_USABLE_ASCII;
+        if (Character.isUpperCase(chr)) {
+            ascii -= FIRST_CAPITAL;
             ascii += offset;
-            ascii %= NUM_USABLE_ASCII;
-            ascii += FIRST_USABLE_ASCII;
+            ascii += NUM_LETTERS;
+            ascii %= NUM_LETTERS;
+            ascii += FIRST_CAPITAL;
+            chr = (char) ascii;
+        }
+        else if (Character.isLowerCase(chr)) {
+            ascii -= FIRST_LOWERCASE;
+            ascii += offset;
+            ascii += NUM_LETTERS;
+            ascii %= NUM_LETTERS;
+            ascii += FIRST_LOWERCASE;
             chr = (char) ascii;
         }
         return chr;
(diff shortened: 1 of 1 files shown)


Adog64:
This is my first PR here, and I'm wondering what you mean by testing code in a production environment. Is that building a jar and using a mod loader or is there some CI/CD task runner that is used or something else? Thank you in advance!

Also, is there any additional housekeeping work that I should do (create and link an issue, write unit tests, etc.)?

crosby-moe:
> This is my first PR here, and I'm wondering what you mean by testing code in a production environment. Is that building a jar and using a mod loader or is there some CI/CD task runner that is used or something else? Thank you in advance!

testing in a production environment just involves using an actual minecraft launcher rather than fabric loom's `runClient` task

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
```diff
@@ -104,6 +108,25 @@ public class BetterChat extends Module {
...
     );
 
+    private final Setting<Boolean> encrypt = sgGeneral.add(new BoolSetting.Builder()
+        .name("encrypt")
+        .description("Applies an encryption routine to chat messages")
+        .defaultValue(false)
+        .build()
+    );
+
+    private final Setting<Integer> cypherOffset = sgFilter.add(new IntSetting.Builder()
```
crosby-moe:
```suggestion
    private final Setting<Integer> cypherOffset = sgGeneral.add(new IntSetting.Builder()
```

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
```diff
@@ -104,6 +108,25 @@ public class BetterChat extends Module {
...
+        .name("cypher-offset")
+        .description("Ceasar cypher offset used for encrypting messages")
+        .defaultValue(20)
+        .min(0)
+        .sliderMin(0)
+        .max(NUM_USABLE_ASCII - 1)
+        .sliderMax(NUM_USABLE_ASCII - 1)
```
crosby-moe:
```suggestion
        .range(0, NUM_USABLE_ASCII - 1)
        .sliderRange(0, NUM_USABLE_ASCII - 1)
```

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
```diff
@@ -511,6 +540,76 @@ private String applyAnnoy(String message) {
         return message;
     }
 
+    // Encrypt
+
+    private String applyEncryption(String message) {
+        int msgLen = message.length();
+        int offset = cypherOffset.get();
+        StringBuilder modString = new StringBuilder();
```
crosby-moe:
```suggestion
        StringBuilder modString = new StringBuilder(msgLen + 2);
```

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
```diff
@@ -511,6 +540,76 @@ private String applyAnnoy(String message) {
...
+        }
+        return chr;
+    }
+
+    // Decrypt
+
+    private String applyDecryption(String message) {
+        int msgLen = message.length();
+        StringBuilder modString = new StringBuilder();
+        int msgStart = message.indexOf("> ") + 2;
```
crosby-moe:
```suggestion
        int msgStart = message.indexOf("> ") + 2;
        StringBuilder modString = new StringBuilder(msgLen - (msgStart + 2));
```
needs double check, i mathed in my head

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
```diff
@@ -257,6 +280,10 @@ public BetterChat() {
     private void onMessageReceive(ReceiveMessageEvent event) {
         Component message = event.getMessage();
 
+        // Always decrypt messages regardless of whether we are encrypting them
+        String decryptedMessage = applyDecryption(message.getString());
+        message = Component.empty().append(decryptedMessage);
```
crosby-moe: i'm not sure whether always decrypting is the behaviour we want, but regardless the message should only be replaced if the text actually contained a caesar cipher encoded message so that:
1. we dont allocate a redundant `Component`
2. we dont override the styling information held in `message` unless we actually replace the text it contains

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
crosby-moe: this code incorrectly assumes all incoming chat messages will be decodable, which isn't true (ex. system messages)
specifically, `message.indexOf("> ")` may return `-1` and cause the second character of the message to be read as the cipher's offset which will result in gibberish

additionally, this function should *never* cause an exception to be thrown otherwise players would be able to remotely crash our users, so please double check everything

crosby-moe (changes requested):

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
Adog64: Your math looks correct to me

Adog64: Added some more robust logic here. Going to do more extensive testing in multiplayer in a prod env to make sure there are no unexpected crashes. This feels like a really good example of something I should make unit tests for though.

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
Adog64: I updated the `applyDecryption` method to have an `Optional` return type, so the caller has information as to whether decryption actually occurred or not.

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
crosby-moe: ideally yeah i guess, but we have no unit test setup in this project and frankly it’d be too much work

Adog64: Yeah as useful as they are, they are a pain to write.

Adog64:
I did some testing with this on 2b and I guess it was tripping the spam filter somehow, since my messages were getting blocked. I'm gonna work on refining my technique here, since 2b is my intended environment.

Adog64:
UPDATE: I found that I could bypass the spam filter by adding some real english text as a prefix to the encrypted text. I was inspired by "spammers" that use the prefix "fix chunk loading" followed by seemingly random characters. I've updated the code to allow prefixes and suffixes to be added without being encrypted. I've also updated the cypher to only modify letters and not add extra symbols to the chat as that seemed to be tripping something.

ilabeu:
its me timeduping , i thought of that the next week of the 2b dictatorship in chat but kinda boring what do u want with this in 2b lol

Adog64:
My goal here is to make a popular, easy to use chat tool that allows anyone to bypass the chat filters. I haven't tested two-way decryption on the server yet since I don't have an alt and I was looking for someone willing to help that was online. I'm curious how you figured out the spamming thing in chat? Is that just a standard random suffix to let you send the same message over and over?

ilabeu:
> My goal here is to make a popular, easy to use chat tool that allows anyone to bypass the chat filters. I haven't tested two-way decryption on the server yet since I don't have an alt and I was looking for someone willing to help that was online. I'm curious how you figured out the spamming thing in chat? Is that just a standard random suffix to let you send the same message over and over?

hause does not care about chat bypassing at all , all he wants is that mojang sees the server and doesnt ban it thats why he added that , and yea if u send messages that are not the same u can keep spamming thats why the letters at the end of the msg , i think once i got muted but been doing for a while and nothing happend

src/main/java/meteordevelopment/meteorclient/systems/modules/misc/BetterChat.java
```diff
@@ -511,6 +543,103 @@ private String applyAnnoy(String message) {
         return message;
     }
 
+    // Encrypt
+
+    private String applyEncryption(String message) {
+        int msgLen = message.length();
+        int offset = Utils.random(0, 26); //cypherOffset.get();
+        StringBuilder modString = new StringBuilder(msgLen + 2);
```
crosby-moe, line 551:
```suggestion
        StringBuilder modString = new StringBuilder(msgLen + 4);
```

crosby-moe (changes requested):

PR #6615, opened 2026-08-28
BetterTab: add Friends only option.

qnxt:
## Type of change

- [ ] Bug fix
- [X] New feature

## Description

Add friends-only filtering to BetterTab. 9b player count is peaking and I rather just see my friends. I also visit other servers like 6b time to time and they have 600 players time to time, and I rather just only see my friends.

## Related issues

None.

# How Has This Been Tested?
2026-08-28T16:38:35+00:00

# Checklist:

- [ ] My code follows the style guidelines of this project. (idk how to style the stream api stuff for ur project, but I do run google's styling in my IDE)
- [ ] I have added comments to my code in more complex areas. (irrelevant)


Powie69:
+1
or maybe they should be on the top/1st index of the tablist

qnxt:
Added that. However it seems like there is another issue, basically if the server has more than 100 players, your friends who are listed after the first 100 aren't shown at all, unless the tablist size is increased. This needs more polish. I'll get around it ~~these days~~ eventually, or welcome anyone else to figure it out in that time.

PR #6639, opened 2026-09-03
Stopped Nuker overwriting the range setting every tick

### LINKED ISSUE #6609: Nuker's "range" setting is silently overwritten every tick in Uniform Cube mode
### Describe the bug

In Nuker.java's onTickPre() handler, the range setting is written to
(not just read) on every single game tick when Uniform Cube shape is
selected:

    if (shape.get() == Shape.UniformCube) range.set((double) Math.round(range.get()));

This calls the setting's set() method (a mutation, likely triggering
change listeners / marking config dirty for saving) every tick, purely to
round the displayed value - even when the value hasn't actually changed
between ticks.

Impact:
- Unnecessary repeated writes to a setting every tick (20 times/sec) for
  as long as Nuker is active with Uniform Cube shape selected, instead of
  only rounding the value once when it's changed or shape is switched.
- If the setting system persists/marks-dirty on every .set() call
  (common pattern), this could cause unnecessary repeated config save
  triggers purely from Nuker being left on.
- If a user is actively dragging the range slider, this per-tick
  overwrite could fight with/snap the UI value unexpectedly mid-drag.

### Steps to reproduce

1. Enable Nuker.
2. Set shape to "UniformCube".
3. Observe (e.g. via a settings-change listener/log, or by dragging the
   range slider while active) that the range value is being rewritten
   every tick rather than only when actually changed.

### Meteor Version

26.1.2-42

### Minecraft Version

26.1.2

### Operating System

Linux

c8dhjp4tyv-bit:
## Type of change

- [x] Bug fix
- [ ] New feature

## Description

`Nuker.onTickPre()` rounds the `range` setting by writing back to it:

```java
if (shape.get() == Shape.UniformCube) range.set((double) Math.round(range.get()));
```

Two problems with doing this from the tick handler:

- It mutates the user's config. A range of `4.4` typed into the slider is silently rewritten to `4.0` and saved that way, so switching the shape back to Sphere or Cube leaves the user with a value they never chose.
- It runs 20 times a second for as long as Nuker is enabled. `Setting#set` has no equality check, so `onChanged` fires on every tick even when the value is already rounded, and the value fights the slider while it is being dragged.

The rounded value is already computed three lines later as `int r`, which the uniform cube branch uses for `pos1`/`pos2`. This drops the `set()` call and compares the chebyshev distance against `r` instead of `range.get()`, so the break shape is unchanged while the setting is left alone.

The `BlockIterator.register()` extents also had to move off the raw `range`. They are the reach of the shared iterator, and the uniform cube filter only accepts blocks within the rounded `r`, so a fractional range would have made the iterator walk a larger volume purely to throw the extra blocks away - `4.4` would grow it from 11x11x9 to 13x13x11, and because the radii are shared, any other module registered that tick would have been dragged along. The extents are now computed per shape alongside the existing `maxh`/`maxv`, which keeps the iteration footprint identical to the old rounded behaviour.

## Related issues

Closes #6609

# How Has This Been Tested?

`./gradlew compileJava` passes.

With shape set to `UniformCube` and range `4`, `r` is `4` both before and after the change, the iterator radii are `5`/`4` as before, and the same blocks are selected. The visible difference is that a fractional range now stays in the settings screen instead of snapping on the next tick.

# Checklist:

### DIFF
--- a/src/main/java/meteordevelopment/meteorclient/systems/modules/world/Nuker.java
+++ b/src/main/java/meteordevelopment/meteorclient/systems/modules/world/Nuker.java
@@ -351,10 +351,9 @@ public class Nuker extends Module {
         double rangeSq = Math.pow(range.get(), 2);
         BlockPos playerBlockPos = mc.player.blockPosition();
 
-        if (shape.get() == Shape.UniformCube) range.set((double) Math.round(range.get()));
-
         double pX_ = pX;
         double pZ_ = pZ;
+        // Uniform cube works on whole blocks, so round the range locally instead of writing it back to the setting
         int r = (int) Math.round(range.get());
 
         if (shape.get() == Shape.UniformCube) {
@@ -410,7 +409,7 @@ public class Nuker extends Module {
                         return;
                 }
                 case UniformCube -> {
-                    if (chebyshevDist(playerBlockPos.getX(), playerBlockPos.getY(), playerBlockPos.getZ(), blockPos.getX(), blockPos.getY(), blockPos.getZ()) >= range.get())
+                    if (chebyshevDist(playerBlockPos.getX(), playerBlockPos.getY(), playerBlockPos.getZ(), blockPos.getX(), blockPos.getY(), blockPos.getZ()) >= r)
                         return;
                 }
                 case Cube -> {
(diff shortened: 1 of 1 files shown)


c8dhjp4tyv-bit:
Hi @MukjepScarlet, when you have a chance, could you please take a look at this PR? It’s been open for about two weeks and I’d appreciate a review. Thanks!

PR #6653, opened 2026-09-09
feat(auto-fish): add offhand fishing rod support

Big-Iron-Cheems:
## Type of change

* [ ] Bug fix
* [x] New feature

## Description

Refactors `AutoFish` to use an explicit state machine for managing the fishing lifecycle.

The previous implementation tracked the fishing lifecycle through `wasHooked` and several conditional checks. 
This change introduces explicit `IDLE`, `WAITING_FOR_BITE`, and `WAITING_TO_REEL` states, making the module's behavior easier to follow and maintain.

Adds support for fishing rods in the offhand and resolves which hand currently contains a usable rod when casting or reeling. 
This also handles moving the rod between the main hand and offhand while fishing.

The existing fishing behavior is preserved, including cast/catch delays, anti-break handling, automatic rod switching, and automatic recasting.

## Related issues

N/A

# How Has This Been Tested?

Tested in-game to verify that:

* Fishing rods in the offhand are used correctly.
* Hotbar rod selection and auto-switching continue to work.
* Casting and reeling use the hand that currently contains a usable fishing rod.
* Moving the fishing rod between the main hand and offhand while fishing works correctly.
* Bite detection and catch delays continue to work as expected.
* Anti-break behavior remains unchanged.
* Automatic recasting continues to work.

# Checklist:

* [x] My code follows the style guidelines of this project.
* [x] I have added comments to my code in more complex areas.
* [x] I have tested the code in both development and production environments.


Eccys:
~~i wonder if it's possible to do double fishing by using both offhand and main hand at the same time and intelligently switch.~~

it won't work; it'll just reel in the first rod.

Big-Iron-Cheems:
Reminder: preserve the existing auto-switch behavior while adding offhand support.

Currently `getRodHand()` short-circuits `findBestRod()` whenever either hand contains a usable rod. 
This means a player holding a worse rod will no longer switch to a better rod in the hotbar.

The intended behavior should remain that `findBestRod()` has priority when `auto-switch` is enabled: 
select the best usable hotbar rod, while still allowing an already-held/offhand rod to be used when appropriate.

Also, `findBestRod()` should also calculate the score of the offhand rod if present.

PR #6655, opened 2026-09-12
reduce allocations in explosion damage calculations

xpncvr:
## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Optimization

## Description

Reduced allocation overhead in `DamageUtils`:
- `getExposure` reuses a single `Vec3` and `ExposureRaycastContext` instead of creating both per ray
- `protectionReduction` hoists its map allocation out of the armor slot loop. No `clear()` is needed because `Utils.getEnchantments` already clears the map

This drops every `crystalDamage` / `bedDamage` / `anchorDamage` call from 94 temporary objects to 3 for a standing player hitbox.

## Related issues

None

# How Has This Been Tested?

Verified that rendered crystal damage is still the same in game

# Checklist:


crosby-moe:
the choice to allocate a `Vec3` rather than using `IVec3` was intentional, i think if we could entirely eliminate the `IVec3` interface and restore `Vec3`’s immutability, it would be optimal

though if it can be profiled/benchmarked to significantly improve performance this way i will concede

xpncvr:
Yeah fair enough its not worth it then, the difference is negligible I profiled it.

crosby-moe:
the enchantments optimization looks correct though, that part should still be merged imo

xpncvr:
I dropped the other changes and left just the enchantment part

crosby-moe (approved):
awesome

